PR #7, opened 2015-11-30
Add proxy config

petehouston:
It's my case, well, other developers might need this proxy option as well.

@DavertMik should add a configuration section for `codecept.json`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,6 +280,24 @@ need to update the `helpers.WebDriverIO.proxy` key.
 }
 ```
 
+For example,
+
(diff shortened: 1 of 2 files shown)


DavertMik:
Can you provide usage example? 
Also could you not overwrite `this.options.desiredCapabilities` entirely? ;)

petehouston:
@DavertMik  sorry, that was my stupid mistake. Just update it with usage on readme.

DavertMik:
Thanks!

Issue #9, opened 2015-12-05
Init test folder not working


antonioribeiro:
When I try to configure:

![image](https://cloud.githubusercontent.com/assets/3182864/11605098/2bc19298-9ade-11e5-9970-09e9fc91a8e9.png)

I get

```
  {
    "tests": "./*_test.js",
    "timeout": ...
  }
```

Output and Include are working fine.

DavertMik:
Thanks. WIll fix it

PR #17, opened 2015-12-09
Closes Codeception/CodeceptJS#9 , add the test folder to the codecept.json object

alfirin:
Add the test folder pass in argument to the init command to the resulted created json file.


DavertMik:
Thanks

PR #28, opened 2015-12-18
fix wrong require path in steps_file.js generated by init command

bigsan:


DavertMik:
Thanks. I thought that `require` should work for a path relative to main file, which is `lib/codecept.json`.
Can you suggest how to configure project so the path `codeceptjs/actor` would actually work? I don't like to have `lib` prefix everywhere

DavertMik:
Fixed in slightly different manner: https://github.com/Codeception/CodeceptJS/blob/master/CHANGELOG.md#022
 Thank you

Issue #31, opened 2015-12-18
Cannot find module 'codeceptjs/actor'


therebelrobot:
Going through the [Quickstart from the website](http://codecept.io/quickstart/), ran into the following error when running `codeceptjs run --steps`:

```
└─[$] codeceptjs run --steps                                                             ◈ (ruby-2.2.1) ⧂ (go1.4) ⬡ (node-4.2.3) [9:12:59]
CodeceptJS v0.2.1
Test root is assumed to be /Users/therebelrobot/git/path/to/project
/Users/therebelrobot/.nvm/versions/node/v4.2.3/lib/node_modules/codeceptjs/lib/container.js:45
      throw new Error(`Initialization failed for ${objects[name]}\n${err.message}`);
      ^

Error: Initialization failed for function () {
  return require('codeceptjs/actor')({

    // Define custom steps here, use 'this' to access default methods of I.
    // It is recommended to place a general 'login' function here.

  });
}
Cannot find module 'codeceptjs/actor'
    at createSupportObjects (/Users/therebelrobot/.nvm/versions/node/v4.2.3/lib/node_modules/codeceptjs/lib/container.js:45:13)
    at Object.module.exports.create (/Users/therebelrobot/.nvm/versions/node/v4.2.3/lib/node_modules/codeceptjs/lib/container.js:62:15)
    at Codecept.init (/Users/therebelrobot/.nvm/versions/node/v4.2.3/lib/node_modules/codeceptjs/lib/codecept.js:38:15)
    at Command.module.exports (/Users/therebelrobot/.nvm/versions/node/v4.2.3/lib/node_modules/codeceptjs/lib/command/run.js:11:12)
    at Command.listener (/Users/therebelrobot/.nvm/versions/node/v4.2.3/lib/node_modules/codeceptjs/node_modules/commander/index.js:301:8)
    at emitTwo (events.js:87:13)
    at Command.emit (events.js:172:7)
    at Command.parseArgs (/Users/therebelrobot/.nvm/versions/node/v4.2.3/lib/node_modules/codeceptjs/node_modules/commander/index.js:615:12)
    at Command.parse (/Users/therebelrobot/.nvm/versions/node/v4.2.3/lib/node_modules/codeceptjs/node_modules/commander/index.js:458:21)
    at Object.<anonymous> (/Users/therebelrobot/.nvm/versions/node/v4.2.3/lib/node_modules/codeceptjs/bin/codecept.js:75:9)
```

Environment:

[...]

PR #33, opened 2015-12-18
docs(install): Updated README to include local dev dependency

therebelrobot:
see https://github.com/Codeception/CodeceptJS/issues/31


DavertMik:
https://github.com/Codeception/CodeceptJS/blob/master/CHANGELOG.md#022

PR #34, opened 2015-12-18
docs(quickstart): added local dev dep install

therebelrobot:
see https://github.com/Codeception/CodeceptJS/issues/31


DavertMik:
No local dev install needed :0 https://github.com/Codeception/CodeceptJS/blob/master/CHANGELOG.md#022

PR #41, opened 2015-12-26
Fix a wrong function call in helper.js

pim:


DavertMik:
Thanks. That's important

Issue #51, opened 2016-01-22
I.waitWhileVisible


fabioel:
Hi
I wanted to ask if it is possible to add a function `I.waitWhileVisible` that waits as long as a given selector is visible in the DOM. It would be really awesome!

DavertMik:
Looks more like `I.waitToHide(el)`. But yes, that can easily be implemented. If you familiar with WebDriverIO you can send a Pull Request to WebDriverIO helper. It's pretty easy

PR #54, opened 2016-01-26
add waitToHide

fabioel:
@DavertMik: I have added the feature that I proposed in https://github.com/Codeception/CodeceptJS/issues/51


DavertMik:
Thanks!

PR #56, opened 2016-01-26
I.countAppearances

fabioel:
@DavertMik 
I came up with a new feature: counting the appearances in the dom of a given element
this is very useful when working with tables or auto generated fields.

### DIFF
--- a/docs/helpers/WebDriverIO.md
+++ b/docs/helpers/WebDriverIO.md
@@ -227,6 +227,21 @@ I.click({css: 'nav a.login'});
 -   `link`
 -   `context`
 
+## countAppearances
+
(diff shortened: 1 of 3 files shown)


DavertMik:
Thanks. 
Actually there is a naming convention - all methods that return values should start with `grab` prefix, so we'd know they require `yield` to work with them. 

But what you propose is more like [seeNumberOfElements](http://codeception.com/docs/modules/WebDriver#seeNumberOfElements) from Codeception. It doesn't return number of elements but performs an assertion (which looks to me more useful)

Can you implement it in a similar syntax?

fabioel:
@DavertMik I updated the function :smile:

DavertMik:
Thanks!

PR #58, opened 2016-01-30
better support for webdriverio options

sourcec0de:
The original method of passing options to webdriverio caused some
confusion. If the user wanted to simply pass a webdriverio config
directly through the codeceptjs config some properties would get
overridden with undefined if they didn’t set `browser` or `url`.

It makes more sense to allow the passing of a webdriverio config rather
than attempting to map these properties and construct that object since
it can get extremely complex.

I.E

```js
// how you're currently doing it
{
  "helpers": {
    "WebDriverIO": {
      "proxy": {
        ...
      }
    }
  }
}

// it should be more like this
{
  "helpers": {
    "WebDriverIO": {
      "desieredCapabilities": {
        "proxy": {
          ...
        }
      }
    }
  }
}
```

### DIFF
--- a/lib/helper/WebDriverIO.js
+++ b/lib/helper/WebDriverIO.js
@@ -1,4 +1,5 @@
 'use strict';
+let _ = require('lodash');
 let Helper = require('../helper');
 let webdriverio = require('webdriverio');
 let stringIncludes = require('../assert/include').includes;
@@ -35,8 +36,8 @@ let withinStore = {};
  *
  * This helper should be configured in codecept.json
  *
- * * `url` - base url of website to be tested
- * * `browser` - browser in which perform testing
+ * * `baseUrl` - base url of website to be tested
+ * * `desiredCapabilities.browserName` - browser in which perform testing
  * * `window_size`: (optional) default window size. Set to `maximize` or a dimension in the format `640x480`.
  *
  *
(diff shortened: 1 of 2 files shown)


lib/helper/WebDriverIO.js
```diff
@@ -117,8 +197,16 @@ class WebDriverIO extends Helper {
   }
 
   _before() {
-    this.browser = webdriverio.remote(this.options).init();
+
+    if (this.options.multiremote) {
```
sourcec0de: @DavertMik added basic support for multi remote

lib/helper/WebDriverIO.js
```diff
@@ -100,12 +162,30 @@ class WebDriverIO extends Helper {
 
   constructor(config) {
     super(config);
-    this.options = config;
```
sourcec0de: @DavertMik better support for directly passing a webdriverio config

DavertMik:
Thanks. I will have some time to review and merge it then.

DavertMik:
Thanks. Sorry for late reply, I was working on Protrator helper, so didn't have time to take major changes into WebDriverIO. I like your PR and I will merge it. The only thing, I'd keep is `browser` configuration in the root of `WebDriverIO`

PR #72, opened 2016-03-02
add grabCssFrom

fabioel:
@DavertMik please review this feature!
this is the adapted version of http://webdriver.io/api/property/getCssProperty.html

### DIFF
--- a/lib/helper/WebDriverIO.js
+++ b/lib/helper/WebDriverIO.js
@@ -427,6 +427,20 @@ class WebDriverIO extends Helper {
   }
 
   /**
+   * Retrieves a cssProperty from an element located by CSS or XPath and returns it to test.
+   * Resumes test execution, so **should be used inside a generator with `yield`** operator.
+   *
+   * ```js
+   * let pin = yield I.grabCssFrom('#pin', 'width');
+   * ```
+   */
+  grabCssFrom(locator, cssProperty) {
+    return this.browser.getCssProperty(withStrictLocator(locator), cssProperty).then(function (prop) {
+      return prop;
+    });
+  }
+
+  /**
    * Retrieves a value from a form element located by CSS or XPath and returns it to test.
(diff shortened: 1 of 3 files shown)


lib/helper/WebDriverIO.js
```diff
@@ -427,6 +427,20 @@ class WebDriverIO extends Helper {
   }
 
   /**
+   * Retrieves a cssProperty from an element located by CSS or XPath and returns it to test.
+   * Resumes test execution, so **should be used inside a generator with `yield`** operator.
+   *
+   * ```js
+   * let pin = yield I.grabCssFrom('#pin', 'width');
```
DavertMik: pin is not suitable variable name in this example

test/unit/helper/WebDriverIO_test.js
```diff
@@ -442,6 +442,12 @@ describe('WebDriverIO', function () {
         .then(() => wd.grabAttribute({css: 'form'}, 'method'))
         .then((val) => assert.equal(val, "get"));
     });
+
+    it('should grab css from element', () => {
+      return wd.amOnPage('/')
+        .then(() => wd.grabCssFrom('#area1', 'prop'))
```
DavertMik: you are not getting a CSS property here, but a value of an attribute. That's what method `grabAttribute` does.

test/data/app/view/index.php
```diff
@@ -11,7 +11,7 @@
 </p>
 
 
-<div id="area1">
+<div id="area1" prop="myTest">
```
DavertMik: test example is invalid here

DavertMik:
Looks like your test doesn't correspond to feature, so tests are failing. Please fix.

Also just curious, can you describe your use-case for this feature?

fabioel:
@DavertMik you are right I got confused with grabAttribute...

so there is no real need for this

PR #73, opened 2016-03-02
grabHTMLFrom

fabioel:
@DavertMik please review, this is the adapted version of `wd.getHTML`
http://webdriver.io/api/property/getHTML.html

### DIFF
--- a/lib/helper/WebDriverIO.js
+++ b/lib/helper/WebDriverIO.js
@@ -427,6 +427,20 @@ class WebDriverIO extends Helper {
   }
 
   /**
+   * Retrieves the innerHTML from an element located by CSS or XPath and returns it to test.
+   * Resumes test execution, so **should be used inside a generator with `yield`** operator.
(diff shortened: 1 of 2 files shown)


lib/helper/WebDriverIO.js
```diff
@@ -427,6 +427,20 @@ class WebDriverIO extends Helper {
   }
 
   /**
+   * Retrieves the innerHTML from an element located by CSS or XPath and returns it to test.
+   * Resumes test execution, so **should be used inside a generator with `yield`** operator.
+   *
+   * ```js
+   * let pin = yield I.grabHTMLFrom('#pin');
```
DavertMik: Exmaple could be better with better names

```
 let postHtml = yield I.grabHTMLFrom('#post');
```

fabioel:
@DavertMik i updated the example

DavertMik:
Thanks

PR #80, opened 2016-03-15
fix WebDriverIO helper scrollTo

sensone:
```
  I.scrollTo('.selector', 1500, 400)
  **FAIL**  this.browser.scrollTo is not a function
```
WebDriverIO has no scrollTo function. I think it was typo.
Now, it's work.


DavertMik:
Thanks!

PR #82, opened 2016-03-17
I.clearField()

fabioel:
@DavertMik please review this feature!

It is the adaption of [clearElement](http://webdriver.io/api/action/clearElement.html)


DavertMik:
Thanks. Looks really useful

Issue #78, opened 2016-03-14
I.seeElement


fabioel:
I have a question regarding `I.seeElement`:
I find it quite strange that the `I.seeElement` method checks for the element in the DOM and not if the element is actually visible in the DOM, because the `I.see` method checks for the visiblity of the text and not its existance

Wouldn't it be better to implement the `I.seeElement` like this: [Webdriver isVisible](http://webdriver.io/api/state/isVisible.html)
And change the current `I.seeElement` to something like `I.locateElement`

DavertMik:
> I find it quite strange that the I.seeElement method checks for the element in the DOM and not if the element is actually visible in the DOM,

You are right, it should check for visibility as well. At least this is how it works in Codeception, we have seeElementInDOM for DOM chekcing

PR #83, opened 2016-03-17
Change I.seeElement()

fabioel:
@DavertMik I have come up with a proposal for Issue: https://github.com/Codeception/CodeceptJS/issues/78

Tell me if this was what you had in mind!

P.S: I think there is something wrong in the test suit. I can't run it locally....


DavertMik:
Thanks, but not really this... I'd rather not stick to the viewport as this makes tests really fragile, you should always make sure that the browser window you run acceptance tests in is the same size of a browser you run manual tests. 

So I'd ask you to use [isVisible](http://webdriver.io/api/state/isVisible.html) instead.

DavertMik:
Could you update PR (merge it with master, and resend it) ?

DavertMik:
Merged manually. Thanks!

Issue #62, opened 2016-02-09
Support for mouseDown, mouseMove, and mouseUp


hanthomas:
While click and rightClick works for most cases, we also need to be able to perform additional mouse functions to simulate drag and drop.

DavertMik:
Thanks for the idea. I think the overall implementation is pretty simple and you can get done it yourself and send a PR. We use WebdriverIO as backend, and they have all those methods implemented in nice manner. I think they have even drag-and-drop thing. And you can submit pull request with adding those methods to CodeceptJS. All you need to write them in analogy with `rightClick` or `doubleClick` (i.e. to allow usage of strict locators and wrapping promises).

I will be glad to merge it

hanthomas:
Seems simple enough.  I noticed that the docs mention being able to extend WebDriverIO via custom helpers.  So, I was heading down that path.  But, this functionality seems generic enough to create a PR so others can benefit.

DavertMik:
Yes, sure, generic solved issues are welcome. That's the mission of CodeceptJS - to solve common issues in nice way )

DanielRuf:
I think this is still relevant.

PR #84, opened 2016-03-17
Feature/mouse features

fabioel:
@DavertMik I implemented the features suggested in Issue: https://github.com/Codeception/CodeceptJS/issues/62

I did not write the tests yet, because my test runner is not working. It does not show any errors, even if I break existing tests.
Can you have a look at the `package.json` and the `gulp` file and check if everything is fine?
It would also be very great if you could add a `development` section to the `README.md`, so that it is more clear how to run tests and what to do in order to help improving `CodeceptJS`


DavertMik:
I'm not fan of adding methods for pressing up and down mouse buttons, as they can't be treated as complete action. If someone needs a complex interaction, they should implement it on their own and name it accordingly. Otherwise it won't be clean what is user trying to do.

However, dragAndDrop should be added for sure. I hope you don't mind if I just copy your code (and won't merge everything else)

gaurav21r:
@DavertMik We are testing a complex GUI drawing application and Codecept has been great so far. Unfortunately, Drag and Drop (the current API) implementation works for apps like Trello but not Drawing Tools like ours. In our case, mouseup & mousedown are actually atomic interactions for us.

Take a look at this animation:

amaze-rectangle

This can only be tested with

```
I.mousedown(x, y)
I.mousemove(x,y)
I.mouseup(x, y)
```

Requesting you to please reconsider merging this completely?

DavertMik:
> Requesting you to please reconsider merging this completely?

Nope 😇 You won't need to have a test looking like 

```js
I.mousedown(x, y)
I.mousemove(x,y)
I.mouseup(x, y)
```
that's not useful and not readable. You will need to implement an action insidea custom helper using webdriverio API directly. This way you can customize the behavior the way you like and create an action that will fulfill your needs. 

This would look far more better:

```js
I.drawRectangle(x1,y1,x2,y2);
```

JosueDOrtiz:
@DavertMik ,

is this still the case, not wanting mouseDown and MouseUp? Seems similar to pressKeyDown and pressKeyUp.

In my case, I am looking into ways to drag a column in a table so as to change the column order.

storenth:
Did't understand the state of the PR. Looks like no feature provided yet.

PR #87, opened 2016-03-22
Chore/fixes v0.3.0

fabioel:
@DavertMik I realized that we use grabTextFrom, grabValueFrom, but we always had grabAttribute
So I changed it to grabAttributeFrom to stick to the style of the other functions

I also implemented doubleClick for selenium (please have a look at this, not 100% sure)


DavertMik:
Thanks!

PR #90, opened 2016-03-27
Implemented WebDriverIO frame() with new method switchTo() in helper

xdrr:
## Overview

I've been a big fan of CodeceptJS but missed support for IFrames since I switched from a Mocha & WebDriverIO testing stack to CodeceptJS. This merge implements the `frame()` method from https://github.com/webdriverio/webdriverio with a new method named `switchTo()` in the WebDriverIO helper.

`switchTo()` works the same as `frame()` by allowing the test case to change context to content inside an IFrame specified by a locator. Specifying a null locator (the default) will return the test case to the parent frame (usually the current window).

## What's in the merge?

In addition to the new method in the WebDriverIO helper, I've added three test cases and updated the helper specific documentation.

### DIFF
--- a/docs/helpers/WebDriverIO.md
+++ b/docs/helpers/WebDriverIO.md
@@ -882,3 +882,12 @@ Waits for a function to return true (waits for 1sec by default).
 
 -   `fn`  
 -   `sec`  
+
+## switchTo
+
+Switches context to the content within an IFrame element.
+A null locator will return the context to the current window.
+
+**Paramters**
+
+-   `locator`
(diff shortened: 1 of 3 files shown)


DavertMik:
> In addition to the new method in the WebDriverIO helper, I've added three test cases and updated the helper specific documentation.

Thanks! 

Btw, Helpers documentation is taken from dockblocks of methods running `gulp docs` command. So probably your doc may be overwritten. But nevermind, I will try to make docs valid. Thank you for your contribution!

Issue #93, opened 2016-03-30
#89 is not fixed


fabioel:
Hi @DavertMik!

The issue #89 is still not fixed. I tried it once again. I am currently using version `0.3.3` and the tests are not failing when they should!

I have a very simple test once again:

```
'use strict';
const config = require('../../config.js');
const expect = require('chai').expect;

Feature('Navigation');

Before((I) => {
  I.amOnPage(config.baseUrl);
});
Scenario.only( 'Test', (I) => {
    expect(true).to.be.eql(false);
});
```
Here is my terminal output:
```
fabio@fabio-XPS-15-9530:~/Documents/Project/git/core-feature-tests$ node ./node_modules/.bin/codeceptjs run --steps
CodeceptJS v0.3.3
Test root is assumed to be /home/fabio/Documents/Project/git/core-feature-tests

Navigation --
Test
 • I am on page "https://www.test.com"
fabio@fabio-XPS-15-9530:~/Documents/Shore/git/core-feature-tests$ 
```

as you can see the tests don't fail they just finish

Here is the output if I change the assertion (`expect(true).to.be.eql(true)`):
```
Navigation --
Test
• I am on page "https://www.test.com"
 ✓ OK in 1ms
```

fabioel:
__UPDATE__:

This also happens with the `CodeceptJS` assertions. Like `I.see('asdfasdf')` when `'asdfasdf'` is not on the page, the test suite does not fail properly

trikonasana:
Hi @fabioel,
 
if you run node ./node_modules/.bin/codeceptjs run --steps; echo $?

i guess there is a '0' at the end of your output. So I think my bug (#98) is a duplicate of yours (or v.v.)

Is it?

Cheers
tri

tjoskar:
I had the same problem until I manually created the `output` folder. I thought it was a windows-related problem (#126) since I only could reproduce the problem on a windows machine but it turned out that the `output` folder was missing when I tried on the windows machine and once I created the folder it starts working again.

smozely:
Same issue for me on OSx ... it seems like the fix in #94 fixes the reporting, but breaks screenshots. Also in all cases, also the browser isn't shut down correctly on failure. The existence of the output folder only seems to effect recording of the screenshots on the current HEAD for me.

current HEAD
- without the output folder existing, no screenshot, tests don't fail properly
- with the output folder existing, screenshot recorded, tests don't fail properly

with PR #94 applied
[...]

PR #94, opened 2016-03-30
Bug/assertion

fabioel:
@DavertMik this is what i meant in #93


DavertMik:
Thanks, however your solution was not perfect as it didn't make screenshots to be saved on failures. Closing in favor of https://github.com/Codeception/CodeceptJS/pull/162

PR #97, opened 2016-04-04
Bug/run test suite

fabioel:
@DavertMik I have found an error in the test runner

It was not possible to run specific folder (suites) with CodeceptJS

I have implemented a way that works like this:

`codeceptjs run services --steps` that would run all tests in the folder `services` that lies in the testRoot


DavertMik:
Thanks!

DavertMik:
Sorry, I didn't notice this at first sight but this PR can't be merged. You assume that testRoot == current working dir, which is not 100% right.

DavertMik:
I was using locally

```
./bin/codeceptjs run examples
```

probably lots of developers use:

```
./node_modules/codeceptjs/bin/codecept run tests
```
or something like that. So testroot != cwd.

fabioel:
@DavertMik for me the `codecept run test` is only working the the directory in which the `codeceptjs.json` lies, so I cannot understand how you would want to run the tests from anywhere else than the test root.

FYI: I am using `./node_modules/codeceptjs/.bin/codceptjs` and so far it has been working fine.

DavertMik:
You can provide a path to your root with 3rd parameter. 
```
./bin/codeceptjs run examples
```

will execute tests from `examples` dir

fabioel:
Oh, ok. I thought this parameter is meant to execute the suite only for a specific directory. Not the execute it from the directory.

Now it makes sense. When I have time, I will improve that to let the suite run a specific folder

DavertMik:
Thanks. Will be waiting for it )

PR #106, opened 2016-04-11
Fix path resolution on the command line to support absolute paths

denis-sokolov:
Before the change, I could not `codecept run /my/absolute/path`.
`path.join` would always stick the current working directory in front.
More proper path resolution is also simpler, as `path.resolve` will stick
the current working directory if the path is not already absolute.


DavertMik:
Thanks!

PR #108, opened 2016-04-11
Fix path resolution for the helpers config values to support absolute paths

denis-sokolov:
Before the change, I could not set an absolute path as my helpers path.
`path.join` would always stick the project directory in front.
`path.resolve` will, however, check if the path is already absolute, in which case it won’t do anything.


DavertMik:
Thank you!

PR #113, opened 2016-04-18
Fixed bug in generate.js

### LINKED ISSUE #105: TypeError: Cannot read property '1' of null in generate.js:44
without regex in braces  command "codeceptjs gt *tests dir* " failed with error TypeError..

So this
 let defaultExt = config.tests.match(/\*(.*?)$/)[1] || '_test.js';
should be replace with
 let defaultExt = config.tests.match((/\*(.*?)$/)[1]) || '_test.js';
It help me to create new test

jlsjonas:
added parentheses around the regex on line 44, fixes #105

### DIFF
--- a/lib/command/generate.js
+++ b/lib/command/generate.js
@@ -41,7 +41,7 @@ module.exports.test = function (genPath) {
   output.print(`Creating a new test...`);
   output.print('----------------------');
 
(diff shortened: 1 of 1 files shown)


DavertMik:
thanks!

Issue #59, opened 2016-01-30
Codecept config should be JS not JSON


sourcec0de:
@DavertMik I think it would be more useful to load the config as a `.js` file rather than `.json`.
I would be happy to chat about this and submit a pull request for this feature if I can get some direction on developing on `codeceptjs`. I don't like wondering in the dark trying to figure out where everything lives :)

You get some beneifits in doing this.
- environemtnal variables can be controlled outside so you don't have to configure a seperate config per environment
- it can be better controlled programatically
- you can check everything into source control without exposing secrets.

``` js
/**
 * CodeCeptJS Config
 */

module.exports = {
  tests: './tests/*_test.js',
  timeout: 10000,
  output: './tests/reports/',
  helpers: {
    WebDriverIO: {

      // load variables from the environment and provide defaults
      url: process.env.CODECEPT_URL || 'http://localhost:3000',
      browser: 'chrome',

      user: process.env.CLOUDSERVICE_USER,
      key: process.env.CLOUDSERVICE_KEY,

      desiredCapabilities: {
        'browserstack.local': true,
        'browserstack.debug': true
      },

      capabilities: [{
        browserName: 'chrome',
        'browserstack.local': true,
        'browserstack.debug': true
      }],

      logLevel: process.env.WEBDRIVERIO_LOG_LEVEL || 'warn',
      coloredLogs: true,
      waitforTimeout: 10000
    }
  },
  include: {},
  bootstrap: false,

  // don't build monolithic configs
  mocha: require('./tests/mocha_config.js') || {},
  name: 'vortex'
};
```

DavertMik:
I was thinking on such declaration but json seemed simpler to me. Maybe you are right. At least I need some more time to think on it.

Sorry, right now I'm busy with my other projects, so I will get back here soon and provide a proper response.

adrianblynch:
This could easily be backwards compatible too.

I think I prefer JS over JSON for config now.

DustinDSchultz:
If you want to be able to provide the full range of Protractor settings (including the callbacks) you'll need to use a .js file. I'm running into this issue right now.

DioNNiS:
+1

stephane-ruhlmann:
Still not released on npm ?

Psychedelix:
+1

stephane-ruhlmann:
Released with : https://github.com/Codeception/CodeceptJS/commit/9d606f9350b790cedc104858900dde57d15a17fb

[...]

PR #141, opened 2016-05-19
Add in loading config from a JS file

smozely:
Make it easier to configure CodeceptJS
- Not include keys in config files / source control
- Run on different browsers via env flags

RE Issue #59

### DIFF
--- a/lib/command/utils.js
+++ b/lib/command/utils.js
@@ -23,8 +23,6 @@ module.exports.getConfig = function (testRoot) {
     jsonConfigFile = path.join(testRoot, 'codecept.json');
   if (fileExists(jsConfigFile)) {
     config = require(jsConfigFile).config;
-    console.log(config);
-
   } else if (fileExists(jsonConfigFile)) {
     config = JSON.parse(fs.readFileSync(jsonConfigFile, 'utf8'));
   } else {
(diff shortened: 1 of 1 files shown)


DavertMik:
Thank you! 

> Run on different browsers via env flags

Btw, could you explain how to use env flags?

smozely:
@DavertMik your onto them quick today :smile:

For using env settings ... with this I can write my codecept.js config file like this ...

```
exports.config = {
  helpers: {
    WebDriverIO: {
      url: process.env.url, 

... more configs ...
```

I can then run 
```url=https://beta.example.com codeceptjs run --steps```

To control what url the tests run against, good for switching between localhost, test, prod ... same logic applies to the keys for things like sauce labs. Even going to use it to control running our test suite against multiple Browser/OS combinations.

smozely:
Im lost as to why the build is breaking, my changes should have no effect if you don't have a codecept.js file sitting where the config lives

DavertMik:
> Im lost as to why the build is breaking, my changes should have no effect if you don't have a codecept.js file sitting where the config lives

Looks like it was Travis issue

DavertMik:
> To control what url the tests run against, good for switching between localhost, test, prod ... same logic applies to the keys for things like sauce labs. Even going to use it to control running our test suite against multiple Browser/OS combinations.

Ok, that makes sense. I think I should mention this in documentation

Issue #120, opened 2016-04-28
Typescript typings?


inakianduaga:
I'm new to CodeceptJS, just went through the install and got a short test running. I noticed no typescript definitions available, which is a bummer since there is no IDE autocompletion (and no compiler check) when writing tests using Typescript. 

I'm using webdriverio and noticed there is a [webdriver.d.ts](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/webdriverio/webdriverio.d.ts) file (for version 3.3.0). The docs of CodeceptJS show a lot of methods that supposedly are based on webdriverio helper, but I can't find a lot of those in the type definition above, so not sure if they were added to later versions or if they are actually part of CodeceptJS and not WebdriverIO. 

At any rate, I just quickly threw an incomplete type file based on the dump from `codeceptjs list` 

```ts
/// <reference path="../../../typings/webdriverio/webdriverio.d.ts" />

type ICodeceptCallback = (i: CodeceptJS.I) => void;

declare const Feature: (string: string) => void;
declare const Scenario: (string: string, callback: ICodeceptCallback) => void;
declare const Before: (callback: ICodeceptCallback) => void;
declare const After: (callback: ICodeceptCallback) => void;
declare const within: (selector: string, callback: Function) => void;

declare namespace CodeceptJS {
  export interface I extends WebdriverIO.Client<void> {
    amOnPage: (url: string) => any;
    rightClick: (locator: string) => any;
    fillField: (field: string, value: string) => any;
    appendField: (field: string, value: string) => any;
    selectOption: (select, option) => any;
    attachFile: (locator: string, pathToFile) => any;
    checkOption: (field: string, context) => any;
    grabTextFrom: (locator: string) => any;
    grabHTMLFrom: (locator: string) => any;
    grabValueFrom: (locator: string) => any;
    grabAttributeFrom: (locator: string, attr) => any;
    seeInTitle: (text: string) => any;
    dontSeeInTitle: (text: string) => any;
    grabTitle: () => any;
    see: (text: string, context) => any;
    dontSee: (text: string, context) => any;
    seeInField: (field: string, value: string) => any;
    dontSeeInField: (field: string, value: string) => any;
    seeCheckboxIsChecked: (field) => any;
    dontSeeCheckboxIsChecked: (field) => any;
    seeElement: (locator: string) => any;
    dontSeeElement: (locator: string) => any;
    seeElementInDOM: (locator: string) => any;
[...]

PR #150, opened 2016-06-12
Typescript support docs

andreportela:
Hi guys! This project is awesome <o/

This is a tiny doc update to make explicit how to use CodeceptJS with TypeScript based on #120 and #135.


DavertMik:
Thank you. This is a good to way to promote typescript definitions usage

PR #155, opened 2016-06-20
Locate element by name attribute bug fixed

borisarzentar:


DavertMik:
Thank you! Such stupid bug )

PR #159, opened 2016-06-30
[WIP] Nightmare

DavertMik:
Initial Nigthmare integration.

* [x] amOnPage
* [x] click: (locato
* [x] doubleClick: (locato
* [x] see: (tex
* [x] dontSee: (tex
* [x] selectOption: (selec
* [x] fillField: (fiel
* [x] pressKey
* [x] attachFile: (locato
* [x] seeInField: (fiel
* [x] dontSeeInField: (fiel
* [x] appendField: (fiel
* [x] checkOption: (fiel
* [x] seeCheckboxIsChecked
* [x] dontSeeCheckboxIsChecked
* [x] grabTextFrom
* [x] grabValueFrom
* [x] grabAttributeFrom: (locato
* [x] seeInTitle
* [x] dontSeeInTitle
* [x] grabTitle
* [x] seeElement
* [x] dontSeeElement
* [x] seeElementInDOM
* [x] dontSeeElementInDOM
* [x] seeInSource
* [x] dontSeeInSource
* [x] executeScript
* [x] executeAsyncScript
* [x] seeInCurrentUrl
* [x] dontSeeInCurrentUrl
* [x] seeCurrentUrlEquals
* [x] dontSeeCurrentUrlEquals
* [x] saveScreenshot
* [x] setCookie
* [x] clearCookie
* [x] seeCookie
* [x] dontSeeCookie
* [x] grabCookie
* [x] resizeWindow: (widt
* [x] wait
* [x] waitForElement: (locato
* [x] waitForVisible: (locato
* [x] waitForText: (text, se

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,9 @@ before_install:
   - wget http://selenium-release.storage.googleapis.com/2.52/selenium-server-standalone-2.52.0.jar
 
 before_script:
-  - tree -d .
+  - chmod -R 777 test/data
+  - ls -ll test/data
+  - ls -ll test/data/output
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
   - "java -jar selenium-server-standalone-2.52.0.jar -port 4444 >/dev/null 2>&1 &"
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ before_install:
   - wget http://selenium-release.storage.googleapis.com/2.52/selenium-server-standalone-2.52.0.jar
 
 before_script:
+  - tree -d .
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
   - "java -jar selenium-server-standalone-2.52.0.jar -port 4444 >/dev/null 2>&1 &"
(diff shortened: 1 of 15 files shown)


DavertMik:
Some key ideas about doing Nigtmare integration. Practically everything is built around `evaluate` function which executes function inside a browser window and returns result back to node. **Only basic types can be passed as arguments to that function**. 

I've added `findElements` action which searches for all elements by CSS/Xpath. As HTMLObject can't be returned back, this function returns array of property values from elements. I think something similar should be done for click, doubleClick, etc, to find and interact with elements. Probably native `click` function can't be used as it accepts only CSS but we can learn [how it is done](https://github.com/segmentio/nightmare/blob/master/lib/actions.js#L85) so to implement it in similar way. 

Current work: 

* update seeElementInDOM, seeInSource, etc to use `findElements` action
* add method similar to `findElements` to interact with elements

DavertMik:
There is a problem: Nightmare doesn't support file upload. There is a plugin for it https://github.com/rosshinkley/nightmare-upload/blob/master/nightmare-upload.js but it handles also only CSS selectors, thus should be copy-pasted and rewritten :(

sergeyklay:
👍

PR #167, opened 2016-07-21
Use a single browser instance per suite for Protractor/Selenium helper

### LINKED ISSUE #156: CodeceptJS Protractor - one instance of the browser
By default each scenario starting new instance of the browser. Is there any chance to change it and run all scenarios and features in one instance of the browser?

P.S. this is performance concern.

Nighthawk14:
Instead of opening/closing the browser before/after each tests, this is doing it before/after each suite. Only available for Protractor/Selenium helper at the moment, I'm happy to bring it to webdriver.io if we agree that the feature is good. Also any suggestion/improvement welcome :).

Fix #156 

ps: That speeds up the tests as well from 11min to 6


DavertMik:
Thanks. The feature is good but I think it should be configurable. I will review it tomorrow and if everything is ok I will merge it and try to do same for webdriverio

DavertMik:
Thank you. SeleniumWebdriver, Protractor and WebDriverIO has `restart` option to enable/disable using single browser session for all tests. By default it is set to `true` which was the default behavior. To not restart browser between tests set `restart: false`

PR #174, opened 2016-07-28
add custom steps to definition list

jayS-de:
When generating a step definition file steps.d.ts the custom defined steps (e.g. form custom_steps.js) are not listed.

The PR gets all supported actions and adds them to the definition file as well as to the list command


DavertMik:
Looks really helpful. Thank you!

PR #186, opened 2016-08-04
Run arbitrary JS code after tests

joefitter:
* added teardown method to Codecept class - this checks config for a teardown path the same way as bootstrap, and requires it if present
* call teardown when mocha fires `end` event

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -11,8 +11,8 @@ Scenario('check Welcome page on site', (I) => {
 }
 ```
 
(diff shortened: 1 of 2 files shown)


DavertMik:
Thank you! Just was thinking about it

joefitter:
@DavertMik many thanks for the merge! Could you please let me know when you can publish a new release?

DavertMik:
Sure. I have some patches on my own to finish. Probably I will do it today

joefitter:
brilliant - thanks again!

PR #188, opened 2016-08-05
fix(webdriverio): missing init for multiremote

douglasduteil:
Following http://webdriver.io/guide/usage/multiremote.html


DavertMik:
Thank you!

PR #189, opened 2016-08-05
Add support to translations to codeceptjs

abner:
Added feature to allow define translations to the helpers actions and also to the default `I`actor.

Added a folder translations where contributors would add translations for their own languagues.

The feature actually works like this: In your concept.json file you add a translation key with some of the name of a supported language or you can also point to a file in your project path.

**Example**

```json
{
  "tests": "./e2e/features/*.js",
  "timeout": 10000,
  "output": "./e2e/outputs",
  "translation": "pt-BR",
  "helpers": {
    "Protractor": {
      "url": "http://localhost:3000/",
      "host": "127.0.0.1",
      "port": 4444,
      "browser": "phantomjs",
      "rootElement": "body"
    }
  },

  "bootstrap": false,
  "mocha": {},
  "name": "my-project"
}
```

Using this will allow to write your steps in portuguese:

```javascript

Feature('Acesso a tela de Login');

Scenario('Testar se a tela de login está acessível', (Eu) => {
    Eu.naoEstouEmAplicacaoAngular();
    Eu.estouNaPagina('http://localhost:3000/')
    Eu.vejo('My App');
});

```

And the report to `codeceptjs run --steps` will show

Acesso a tela de Login --
 Testar se a tela de login está acessível
 • Eu nao estou em aplicacao angular 
 • Eu estou na pagina "http://localhost:3000/"
 • Eu vejo "My App"
 ✓ OK in 413ms

  OK  | 1 passed   // 462ms


lib/translations/pt-BR.js
```diff
@@ -0,0 +1,56 @@
+
+
+module.exports = {
```
mfdeveloper: **Suggestion:** The explicit call to `module.exports` is really necessary? I think that the translation file needs be a plain `.json` file instead a `.js` file

mfdeveloper:
Awesome @abner !! This is a amazing feature *o*

DavertMik:
Wow! Really cool.

lib/translations/pt-BR.js
DavertMik: Looks reasonable to me to have it as `.json` but I don't see if it is that important

DavertMik:
Thank you. I will play a bit with that

DavertMik:
I have done some refactoring there #194 
The next thing we should think about is to use localized Scenario/Before/Feature function names.
Also the question is about documentation. Ideally we should translate function references, in reality it will be pretty hard to do, so probably just add method aliases into the Helper references.

DavertMik:
Ok, I have written small reference on using translations: http://codecept.io/reports/
@abner could you help me a bit and add some Portuguese examples to  https://github.com/Codeception/CodeceptJS/blob/master/docs/translation.md ? 

Thanks!

DavertMik:
Hey @mfdeveloper @abner if you still use CodeceptJS please join or Slack. https://join.slack.com/t/codeceptjs/shared_invite/enQtMzA5OTM4NDM2MzA4LTNiMTE5YzNlMjQyNWYwZjAzODE3ZTI1MmVkYjY0YTYyNDZmMWJjM2VlODg3NTQzYzM2OGE5ZmIwODFjODk1MzI

We have Portuguese channel as well!

sacuribraimo:
Hi,
Is there a way to localize the messages used in I.see('My App') to a message file where there is translation to 'Minha App'?

PR #191, opened 2016-08-05
Bootstrap - Default file to run phantomjs/selenium

mfdeveloper:
Added new `bootstrap.js` default file to start [phantomjs ](https://github.com/ariya/phantomjs) or selenium server using [webdriver-manager](https://github.com/bonigarcia/webdrivermanager). Now, the `codecept.json` can be configure with the keys:

``` js
{
 ...
  bootstrap: "bootstrap.js" //Added by default
  bootstrapAsync: true //Allow async tests execution
}
```

Added to a **async** execution to allow run tests only after run the bootstrap.

Some these adjusts were made with help of @abner

### DIFF
--- a/lib/bootstrap.js
+++ b/lib/bootstrap.js
@@ -1,0 +1,65 @@
+var spawn = require('child_process').spawn;
+var path = require('path');
+var sleep = require('sleep').sleep;
+
+function Bootstrap(codecept, callback) {
+
+  var phantomjs, seleniumPid, callbackCalled = false;
+
+  (function () {
+
+    var protractor = codecept.config.helpers.Protractor;
+    if (protractor && protractor.browser) {
+      if (/phantom/.test(protractor.browser)) {
+        var phantomPath = require('phantomjs').path || '/usr/local/bin/phantomjs';
+        phantomjs = spawn(phantomPath,['--webdriver=4444']);
+
+        phantomjs.stdout.on('data', function (data) {
+          if (!callbackCalled) {
+            callback();
+            callbackCalled = true;
+          }
+        });
+
+        process.on('exit', function () {
+          phantomjs.stdin.pause();
+          phantomjs.kill();
+        });
+
+      }else {
+
+        var binPath = path.join(path.dirname(require.resolve('protractor')),'..','..','.bin');
+        var webdriver = spawn(path.join(binPath,'webdriver-manager'),['start']);
+
+        webdriver.stdout.on('data', function (data) {
+
+          // Workaround to wait execution of selenium .jar
+          sleep(2);
+
+          var output = data.toString();
+          if (/seleniumProcess.pid/.test(output)) {
+            seleniumPid = output.match(/:\s\d{1,}/)[0].replace(/:\s/,'');
+          }
+          if (!callbackCalled) {
+            callback();
+            callbackCalled = true;
+          }
+        });
+
+        process.on('exit', function () {
+          webdriver.stdout.pause();
+          webdriver.kill();
+
+          if (seleniumPid) {
+            process.kill(seleniumPid, 'SIGINT');
+          }
+        });
+
+      }
+    }
+
+  })();
+
+}
+
+module.exports = Bootstrap;
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -11,8 +11,8 @@ Scenario('check Welcome page on site', (I) => {
 }
 ```
 
-Tests are expected to be written in ECMAScript 6. 
-Each test is described inside a `Scenario` function with `I` object passed into it. 
+Tests are expected to be written in ECMAScript 6.
+Each test is described inside a `Scenario` function with `I` object passed into it.
 I object is an **actor**, an abstraction for a testing user. I is a proxy object for currently enabled **Helpers**.
 
 ```json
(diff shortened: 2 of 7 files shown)


lib/bootstrap.js
```diff
@@ -0,0 +1,65 @@
...
+var path = require('path');
+var sleep = require('sleep').sleep;
+
+function Bootstrap(codecept, callback) {
+
+  var phantomjs, seleniumPid, callbackCalled = false;
+
+  (function () {
+
+    var protractor = codecept.config.helpers.Protractor;
```
DavertMik: this works not only for Protractor but for `WebDriverIO`, `SeleniumWebDriverJS` as well.
probably you should match the first helper with `browser` property in it

lib/bootstrap.js
```diff
@@ -0,0 +1,65 @@
...
+        });
+
+        process.on('exit', function () {
+          phantomjs.stdin.pause();
+          phantomjs.kill();
+        });
+
+      }else {
+
+        var binPath = path.join(path.dirname(require.resolve('protractor')),'..','..','.bin');
```
DavertMik: could you comment, what does that mean?

DavertMik:
Thanks, look interesting. I need some more time to review and think on it before merging.

lib/bootstrap.js
```diff
@@ -0,0 +1,65 @@
...
+
+function Bootstrap(codecept, callback) {
+
+  var phantomjs, seleniumPid, callbackCalled = false;
+
+  (function () {
+
+    var protractor = codecept.config.helpers.Protractor;
+    if (protractor && protractor.browser) {
+      if (/phantom/.test(protractor.browser)) {
```
DavertMik: I'd suggest to move phantomjs initialization and webdriverio initialization into functions (will make the code more readable)

lib/bootstrap.js
mfdeveloper: Yeah, sure!! I made this feature using only  `Protractor` because my app use `Angular`. I don't use `WebDriverIO`, but  I can change a refactory to check the others helpers.

lib/bootstrap.js
mfdeveloper: Which line? The `require.resolve()` call to mount the path? So, this is to get the binary path of `webdriver-manager` package. It is a dependency of `Protractor`

lib/bootstrap.js
mfdeveloper: Sure, sure!! I'm agree with you. I think that a better approach is use a `ES6` class structure, or a Closure. I did this feature during one day, writing a behavior for a app in my job. I will make a refactor for that. Many thanks :)

lib/bootstrap.js
DavertMik: Ah, got it

DavertMik:
I have some suggestions on improving that feature. It is really cool but It is not general and configurable enough to make it work for everyone. So here are my ideas:

Create a new directory `lib/driver` and place classes for `webdriver-manager` and `phantomjs` into it.
Then add `driver` option to configuration, so it could be set like:

```
"driver": "webdriver-manager"
```

(for instance we can create a docker driver which will start/stop selenium inside a container)

In your current implementation you are trying to guess driver by name of a browser. That's ok, but explicit configuration might be useful in some cases. So you can provide option like

```
"driver": "auto"
```

For autodetection. Actually you can even create `lib/driver/auto` file to provide such detection and instantiate either webdriver or PhantomJS. 

This should not be added `bootstrap` because current bootstrap/teardown feature is for arbitrary user's code. So let's call it driver, create ES6 classes for each driver, place them into `lib/driver` and instantiate them if `driver` option is set. Does it sound ok to you? If so I'm waiting for updates :+1:

DavertMik:
@mfdeveloper added some suggestions to improve this :up:

mfdeveloper:
@DavertMik Many thanks for your attention and suggestions! 

So, the `driver` approach is great, thinking in separate of `bootstrap`. But, if the user pass a value to the `browser` key that conflict with the new driver config? Something like this:

```js
{
 "helpers": {
   "Protractor": {
     "browser": "firefox"
   }
 },
  "driver": "phantomjs" //Conflict with "browser" key. What happens?
}
```
The default bootstrap execution suggested by this MR is copied to the user project directory, and can be customized by him. I understand your concerns, but, I'm confuse like you thought about this flow execution. 

About the `selenium` execution, How the steps will wait for `docker` execution? Notice that, I need use the [sleep](https://www.npmjs.com/package/sleep) node package to wait for seconds to start the `.jar` of selenium. Do you have a more elegant solution for this?

Futhermore, the `bootstrap` execution can continue async?

DavertMik:
> About the selenium execution, How the steps will wait for docker execution? Notice that, I need use the sleep node package to wait for seconds to start the .jar of selenium. Do you have a more elegant solution for this?

No, I don't, so it will be ok to wait a bit for that.

Yes, I thought about the controversy between browser config and driver config. I think your current config will work, PhantomJS won't throw an excpetion if a firefox will be set as a browser. 
I don't see a good way to bind driver to browser config, only by "guessing" it. This way I propose to have "auto" option for a driver which will pick the right driver. However, the config should be explicit so anyone could use the driver of their choice, easily switch it, or add their own drivers.

DavertMik:
Implemented via hooks. See documentation

Issue #169, opened 2016-07-25
Mocha jUnit reporter file not created when path is specified


Nighthawk14:
When using mocha-junit-reporter with a specific path for the ouput file: 
`--reporter-options mochaFile=test-results.[hash].xml`

The file is not created.  It works fine when running tests directly with mocha and the command line like that :
`mocha --reporter mocha-junit-reporter --reporter-options mochaFile=./reports/api/test-results.[hash].xml`

Not sure if the issue is with CodeceptJs but as it is working directly with mocha I think is good to get it here anyway.

ps: When the option is not specified the file is created but in root directory.

Thanks!

DavertMik:
Thanks, I will take a look

aldarund:
I have similar problem. But on circleci test results doesnt appear at all no matter what i try to do

aldarund:
On windows test-results.xml appears when both specify env var and skipping it. On linux i cant make it create test no matter what, so i cant make it work on circleci

aldarund:
Ok, even more. Sometime test result created, sometime they are not. same code same test. Its very unreliable.  And its not only with mocha junit. Also with other reporters like mocha-multi etc.

DavertMik:
Fixed in #193

PR #193, opened 2016-08-07
Reporters

DavertMik:
* improved generation of reports
* step trace generation moved to corresponding listener
* steps do not include ASCII chars
* error lines are added to step trace
* internal stack trace is not show in non-verbose mode


DavertMik:
Should also fix #169

PR #194, opened 2016-08-08
Translation

DavertMik:
* Refactored trasnlation support
* Added translation option to generator
* Added Russian translation


abner:
Oh, awesome!!

When a new version with all lastest changes will be published on npmjs?

PR #196, opened 2016-08-08
Fixed using single browser session for testing

DavertMik:


romal-fitbmc:
Is not fixed. 
Option restart: false/true - does not matter - the last opened browser does not close.

PR #223, opened 2016-09-05
Fix def command

jankaspar:
Container does not have getTranslations() method, just translation().


DavertMik:
Yeah, thank you! I just was working on this too.

DavertMik:
Published new release

PR #226, opened 2016-09-12
make seleniumAddress for SeleniumWebdriver

vectoroc:
If you do not want to bother with selenium-standalone-server, SeleniumWebdriver can do it for you. You have just comment line `.usingServer(this.options.seleniumAddress);`


vectoroc:
I missed that it is possible to specify seleniumAddress: false in config and it will work

PR #231, opened 2016-09-21
BeforeSuite and AfterSuite implement

APshenkin:
For https://github.com/Codeception/CodeceptJS/issues/232


DavertMik:
Thank you. I will need some time to have a deep look. So this is Work In Progress?

APshenkin:
@DavertMik Yes. This enhancement is "In progress". I have problem with `AfterSuite` hook, when any scenario was failed during test run. Please check https://github.com/Codeception/CodeceptJS/issues/232

APshenkin:
@DavertMik Looks like I fixed the issue with AfterSuite Hook. Now it works! Please review. 
Any suggestions will be interesting.

BTW tested with WebdriverIO helper on Windows

APshenkin:
@DavertMik When you will be able to check this changes? Looks like it's working!

DavertMik:
That's really cool. I hope I will have a time to review and merge it after the weekend. Thanks for your work.

DavertMik:
What I need from you is an example of usage of this feature so I could document it. 
If you can, please add some real life sample. Why you needed this feature.

APshenkin:
@DavertMik 
Here is sample from my real life:)
During tests I use webdrivercss to test CSS https://github.com/webdriverio/webdrivercss . So for this I need reference screenshots. It's very expensive, to save all screenshots to repository, because the repository will be too large and also during tests run you don't need all screenshots. Because of this I use the remote server to save screenshots. Before tests I have to download them and during scenarios I check the CSS using these screenshots. And after test suite I upload difference to remote server.

Also You can use these feature, If your web application has any integration with client application. E.g. web application connects to client application in my project using Websockets to send/get information about installed apps(games) and statuses about installation on PC. Also webapplication calls client app to launch the game. So in my web tests sometimes I need to install different versions of client app for tests (with different settings) to check that everything works well. But I don't need to reinstall app after each test or one time (on launching tests). Because of this BeforeSuite|AfterSuite are best for me

DavertMik:
Thank you very much! That was awesome work

DavertMik:
Just a question BeforeSuite will work for *all files in suite* or only for a file where it was declared?

DavertMik:
Ok, I didn't come to a better way to write about it than to cite you on site http://codecept.io/basics/#beforeSuite 

If you think this can be improved, please edit https://github.com/Codeception/CodeceptJS/blob/master/docs/basics.md

APshenkin:
@DavertMik Only for a file, where it was declared.

BTW, beforeSuite and afterSuite have access to I object. here is a sample from my tests:
```
BeforeSuite((I) => {
    I.syncDown('4game-referral-system', 'acc-1-qa-auth-referral');
});
```
note: I create custom handler in my custom helper to sync files from remote server. It'works as usual `I` function. But in `BeforeSuite/AfterSuite` you don't have an access to browser because it's not running in this moment. Will update the docs and send new PR

PR #240, opened 2016-10-03
Update docs for beforeSuite/afterSuite and fix restart false

APshenkin:
Update docs and also fix `restart:false` option for WebdriverIO helper (before_suite was called two times (before run all suites(files) and before run the file. It causes two calls of start webdriver and break codeceptjs). Now Browser restarts only when new suite(file) started


APshenkin:
@DavertMik Hi, please review and merge because now `restart:false` is broken

DavertMik:
Hi, thanks. Merging

PR #249, opened 2016-10-10
Fix typo in attachFile documentation

jplegoff:
There is a typo in the Webdriverio helper documentation, in function attachFile it mentions remove system instead of remote system


DavertMik:
Thanks for fix. However those docs are generated from the source code of corresponding helpers. Could you fix that in https://github.com/codeception/codeceptjs/blob/master/lib/helper/WebDriverIO.js

DavertMik:
Fixed in https://github.com/Codeception/CodeceptJS/commit/e475d627bed5f0a48989e694294ee5de47b51ceb

Issue #248, opened 2016-10-07
Documentation for SeleniumWebdriver lists Protracter helper


SuneRadich:
In the section about SeleniumWebdriver, the example in the section Access from helpers, lists 

```
this.helpers['Protractor'].browser
```

Where it should list
```
this.helpers['SeleniumWebdriver'].browser
```

DavertMik:
Thank you for noticing. 

Actually you can help to solve that. Just edit `helpers/SeleniumWebdriver.js` file and its docblock comments. Update it and make a pull request. That would be your contribution to CodeceptJS )

PR #250, opened 2016-10-10
Fix typo in SeleniumWebdriver helper doc

jplegoff:
Fixes doc issue reported in https://github.com/Codeception/CodeceptJS/issues/248


DavertMik:
Thanks, but thise docs are geenratedfrom docblocks of source code of that helpers. In this case this is teh line https://github.com/Codeception/CodeceptJS/blob/master/lib/helper/SeleniumWebdriver.js#L54

DavertMik:
Fixed in https://github.com/Codeception/CodeceptJS/commit/2a95079037e5efed0dbebf867c87f7a26dee4740. Anyway, thanks for pointing!

PR #251, opened 2016-10-12
Use global actor function in Page Objects docs

giuband:
As suggested in https://github.com/Codeception/CodeceptJS/blob/master/CHANGELOG.md#022 , otherwise the example code might not work as expected.


DavertMik:
Thanks!

PR #254, opened 2016-10-14
Fix attachFile on WebDriverIO helper

giuband:
I noticed that attachFile has not been working as expected in the WebDriverIO helper. It seems there is a small bug in the code when using the object returned from findFields. This change should fix it (I've tested it and it's now working as expected).

### DIFF
--- a/lib/helper/WebDriverIO.js
+++ b/lib/helper/WebDriverIO.js
@@ -478,7 +478,7 @@ class WebDriverIO extends Helper {
     }
     return findFields(this.browser, locator).then((el) => {
       this.browser.uploadFile(file).then((res) => {
-        if (!el.value.length) {
+        if (!el.value || el.value.length === 0) {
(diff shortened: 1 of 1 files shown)


DavertMik:
Sorry it looks like I had false positive test for this case :(
Thank you for fixing it

lib/helper/WebDriverIO.js
```diff
@@ -478,10 +478,10 @@ class WebDriverIO extends Helper {
     }
     return findFields(this.browser, locator).then((el) => {
       this.browser.uploadFile(file).then((res) => {
```
DavertMik: here should be:

```
return this.browser.uploadFile
```

DavertMik:
Ok, here goes some improvements from my side with fixed tests :sunglasses:

giuband:
Awesome job, thank you for all your efforts on this outstanding library!

Issue #190, opened 2016-08-05
When using webdriverio helper configured for saucelab scenario name not passed to saucelab


aldarund:
In saucelab all test are named as Unnamed job instead of scenario name
http://c2n.me/3AW2fsa
its not very easy to find right test when there no name for it.
Is there a way to pass scenario name to saucelab ?

aldarund:
To do this its needed to pass `name` param to the desiredCapabilities of webdriver.io
I tried to do it in _before but despite in jsdoc its specified that test param is passed, its not passed, so i cant get a test name there.

DavertMik:
Yes, unfortunately due to restrictions in mocha I didn't manage to run `before`/`after` hooks inside a test, so they do not have test name in it. I will think how this can be improved

DavertMik:
To be honest I'm not sure how to make it work. Browser session is started before the actual test executed, and probably desired capabilities can't be updated during test execution.

cnworks:
I have the same problem trying to give meaningful names to sessions on browserstack.

The most promising approach was to create a custom helper and set the name in the _before() or _beforeSuite() hook.
`this.helpers["WebDriverIO"].options.desiredCapabilities.name = "foo";`
Unfortunately I couldn't find a solution to access the name of the suite/feature or the scenario. Also  be aware that this solution depends on the sorting in the configuration file, cause it needs to be run before the WebDriverIO Helper _before() / _beforeSuite() hook.

Maybe someone can improve on that.

APshenkin:
https://github.com/Codeception/CodeceptJS/issues/190

PR #255, opened 2016-10-14
Added option to manually start the browser.

cnworks:
To work with cloud services like browserstack it can be beneficial to control when the session is started and therefore start the session manually.

One case in which this is useful is issue #190. You can set the `desiredCapability.name` and then start the session by executing `this.helpers['WebDriverIO']._startBrowser()` 

As the manual option is set to false by default the behavior doesn't change.

### DIFF
--- a/docs/helpers/WebDriverIO.md
+++ b/docs/helpers/WebDriverIO.md
@@ -23,6 +23,7 @@ This helper should be configured in codecept.json
 -   `url` - base url of website to be tested
 -   `browser` - browser in which perform testing
 -   `restart` - restart browser between tests (default: true), if set to false cookies will be cleaned but browser window will be kept.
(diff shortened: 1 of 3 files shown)


DavertMik:
Thanks for the PR. I like the implementation but I don't like naming, probably they should be more explicit about the problem it solves. `manualStart` looks like it expects you to launch a browser locally and wait the Selenium to handle its session (somehow). 

Maybe it should be named like `reuseSession` or smth else... With documentation for potential use case at cloud service providers

DavertMik:
@cnworks when I merge your PR could you help the project a bit more. I was thinking it would be useful to have dedicated "Cloud Testing" guide in documentation. Could you write some paragraphs on it, to share your experience?

PR #258, opened 2016-10-14
Add support for webdriver.io timeouts

easternbloc:
This adds support for each possible timeouts option in webdriver.io

In your codecept.conf.js you can express it something like:

```
{
  ...
  WebDriverIO: {
    url: 'http://localhost:8080',
    browser: 'phantomjs',
    timeouts: {
      implicit: 10000,
     'page load': 20000
      script: 5000
    }
  }
}
```

### DIFF
--- a/lib/helper/WebDriverIO.js
+++ b/lib/helper/WebDriverIO.js
@@ -229,6 +229,18 @@ class WebDriverIO extends Helper {
       this.browser = webdriverio.remote(this.options).init();
     }
 
+    if (this.options.timeouts) {
+      if (this.options.timeouts.implicit) {
+        this.browser.timeouts('implicit', this.options.timeouts.implicit);
+      }
+      if (this.options.timeouts['page load']) {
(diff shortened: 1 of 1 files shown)


DavertMik:
Wow, thanks. Those options should be documented as well, I think I can do it.

easternbloc:
thanks!

DavertMik:
I think this should options should be defined per tests, as well. Or inside the code...
It would be useful to define special timeout settings before some heavy pages or data to be loaded

easternbloc:
@DavertMik yeah although I think some of the timeouts once set run for the duration of the webdriver instance so not quite sure how that'd play out...

DavertMik:
Probably we need more methods to be added to do this, like:

```
I.defineTimeouts({ script: 5000 });
```

DavertMik:
better to call this function another way for not to be confused with real `setTimeout` ;)

PR #260, opened 2016-10-16
fix codeceptjs gt file extension and glob support

Zougi:
codeceptjs gt does not generate proper file without this, file has no extension. Also config.tests globs like `./test/folder/**/*.js` are not supported without this


Zougi:
also I haven't made the change in the pull request but may I suggest that default extension would be `.js` instead of `_test.js` ...we already are in a folder dedicated for test automation. We know these files are test files. It makes little sense to add in the filename `_test`

DavertMik:
Thanks!

Issue #262, opened 2016-10-17
Yahoo Example fail and hangs with nightmarejs


Zougi:
yahoo example from page <http://codecept.io/nightmare/> fails & hangs forever...
That's ok that it fails. In my opinion that makes it even a better example than a test that succeed..but codecept should not hang forever!

```
$ node .\node_modules\codeceptjs\bin\codecept.js run --verbose
CodeceptJS v0.4.7
Using test root "C:\Users\zougi\Documents\GitHub\myproject"

FirstTest.js --
   [1] Starting recording promises
   Emitted | suite.before
 > Nightmare._beforeSuite
   [1] Queued | hook Nightmare._beforeSuite()
 test something
   [2] Starting recording promises
   Emitted | test.before
 > Nightmare._before
   [2] Queued | hook Nightmare._before()
   Emitted | test.start ([object Object])
   [2] Queued | amOnPage: http://yahoo.com
   [2] Queued | fillField: p, github nightmare
   [2] Queued | click: Search Web
   [2] Queued | waitForElement: #main
   [2] Queued | seeElement: #main .searchCenterMiddle li a
   [2] Queued | seeElement: //a[contains(@href,'github.com/segmentio/nightmare')]
   [2] Queued | see: segmentio/nightmare, #main li a
   [2] Queued | finish test
   Emitted | step.before (I am on page "http://yahoo.com")
 * I am on page "http://yahoo.com"
   Emitted | step.after (I am on page "http://yahoo.com")
 > log: DARLA notice: 405
 > log: instantiating uh search component
 > [URL] https://fr.yahoo.com/?p=us
 > [Code] 200
[...]

PR #263, opened 2016-10-17
do not throw error in dispatcher

Zougi:
fix issue https://github.com/Codeception/CodeceptJS/issues/262

### DIFF
--- a/lib/codecept.js
+++ b/lib/codecept.js
@@ -94,6 +94,8 @@ class Codecept {
     }
     this.mocha.run(() => {
       event.emit(event.all.result, this);
+    }).on('fail', () => {
+      event.emit(event.test.failed, this);
     }).on('end', () => {
       this.teardown();
     });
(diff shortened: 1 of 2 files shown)


Zougi:
this is not enough, because error is not catch, we should add `.on('error', callback)` on error dispatcher

Zougi:
ideally we should send error from dispatcher (see what I did initially in my first commit), then we should listen to the error in the file we first use lib/event (that would be codecept,js!?) and log the error from there, not in lib/event itself

Zougi:
my third commit is to catch fail event gotten from mocha. without this, process do not exit even if all tests finished

DavertMik:
Thanks for the report and fix. Yeah, I have failures in examples just to demonstrate that things actually are executed and actually test what they should. Thanks for noticing that hanging

DavertMik:
Looks like Yahoo page changed and it started to fail in Nightmare...

PR #275, opened 2016-10-28
Ability to set restart option as string

APshenkin:
Hi

I use process.env variables to run tests in CI to support changing config parameters

```
helpers: {
        WebDriverIO: {
            url: process.env.BASE_URL || 'https://testsite.com',
            browser: process.env.BROWSER || 'chrome',
            windowSize: process.env.SCREEN || '1360x840',
            restart: process.env.RESTART || false,
            port: 4444,
            waitForTimeout: 30000,
            desiredCapabilities: {
                "IE_ENSURE_CLEAN_SESSION": true,
                "javascriptEnabled": true,
                "phantomjs.page.customHeaders.Accept-Language": "en_GB"
            }
        }
```

but all env Variables are string, so I can't set restart option to false. Here is a small change in WebDriverIO helper to support string and boolean values for `restart` option


APshenkin:
Also added to other helpers

DavertMik:
Thanks. But can you actually use that comparison inside your config.js file?

```js
restart: process.env.RESTART === 'true',
```

APshenkin:
Yes, this will works for me, thanks!

Issue #282, opened 2016-11-08
Configuration file in different directory won't work as expected


flipace:
Hi, first - thanks for this awesome lib.

We're using [builder](https://github.com/FormidableLabs/builder) for our projects and would like to include codeceptjs in our archetype(s).

This means we'd like to put the configuration file into a folder like ```node_modules/org/archetype/config/codeceptjs/config.js```.

As far as I can see, there's no option to specify the configuration file via the command line yet (as for example with webpack --config=....)

I tried to do

```codeceptjs run node_modules/org/archetype/config/codeceptjs/``` but it won't run the tests.

It only outputs  "OK | 0 passed".

Do I have to do anything different?

**Command**:
```shell
flipace@mbpro:~/Development/project$ codeceptjs run node_modules/org/builder-archetype/config/integration-tests
CodeceptJS v0.4.8
Using test root "/Users/flipace/Development/project/node_modules/org/builder-archetype/config/integration-tests"


  OK  | 0 passed   // 0ms
```

**Config**:
```Javascript
exports.config = {
  tests: path.join(ROOT, 'src/integration-tests/build/tests.bundle.js'),
  timeout: 10000,
  output: path.join(ROOT, 'reports/integration-tests'),
  helpers: {
    WebDriverIO: {
      url: process.env.URL || 'http://localhost:3000',
      user: process.env.BROWSERSTACK_USER || 'none',
      key: process.env.BROWSERSTACK_KEY || 'key',
      browser: browser.browserName,
      desiredCapabilities: Object.assign({}, browser, {
        'browserstack.local': process.env.BROWSERSTACK_LOCAL || false
      })
    }
  },
  include: {
    I: path.join(ROOT, 'src/integration-tests/custom-steps.js')
  },
  bootstrap: false,
  mocha: {},
  name: 'someName'
};
```

DavertMik:
Please try new `--config` option implemented by @cnworks . Available since 0.4.9

flipace:
@DavertMik i think you have to publish it to npm ;)

DavertMik:
Yep, I need to test it locally and include some more fixes before that

14 лист. 2016 р. 17:16 "Patrick Neschkudla" <notifications@github.com> пише:

flipace:
Alright, I just had a chance to test the new --config command!

At first, I still experienced the same issue, except that the "Using test root" message displayed the current working directory (which seemed like an improvement :)).

[...]

PR #284, opened 2016-11-10
added option to select config file

cnworks:
Please let me know if this is ok or if anything needs to be changed.

I think it is vary useful to have different config files for different testing situations. For example, before  the deployment to a cloud env and after.

This could also solve issue #282.


cnworks:
I think the code improved a lot. Maybe the check for .js or .json could be improved.

DavertMik:
Thanks! I like this more.

DavertMik:
Thanks, I'm merging

PR #288, opened 2016-11-14
Add waitForInvisible and waitForNotAttached methods

Nighthawk14:
Add waitForInvisible and waitForNotAttached in Selenium/Webdriver.io and Protractor helpers.

Not implemented in Nightmare as I'm unfamiliar with the tool.

### DIFF
--- a/docs/webapi/waitForNotAttached.mustache
+++ b/docs/webapi/waitForNotAttached.mustache
@@ -1,0 +1,9 @@
+Waits for an element to become not attached to the DOM on a page (by default waits for 1sec).
+Element can be located by CSS or XPath.
+
+```
+I.waitForNotAttached('#popup');
+```
+
+@param locator element located by CSS|XPath|strict locator
+@param sec time seconds to wait, 1 by default
--- a/lib/helper/Protractor.js
+++ b/lib/helper/Protractor.js
@@ -224,6 +224,22 @@ class Protractor extends SeleniumWebdriver {
   }
 
   /**
+   * {{> ../webapi/waitForInvisible }}
+   */
+  waitForInvisible(locator, sec) {
+    sec = sec || 1;
+    let el = this.browser.element(guessLocator(locator) || by.css(locator));
+    return this.browser.wait(EC.invisibilityOf(el), sec*1000);
+  }
+
+  /**
+   * {{> ../webapi/waitForNotAttached }}
+   */
+  waitForNotAttached(locator, sec) {
+    return this.waitForInvisible(locator, sec);
+  }
+
+  /**
    * {{> ../webapi/waitForText }}
    */
   waitForText(text, sec, context) {
--- a/lib/helper/Protractor.js
+++ b/lib/helper/Protractor.js
@@ -233,9 +233,9 @@ class Protractor extends SeleniumWebdriver {
   }
 
   /**
-   * {{> ../webapi/waitForNotAttached }}
+   * {{> ../webapi/waitForStalenessOf }}
    */
-  waitForNotAttached(locator, sec) {
+  waitForStalenessOf(locator, sec) {
     return this.waitForInvisible(locator, sec);
   }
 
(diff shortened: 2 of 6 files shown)


docs/webapi/waitForNotAttached.mustache
```diff
@@ -0,0 +1,9 @@
+Waits for an element to become not attached to the DOM on a page (by default waits for 1sec).
+Element can be located by CSS or XPath.
+
+```
+I.waitForNotAttached('#popup');
```
DavertMik: Can you provide a better naming? By `attached` I'd think about attached file. And yes, we have `attachFile` method so this can be confusing. How about `waitForDetached` ?

lib/helper/Protractor.js
```diff
@@ -224,6 +224,22 @@ class Protractor extends SeleniumWebdriver {
...
+    sec = sec || 1;
+    let el = this.browser.element(guessLocator(locator) || by.css(locator));
+    return this.browser.wait(EC.invisibilityOf(el), sec*1000);
+  }
+
+  /**
+   * {{> ../webapi/waitForNotAttached }}
+   */
+  waitForNotAttached(locator, sec) {
+    return this.waitForInvisible(locator, sec);
```
DavertMik: so what's the difference?

lib/helper/WebDriverIO.js
```diff
@@ -992,18 +992,23 @@ class WebDriverIO extends Helper {
   /**
    * {{> ../webapi/waitForVisible }}
    */
-  waitForVisible(locator, sec) {
+  waitForVisible(locator, sec, reverse) {
```
DavertMik: please don't use boolean inside helper methods. Especially if they transform the behavior in the opposite way. It is unclear that `wait for visible "#user", 10, true` is actually waits for invisible.

lib/helper/WebDriverIO.js
```diff
@@ -992,18 +992,23 @@ class WebDriverIO extends Helper {
...
-    return this.browser.waitForVisible(withStrictLocator(locator), sec * 1000);
+    return this.browser.waitForVisible(withStrictLocator(locator), sec * 1000, reverse);
   }
 
   /**
-   * Waits for an element to become invisible on a page (by default waits for 1sec).
-   * Element can be located by CSS or XPath.
+   * {{> ../webapi/waitForInvisible }}
    */
-  waitToHide(locator, sec) {
```
DavertMik: `waitToHide` was nice. Maybe we are ok to name method like this? Also some people may use it so it's not a good idea to remove it completely

DavertMik:
Thanks! This looks good. However, there are some changes to be made for consistency.

DavertMik:
Could you do these fixes today, so I can add this PR to 0.4.9 release?

docs/webapi/waitForNotAttached.mustache
Nighthawk14: Thing is it's not `detached` by default, it could never have been attached as well. We can have `waitForStalenessOf` or `waitForNotExists`?

lib/helper/Protractor.js
Nighthawk14: There isn't, in Protractor `invisibilityOf` does hidden or not attached to the DOM. We can have `stalenessOf` which does not attached to the DOM but that would give the same result at the end.

Nighthawk14:
I made some of the changes. Just waiting for you to decide between `waitForStalenessOf` and `waitForNotExists` 👍

docs/webapi/waitForNotAttached.mustache
DavertMik: `waitForStalenessOf` sounds better to me

lib/helper/Protractor.js
DavertMik: OK, I see

DavertMik:
`waitForStalenessOf` :+1:

DavertMik:
please also remove the `reverse` flag. It's ok to copy paste things while methods do quite the opposite )

Nighthawk14:
Done, sorry I forgot this one 👍

DavertMik:
Cool! Thank you, that was really fast )

PR #292, opened 2016-11-15
Test can be substring of filename

cnworks:
#### Current Situation
If you use `codeceptjs run . test`, `test` needs to be the filename exept .js or _test.js ending. `test` and the extension needs to be equal to a filename. Therefore only one single file can be run separatly.

#### Future Situation
`test` can match any substring of any testfile that meets the configured pattern. The filter can be used more flexible and to run multiple selected files.

### DIFF
--- a/lib/codecept.js
+++ b/lib/codecept.js
@@ -90,7 +90,7 @@ class Codecept {
   run(test) {
     this.mocha.files = this.testFiles;
     if (test) {
-      this.mocha.files = this.mocha.files.filter((t) => fsPath.basename(t, '_test.js') === test || fsPath.basename(t, '.js') === test || fsPath.basename(t) === test);
+      this.mocha.files = this.mocha.files.filter((t) => fsPath.basename(t).includes(test));
     }
     this.mocha.run(() => {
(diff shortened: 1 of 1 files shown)


DavertMik:
I'm planning to remove that `.` in command. It is really confusing. Probably that will be done in 0.5 version or better in 1.0 (to break compatibility)

lib/codecept.js
```diff
@@ -90,7 +90,7 @@ class Codecept {
   run(test) {
     this.mocha.files = this.testFiles;
     if (test) {
-      this.mocha.files = this.mocha.files.filter((t) => fsPath.basename(t, '_test.js') === test || fsPath.basename(t, '.js') === test || fsPath.basename(t) === test);
+      this.mocha.files = this.mocha.files.filter((t) => fsPath.basename(t).includes(test));
```
DavertMik: but... `includes`? Can you provide something better for a filter? Maybe regex or glob patterns. Because that makes test execution unpredictable.

cnworks:
I close this PR since I use the grep functionality to fulfill the requirement I had.

PR #299, opened 2016-11-23
Set unique screenshot name for failed tests

APshenkin:
Hi!
During working I was faced with two issues with failed screenshots:
1. Scenario name can't be so long as I want, because the failed screenshot name will be also very long and will not open correctly
2. I can't use the same scenario names in different suites (files), because if they failed, then screenshot will be overrided.

So I add new option for WebdriverIO to use hashCode of title of scenario and file name for screenshot name


DavertMik:
Can you provide an example of hashed filename? 
Also will it be recognizable to which test does screenshot belong?

APshenkin:
Here is an example
`1963136161-940277181.failed.png`
There is no way to recognize to which test does screenshot belong, but if you use any html reporter You can set the same hash function to link screenshot to scenario. (I did it for mochawesome)

Please, share your thoughts, if you have any idea how we can set unique name for screenshot so that we will be able to understand to which test does screenshot belong.

DavertMik:
How about prepending testname? Just to know what it is

```
testname-1963136161-940277181.failed.png
```

APshenkin:
If we will add testname in the screenshot name then this point will not be included to this feature:
`Scenario name can't be so long as I want, because the failed screenshot name will be also very long and will not open correctly`
Maybe we can try some workaround: wor this mode, get first 30-40 characters of testname and include them in screenshot name

DavertMik:
> Maybe we can try some workaround: wor this mode, get first 30-40 characters of testname and include them in screenshot name

Yep. That would be ok.

lib/helper/WebDriverIO.js
```diff
@@ -288,7 +288,11 @@ class WebDriverIO extends Helper {
   }
 
   _failed(test) {
-    let fileName = test.title.replace(/ /g, '_') + '.failed.png';
+    let fileName = '';
+    if (this.options.hashCodeScreenshotsNames)
+            fileName = hashCode(test.title) + '-' + hashCode(test.file) + '.failed.png';
+        else
+            fileName = test.title.replace(/ /g, '_') + '.failed.png';
```
BogdanLivadariu: wouldn't it be more readable if you would use `${string interpolation}` ?

lib/helper/WebDriverIO.js
```diff
@@ -194,6 +194,7 @@ class WebDriverIO extends Helper {
       waitForTimeout: 1000, // ms
       desiredCapabilities: {},
       restart: true,
+      hashCodeScreenshotsNames: false,
```
DavertMik: better to name it `uniqueScreenshotNames` or `uniqueScreenshots`

lib/helper/WebDriverIO.js
```diff
@@ -288,7 +288,11 @@ class WebDriverIO extends Helper {
   }
 
   _failed(test) {
-    let fileName = test.title.replace(/ /g, '_') + '.failed.png';
+    let fileName = '';
+    if (this.options.hashCodeScreenshotsNames)
```
DavertMik: `{  }` - are important in `if`

lib/helper/WebDriverIO.js
```diff
@@ -1216,4 +1220,16 @@ function withStrictLocator(locator) {
   }
 }
 
+function hashCode(str) {
```
DavertMik: pleease move that to `lib/utils` with some comments about how the hashing is done and for what purpose. Because it would be hard to understand that for later users

lib/helper/WebDriverIO.js
```diff
@@ -288,7 +288,11 @@ class WebDriverIO extends Helper {
   }
 
   _failed(test) {
-    let fileName = test.title.replace(/ /g, '_') + '.failed.png';
+    let fileName = '';
+    if (this.options.hashCodeScreenshotsNames)
+            fileName = hashCode(test.title) + '-' + hashCode(test.file) + '.failed.png';
```
DavertMik: can this be `test.title.substring(0,10)` or smth like that?

APshenkin:
@DavertMik please check the changes after review and merge them :)

DavertMik:
Thanks, looks good now. I will try it locally then I will release a new version

PR #303, opened 2016-11-24
Enable asyc bootsrap file

abachar:


hassen-yahyaoui:
+1

igorlovich:
+1

DavertMik:
:+1:

lib/codecept.js
```diff
@@ -45,13 +45,18 @@ class Codecept {
...
-    this.bootstrap();
+    this.bootstrap(callback);
   }
 
   // loading bootstrap
-  bootstrap() {
+  bootstrap(callback) {
     if (this.config.bootstrap && fileExists(fsPath.join(codecept_dir, this.config.bootstrap))) {
-      require(fsPath.join(codecept_dir, this.config.bootstrap));
+      var bootsrap = require(fsPath.join(codecept_dir, this.config.bootstrap));
```
DavertMik: boot**srap**

lib/codecept.js
```diff
@@ -45,13 +45,18 @@ class Codecept {
...
+      if (typeof bootsrap === 'function') {
+          bootsrap(callback);
+      } else {
```
DavertMik: don't do else, just return

lib/command/definitions.js
```diff
@@ -36,42 +36,46 @@ module.exports = function (genPath) {
...
-        methodsOfObject(helper).forEach((action) => {
-            let actionAlias = container.translation() ? container.translation().actionAliasFor(action) : action;
-            if (!actions[actionAlias]) {
-                let params = getParamNames(helper[action]);
-                if (params) params = params.join(', ');
-                if (!params) params = '';
-                methods.push(`    ${(actionAlias)}: (${params}) => any; \n`);
-                actions[actionAlias] = 1
+    codecept.init(testsPath, function(err) {
+      if (!err) {
```
DavertMik: please don't add another nesting level this way. Probably here you can do

```js
if (err) {
   // maybe we should do smth with err here?
  return;
}
// continue basic flow
```

PR #304, opened 2016-11-24
Add wait for visible then click method to helpers

lots0logs:
This is one of the custom methods we're using in our test suite. We use it so much that I'm betting others would find it useful as well (hence this PR :smiley: )

### DIFF
--- a/lib/helper/Protractor.js
+++ b/lib/helper/Protractor.js
@@ -224,6 +224,24 @@ class Protractor extends SeleniumWebdriver {
   }
 
   /**
+   * {{> ../webapi/waitForVisibleThenClick }}
+   */
+  waitForVisibleThenClick(locators, sec, delay = 0) {
+    if ('string' === typeof locators) {
+      locators = [locators];
+    }
+
+    return locators.forEach(locator => {
+      this.waitForVisible(locator, sec);
+      this.click(locator);
+
+      if (delay > 0) {
+        this.wait(delay);
+      }
+    });
+  }
+
+  /**
    * {{> ../webapi/waitForInvisible }}
    */
   waitForInvisible(locator, sec) {
--- a/lib/helper/Protractor.js
+++ b/lib/helper/Protractor.js
@@ -224,24 +224,6 @@ class Protractor extends SeleniumWebdriver {
   }
 
   /**
-   * {{> ../webapi/waitForVisibleThenClick }}
-   */
-  waitForVisibleThenClick(locators, sec, delay = 0) {
-    if ('string' === typeof locators) {
-      locators = [locators];
-    }
-
-    return locators.forEach(locator => {
-      this.waitForVisible(locator, sec);
-      this.click(locator);
-
-      if (delay > 0) {
-        this.wait(delay);
-      }
-    });
-  }
-
-  /**
    * {{> ../webapi/waitForInvisible }}
    */
   waitForInvisible(locator, sec) {
--- a/docs/webapi/waitForVisibleThenClick.mustache
+++ b/docs/webapi/waitForVisibleThenClick.mustache
@@ -1,0 +1,19 @@
+Waits for an element to become visible on a page and then clicks the element.
+Element can be located by CSS or XPath. Accepts either a single locator string or
+an array of locator strings. The latter makes it possible to perform multi-step actions
+with a single call to this method.
+
+Wait for a single element to become visible and then click it:
+```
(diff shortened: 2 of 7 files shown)


lib/helper/Protractor.js
```diff
@@ -224,6 +224,24 @@ class Protractor extends SeleniumWebdriver {
   }
 
   /**
+   * {{> ../webapi/waitForVisibleThenClick }}
+   */
+  waitForVisibleThenClick(locators, sec, delay = 0) {
```
DavertMik: I think you can remove that from Protractor as it already inherits this method for SeleniumWebdriverJS

lots0logs: Ah, okay..wasn't sure about that! :+1:

lots0logs:
@DavertMik The tests should be passing :confused: Could you take a look and let me know if I missed something? Thanks!

DavertMik:
Sorry, can't say for sure what's happening ( I will merge this PR and debug in next days
Maybe you provide me with permissions to push to your branch?

lots0logs:
@DavertMik I've tried a few things to get the tests to pass (as you can see) but was not successful. I don't understand why they are failing. I use this exact method in all of our tests (at work) and it never fails. I'm guessing it has something to do with calling it from inside CodeceptJS container vs. calling it from directly inside a Mocha test?

DavertMik:
Thanks, I will try to investigate and update the test

DavertMik:
I found that you are not using promises right in your implementation and that's the reason tests are failing. I will try to update and fix that

DavertMik:
Sorry I couldn't make those tests pass and reverted your PR. Also I looked deeper into the code and I don't see it can work the way you expect it too. For instance, this code won't work because each of this methods returns a promise which should be handled into a promise chain

```js
  waitForVisibleThenClick(locators, sec, delay = 1) {
    if ('string' === typeof locators) {
      locators = [locators];
    }

    return new Promise((res) => {
      for (let locator of locators) {
        this.waitForVisible(locator, sec);
        this.click(locator);

        if (delay > 0) {
          this.wait(delay);
        }
      }

      return res(true);
    });
  }
```

I reworked it to work this way:

```js
  waitForVisibleThenClick(locators, sec, delay) {
    delay = delay || 1;
    if ('string' === typeof locators) {
      locators = [locators];
    }

    let p = Promise.resolve();
    for (let locator of locators) {
      p.then(() => this.waitForVisible(locator, sec));
      p.then(() => this.click(locator));

      if (delay > 0) {
        p.then(() => this.wait(delay));
      }
    }
    return p;
  }
```

however, it still doesn't work :(

---

As this action is just a combination of other methods of a helper, probably there is no real need to make it implemented inside a helper. Maybe you can add it just to `custom_steps.js` like this:

```js
module.exports = function() {
  return actor({

    waitForVisibleThenClick: function(locator) {
      this.waitForVisible(locator);
      this.click(locator);
    }
  });
}
```
More: http://codecept.io/pageobjects/#actor

The difference here is that all methods executed from `I.` (actor) are already added to global promise chain. While methods executed from helper itself know nothing about global promises.

lots0logs:
@DavertMik What are the methods inside the helper expected to return? Are the supposed to return promises? Not sure if you remember, but [when I first opened this PR](https://github.com/Codeception/CodeceptJS/pull/304/commits/d9e57e57b2b165d0e12255546ffd7f429447c16b) I wasnt using a promise at all. What you saw today was the result of my trying a few different approaches to get the test to pass.

DavertMik:
@lots0logs 

Yes, I saw your comment

> Could this be why the tests for my PR (#304) were failing? Not sure if you remember the details, but basically the implementation works …

It probably can be related. 

Not sure, I'm to sleepy to analyze that. Could you resend the PR with a test? I has left your test implementation inside a repository so you can easily pick it up. It is located inside `tests/data/app/form/wait_visisble.php`. And it is used by a test https://github.com/Codeception/CodeceptJS/blob/master/test/helper/webapi.js#L477

You can add your test and make a Pull Request. Travis will show us if it works

lots0logs:
@DavertMik I realized that it doesn't explain the failures on all helpers since this only affected SeleniumWD. That's why I deleted the comment (I'm also too tied to think straight at the moment :dizzy_face: ) I will pick it up again as soon as I have some time tho. Anyway, Happy Holidays! :christmas_tree:

DavertMik:
Merry Christmas you too! :clinking_glasses:

PR #319, opened 2016-12-08
Extend bootstrap/teardown configuration options behaviour

pscanf:
### `bootstrap` config option
Current behaviour:
- when `bootstrap` is a path to a js file, the file is `require`d. If the file exports a function, the function is called with a callback

Proposed behaviour:
- when `bootstrap` is a path to a js file, the file is `require`d. If the file exports a function, the function is called with a callback
- when `bootstrap` is a function, it's called with a callback

### `teardown` config option
Current behaviour:
- when `teardown` is a path to a js file, the file is `require`d. If the file exports a function, the function is *not* called (inconsistent with the `bootstrap` behaviour)

Proposed behaviour:
- when `teardown` is a path to a js file, the file is `require`d. If the file exports a function, the function is called (consistent with the `bootstrap` behaviour, aside from the callback which is not necessary)
- when `teardown` is a function, it's called

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -13,8 +13,18 @@ Here is an overview of available options with their defaults:
 * **helpers**: `{}` - list of enabled helpers
 * **mocha**: `{}` - mocha options, [reporters](http://codecept.io/reports/) can be configured here
 * **name**: `"tests"` - test suite name (not used)
-* **bootstrap**: `"./bootstrap.js"` - JS file to be executed before tests
-* **teardown**: - JS file to be executed after tests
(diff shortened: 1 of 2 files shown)


DavertMik:
Please make sure your changes are compatible with https://github.com/Codeception/CodeceptJS/pull/303 
I.e. we need async support for bootstrap files

DavertMik:
Yeah, looks they really are. Thanks for the PR. I think I can merge it )

pscanf:
Yeah, sorry. I had a busy day and forgot to reply. But I also confirm, they're intended to be compatible. Thanks for the merge. 🙂

PR #328, opened 2016-12-14
Fix for waitForElement of SeleniumWebdriver helper

VaderZ:
Fixes "Invalid Locator" exception on calling waitForElement method of SeleniumWebdriver helper.


DavertMik:
Thanks, I will need some time to check it and then I will merge

lots0logs:
@DavertMik Could this be why the tests for my PR (#304) were failing? Not sure if you remember the details, but basically the implementation works fine in the real world but we couldnt get the tests to pass for some reason

PR #340, opened 2016-12-22
Added flaky mode for Feature

tfiwm:
Proposal for flaky mode in Features

I my opinion this should be enough because the problems with flaky states is almost when an emulator or an browser at selenium / cloud based system like browserstack, saucelab, ... is not starting. Then the test should be retried.

It shouldn't be used to mark tests as flaky because i think in this case the test was not correct implemented.

That's why i think in Feature it should be enough but i am open for any other ideas / problems ;)

### DIFF
--- a/lib/interfaces/bdd.js
+++ b/lib/interfaces/bdd.js
@@ -24,7 +24,7 @@ var escapeRe = require('escape-string-regexp');
 module.exports = function (suite) {
   var suites = [suite];
   suite.timeout(0);
-  
+
   suite.on('pre-require', function (context, file, mocha) {
     var common = require('mocha/lib/interfaces/common')(suites, context);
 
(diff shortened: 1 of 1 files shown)


DavertMik:
...that simple... wow!

lib/interfaces/bdd.js
```diff
@@ -40,11 +40,14 @@ module.exports = function (suite) {
      * and/or tests.
      */
 
-    context.Feature = function (title) {
+    context.Feature = function (title, retries) {
```
DavertMik: I'd make pass it as hash: `{ retries: number }`
probably we will use it in future for other configuration options

DavertMik:
> It shouldn't be used to mark tests as flaky because i think in this case the test was not correct implemented.

Sorry but what do you mean? What tests on what conditions should be marked as flacky?

Also on webdriverio website there are also options to set retries for specific test and before/after hooks. What do you think about these options, how useful they are.

DavertMik:
Thanks, I'm merging. But still I'd like to continue the discussion. 
As I said I will change the APi to use `{retries: x}` in implementation

tfiwm:
Sorry I was on long vacation ;)

we can integrate it in the same way. the only think what we have to do is to bind the mocha test context to the scenario. At the moment it is empty and should make problems or do you have any plans with the scenario instance self? Is it possible to bind the mocha test context?

Issue #274, opened 2016-10-28
Issue with I.resizeWindow when using Nightmare


SuneRadich:
It seems that a recent release of Nightmare to 2.8.1 causes my test to fail (worked fine with Nightmare 2.7.0). Running on Node 6.9.1.

In a very simple test, where I have this before block

```
Before((I) => {
    I.resizeWindow(1024, 768);
});
```

The test fails like this

```
Search --
 ✖ "before each" hook: Before for "Can search" in 309ms
(node:6279) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): TypeError: this.browser.manage is not a function
(node:6279) PromiseRejectionHandledWarning: Promise rejection was handled asynchronously (rejection id: 1)

-- FAILURES:

  1) Search "before each" hook: Before for "Can search":
     TypeError: this.browser.manage is not a function
      at Nightmare.resizeWindow (/Users/srv/.nvm/versions/node/v6.9.1/lib/node_modules/codeceptjs/lib/helper/Nightmare.js:457:25)
      at Step.run (/Users/srv/.nvm/versions/node/v6.9.1/lib/node_modules/codeceptjs/lib/step.js:26:47)
      at add (/Users/srv/.nvm/versions/node/v6.9.1/lib/node_modules/codeceptjs/lib/recorder.js:81:35)
      at process._tickCallback (internal/process/next_tick.js:103:7)
```

From what I can see, both electron and nightmare have released updated versions since I wrote my test, but it seems that this is something in Nightmare that have changed, and that the Nightmare helper now fails.

PR #341, opened 2016-12-22
Fix for Issue #274 with I.resizeWindow (nightmare)

norisk-it:
Fix for issue #274. Didn't find a way for maximize to work, but this should help to at least resize windows.


DavertMik:
Thank you! I didn't have time to look for that issue, it appeared to be very tiny but still I didn't make it. Thanks for resolving.

PR #350, opened 2016-12-29
Nightmare restart option

nairvijays99:
Implemented restart option to run suites in a single instance of the browser. This feature is configuration driven similar to WD and Protractor.


nairvijays99:
All checks passed.

PR #352, opened 2016-12-29
Nightmare restart

nairvijays99:
`restart` option for tests run using nightmare helper. Defaults to
`true`. Setting the option `false` with run all the scenario’s of a
suite in a single browser instance.

### DIFF
--- a/lib/helper/Nightmare.js
+++ b/lib/helper/Nightmare.js
@@ -46,7 +46,8 @@ class Nightmare extends Helper {
     this.options = {
       waitForAction: 500,
       waitForTimeout: 1000,
-      rootElement: 'body'
+      rootElement: 'body',
+      restart: true
     };
 
     // override defaults with config
@@ -156,7 +157,30 @@ class Nightmare extends Helper {
     });
   }
 
+  _beforeSuite() {
+    if (!this.options.restart)
+      this._startBrowser();
+  }
+
   _before() {
+    if (this.options.restart)
+        this._startBrowser();
+    return this.browser;
+  }
+
+  _after() {
+    if (this.options.restart)
+        return this._stopBrowser();
+    else
+        return this.browser.cookies.clearAll();
+  }
+
+  _afterSuite() {
+    if (!this.options.restart)
+        this._stopBrowser();
+  }
+
+  _startBrowser() {
     this.browser = this.Nightmare(this.options);
     this.browser.on('dom-ready', () => this._injectClientScripts());
     this.browser.on('console', (type, message) => {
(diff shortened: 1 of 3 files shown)


nairvijays99:
We have a suite with multiple scenarios revolving the same page. Suites are slow because a new browser instance is called before every test. This option is already available for Webdriver, Selenium and Protractor. Can this be reviewed?

lib/helper/Nightmare.js
```diff
@@ -46,7 +46,8 @@ class Nightmare extends Helper {
     this.options = {
       waitForAction: 500,
       waitForTimeout: 1000,
-      rootElement: 'body'
+      rootElement: 'body',
+      restart: true
```
nairvijays99: Defaults to truthy.

lib/helper/Nightmare.js
```diff
@@ -156,7 +157,30 @@ class Nightmare extends Helper {
     });
   }
 
+  _beforeSuite() {
+    if (!this.options.restart)
+      this._startBrowser();
```
nairvijays99: Creates new instance, is restart is falsy. Will not impact existing functionality.

lib/helper/Nightmare.js
```diff
@@ -156,7 +157,30 @@ class Nightmare extends Helper {
...
+  }
+
   _before() {
+    if (this.options.restart)
+        this._startBrowser();
```
nairvijays99: Retains default functionality of new browser for every test.

lib/helper/Nightmare.js
```diff
@@ -156,7 +157,30 @@ class Nightmare extends Helper {
...
+    return this.browser;
+  }
+
+  _after() {
+    if (this.options.restart)
+        return this._stopBrowser();
```
nairvijays99: Retains default functionality of killing current browser for every test.

lib/helper/Nightmare.js
```diff
@@ -156,7 +157,30 @@ class Nightmare extends Helper {
...
+    else
+        return this.browser.cookies.clearAll();
```
nairvijays99: Clear the cookies before handing it over to the next test. Should this just return the browser without clearing it? I don't know.

lib/helper/Nightmare.js
```diff
@@ -156,7 +157,30 @@ class Nightmare extends Helper {
...
+  }
+
+  _afterSuite() {
+    if (!this.options.restart)
+        this._stopBrowser();
+  }
```
nairvijays99: Kill the browser after every suite, if restart is falsy

DavertMik:
Thanks for comprehensive PR and comments. 
My only proposal to include comments within code next time. But I like the implementation. Thanks for your work!

nairvijays99:
Sure. Will keep that in mind. Is the master branch (0.4.14) published?

Issue #347, opened 2016-12-27
Scenario.skip causes 'Cannot read property retries of undefined'


hubidu:
#### What are you trying to achieve?

I have a **Scenario.skip** call in a feature file.

#### What do you get instead?

When I run the test with version 0.4.13 it fails with the following error:

```
Cannot read property 'retries' of undefined
27-Dec-2016 10:30:37	
27-Dec-2016 10:30:37	TypeError: 
27-Dec-2016 10:30:37	    at context.Scenario (/var/atlassian/application-data/bamboo/xml-data/build-dir/VEVE-UT-JOB1/node_modules/codeceptjs/lib/interfaces/bdd.js:102:15)
27-Dec-2016 10:30:37	    at Function.context.xScenario.context.Scenario.skip (/var/atlassian/application-data/bamboo/xml-data/build-dir/VEVE-UT-JOB1/node_modules/codeceptjs/lib/interfaces/bdd.js:122:15)
```

#### Cause

Scenario.skip calls Scenario, but only with the title parameter. **The opts parameter is undefined**. Therefore it fails on line 102 in bdd.js.

There should probably be default parameters in Scenario.

DavertMik:
Yes, thank you, this should be updated

PR #359, opened 2017-01-11
Pass empty opts object to test when skipping

MasterOfPoppets:
We came across this bug also and decided to fix it to get us moving - so I made a pull request for you!

#347


DavertMik:
Thank you!

PR #363, opened 2017-01-13
Ignore errors from trying to delete non-existent file

lennym:
Test fails when run for the first time on a new clone because it attempts to delete a file that does not exist (it is created by the same test).

Additionally the same test is currently failing on the `master` travis build - https://travis-ci.org/Codeception/CodeceptJS/builds/191168878


DavertMik:
Yep, thanks, and noticed this and fixed just the same way )

PR #364, opened 2017-01-13
Apply an assertion to existence of definition file

lennym:
`fs.existsSync` will not throw if the file doesn't exist (it would probably not be much use if it did) so an assertion needs to be added to the return value for a meaningful test.


DavertMik:
Thanks

PR #365, opened 2017-01-13
Update guppy-pre-commit version

lennym:
Previous version installed a pre-commit hook which required a global install of gulp to be present. New version uses the locally installed binary.

See https://github.com/therealklanni/guppy-cli/commit/7ddf899b6943e46a328f79c8d24dbd37f015d7c1 (Note: `guppy-cli` version was updated in `guppy-pre-commit@0.4.0`)


DavertMik:
Thanks!

PR #366, opened 2017-01-13
Use `path.resolve` in favour of `path.join`

### LINKED ISSUE #361: Cannot provide absolute file paths for `tests` in configuration

Pass the location of my tests as an absolute file path.

#### What do you get instead?

Because the code [here](https://github.com/Codeception/CodeceptJS/blob/master/lib/codecept.js#L64) uses `path.join` instead of `path.resolve` then the test path is concatenated to the working directory if an absolute path is provided.

Happy to raise a PR to change the code above, but I've only been looking at codeceptjs for literally an hour, so will be highly likely to miss something somewhere else. Specifically, if modifying to allow absolute paths here then it probably makes sense to do the same for other places where a file path is provided as a point of config.

lennym:
Using `join` to attach the configured test path to the project root prevents tests being loaded from an absolute path. Using `resolve` instead allows both relative and absolute paths to be used.

Fixes #361

### DIFF
--- a/lib/codecept.js
+++ b/lib/codecept.js
@@ -61,7 +61,7 @@ class Codecept {
 
   loadTests(pattern) {
     pattern = pattern || this.config.tests;
-    glob.sync(fsPath.join(codecept_dir, pattern)).forEach((file) => {
+    glob.sync(fsPath.resolve(codecept_dir, pattern)).forEach((file) => {
       this.testFiles.push(fsPath.resolve(file));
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks for the solution

lennym:
No problem. I spent a bit more time looking around inside the project (as you can probably tell from your notifications!) so felt confident enough to offer a solution.

PR #381, opened 2017-01-20
fix for _test() helpersHook not working

cjhille:
Could this be right? (I know, I f*d up the commit message ..) ¯\_(ツ)_/¯


DavertMik:
Yes. good catch. Thanks, that was the problem with that hook.

PR #388, opened 2017-01-24
Fix 'fullTitle() is not a function' in exit.js

hubidu:
To reproduce create a test suite with beforeAll or beforeEach hooks. Then
run the tests without previously starting selenium/phantomjs. You should see the above error message which is also misleading since the actual error is 

```
 [1] Error | Error: Couldn't connect to selenium server
```


DavertMik:
Thanks, looks good to me.

PR #389, opened 2017-01-24
Added yarn and updated deps

KennyRules:
* Added [Yarn](https://yarnpkg.com/) lockfile.
* Ran `yarn upgrade-interactive` and updated:
    * glob
    * inquirer
    * chai-as-promised
    * git-guppy
    * gulp-documentation
    * gulp-istanbul
    * gulp-mocha
    * selenium-webdriver

All tests have passed with 0 code changes! :D


KennyRules:
New protractor wasn't happy with travis - but that might be fixable by upping the node version in travis.yml. But I'll keep this commit laser-focused :)

KennyRules:
Looks like I missed something when checkout out travis/this repo - apologies!

DavertMik:
Yes, something strange happens during installation.
I think Travis should still prefer npm as installer but looks like its not. 
To be honest I don't think we need .lock file, as there are plenty of users who won't use yarn and they will get all the updated deps. If we don't track change in dependencies and stick to those set to yarn.lock we won't be notified of new errors.

PR #391, opened 2017-01-24
Fix for Nightmare waitTimeout. Resolves #236

### LINKED ISSUE #236: Wait For text timeout not working using nightmare helper
With the following scenario, the timeout is failing only after 30000msec 

```js
Scenario('Timeout  failing', (I) => {
  I.amOnPage('http://www.github.com');
  I.see('Welcome home');
  I.click('.header-logo-invertocat');
  I.waitForText('Priced for everyone', 2); // should fail after 2 seconds
});
```

```
codeceptjs run --steps 

-- FAILURES:

  1) login: Login with user list:
     .wait() timed out after 30000msec
```

Update: If I run 
I.wait(1); I.see('Priced for everyone');
then it works.

but if I run only the 
I.see('Priced for everyone')
then i got:

(node:47217) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): TypeError: Cannot create property 'stack' on string 'codeceptjs is not defined' ^C%

HughZurname:
Hi,

I spotted an issue with the waitFor functions in the Nightmare helper as mentioned in #236. 

In cases where the element you're waiting for takes longer than 30 seconds to appear, the default nightmare timeout is hit. The reason this is happening is that `optionWaitTimeout` isn't valid in Nightmare and was causing it to fall back to it's default of 30 seconds.

I've changed this to correctly set the Nightmare `options.waitTimeout` value. I also removed the /1000 on `waitForTimeout` as Nightmare requires milliuseconds and the functions below are the only place this was being used.

I added a test for for the timeout function on `waitForText`

### DIFF
--- a/lib/helper/Nightmare.js
+++ b/lib/helper/Nightmare.js
@@ -58,7 +58,7 @@ class Nightmare extends Helper {
     Object.assign(this.options, config);
 
     this.context = this.options.rootElement;
-    this.options.waitForTimeout /= 1000; // convert to seconds
+    this.options.waitForTimeout;
   }
 
   static _config() {
@@ -759,7 +759,7 @@ class Nightmare extends Helper {
       context = this.context;
     }
     let locator = guessLocator(context) || { css: context};
-    this.browser.optionWaitTimeout = sec * 1000 || this.options.waitForTimeout;
+    this.browser.options.waitTimeout = sec * 1000 || this.options.waitForTimeout;
 
     return this.browser.wait(function (by, locator, text) {
       return codeceptjs.findElement(by, locator).innerText.indexOf(text) > -1;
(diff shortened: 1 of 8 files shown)


lib/helper/Nightmare.js
```diff
@@ -759,7 +759,7 @@ class Nightmare extends Helper {
       context = this.context;
     }
     let locator = guessLocator(context) || { css: context};
-    this.browser.optionWaitTimeout = sec * 1000 || this.options.waitForTimeout;
+    this.browser.options.waitTimeout = sec * 1000 || this.options.waitForTimeout;
```
DavertMik: yeah, i see the bug. It was due different implementations in WebDriverIO, SeleniumWebDriver and Nightmare

test/helper/webapi.js
```diff
@@ -502,6 +502,13 @@ module.exports.tests = function() {
         .then(() => I.waitForText('Dynamic text', 2, '#text'))
         .then(() => I.see('Dynamic text'));
     });
+
+    it('should wait for text after timeout', () => {
```
DavertMik: thanks for the test!

HughZurname:
No worries, happy to help! :)

I've pushed some updates to make the test work. It was weirdly failing on the first commit but not returning an exit code of 1 from mocha so the build was passing.

DavertMik:
Thank you, getting it merged!

PR #398, opened 2017-01-27
added an uncheckOption method similar to the checkOption method

thawk55:


DavertMik:
If this method is 90% the same as checkOption, could they be moved to a private function which takes an extra parameter?

thawk55:
Made a private function.

PR #404, opened 2017-02-04
Dockerised CodeceptJS

artiomnist:
Created several Dockerfiles for CodeceptJS to allow executing of CodeceptJS tests within a docker container. Currently only for Nightmare and WebDriverIO Drivers. 

**Base** image comes with Node 6.9.1 & scripts.

**WebDriverIOBase** builds on top of `Base` and comes with latest WebDriverIO 
**NightmareBase** builds on top of `Base` and comes with necessaries for nightmare (Electron) & an entrypoint for nightmares XVFB

**WebDriverIO** Builds on top of `WebDriverIOBase` and comes with the latest version of CodeceptJS
**Nightmare** Builds on top of `NightmareBase` and comes with the latest version of CodeceptJS

I have provided running and building instructions within `README.md` for each Dockerfile.

Once this has been tested and approved, I suggest creating a Docker hub account for CodeceptJS, (If not already created). Once created, the Images can be built and pushed to the docker hub registry for public use. 

(P.S: Massive fan of your work! Thought I may try and contribute to this awesome project)


DavertMik:
Awesome!

docker/webDriverIOCodeceptjs/README.md
```diff
@@ -0,0 +1,61 @@
+# Codeceptjs Nightmare
```
DavertMik: ups )

artiomnist: Thanks 😄 , 

Great, we can use that organization docker hub to hold the image.

I see, I mainly created the Dockerfiles this way to keep them as light as possible, and perhaps easier to maintain. However, I can see that If a WebDriverIO changes, then subsequent images must be rebuilt. Likewise, if a new version of CodeceptJS is released, several images will have to be built and pushed. 

I have merged all Dockerfiles into one. It comes with Node, Nightmare, WebDriverIO and now Protractor. Additionally, I have updated the `README.md` to include running and building instructions.

_Note: I've kept the `npm install ...` commands separate to allow docker to cache them if there are no changes to them. For example, If there is a new version of CodeceptJS the build will not have to reinstall WebDriverIO or Nightmare or Protractor etc, it can just used the cached layer. Overall, this should speed up image build times._

artiomnist:
Thanks 😄 ,

Great, we can use that organization docker hub to hold the image.

I see, I mainly created the Dockerfiles this way to keep them as light as possible, and perhaps easier to maintain. However, I can see that If a WebDriverIO changes, then subsequent images must be rebuilt. Likewise, if a new version of CodeceptJS is released, several images will have to be built and pushed.

I have merged all Dockerfiles into one. It comes with Node, Nightmare, WebDriverIO and now Protractor. Additionally, I have updated the README.md to include running and building instructions.

Note: I've kept the npm install ... commands separate to allow docker to cache the layer if there are no changes to them. For example, If there is a new version of CodeceptJS the build will not have to reinstall WebDriverIO or Nightmare or Protractor etc, it can just use the cached layer. Overall, this should speed up image build times.

DavertMik:
Thank you, looks really good. I will merge it and try it in action, and make it described as official install process in documentation. I'm not sure if DockerHub can handle Dockerfile to be located outside of root directory, so there is a chance I will move it to the root of a project.

artiomnist:
Great!

If you need any help setting it up let me know! I think Dockerhub does require the Dockerfile to be in the root... If this is the case, It might be worth creating a `scripts` directory for the docker container scripts. The line that copies the in the scripts and entry point would need some adjusting. 

Something like: 
```		
		# Copy across the scripts
COPY 	/scripts/*.sh /scripts/

		# Copy accross the entrypoint for Nightmare
COPY	/scripts/entrypoint /

```

DavertMik:
> Dockerhub does require the Dockerfile to be in the root

Yes, it does.... Thanks, I will update Dockerfile

DavertMik:
Also I don't agree that CodeceptJS should be places inside `tests` directory. The `codecept.json` should be placed into the root but set a path inside a config

DavertMik:
Oh, on another thought it is better not to install everything globally, but use CodeceptJS from this exact repository, just with:

```
npm install
```

artiomnist:
The `tests` directory is just the users workspace. (Perhaps we could rename it to `workspace`) Within this directory the user can have their test scripts wherever they want as long as they are specifying this within their `codecept.json` or `codecept.conf.js`

DavertMik:
> The tests directory is just the users workspace

Ah, yeah, sorry, I understood that. Yep, you are right.

artiomnist:
> Oh, on another thought it is better not to install everything globally, but use CodeceptJS from this exact repository, just with: ```npm install```

Yes, this could work. It would mean the Dockerfile would need to copy in the necessary files to build / install CodeceptJS and run `npm install`.   
Alternativly, the Dockerfile could clone in the repo (via a deploy key) and run `npm install`.

DavertMik:
Btw, `--link` is depreacted, so I don't think it is recommended to use it. Probably in Docker compose should be used

DavertMik:
Available!
```
docker pulll codeception/codeceptjs
```

artiomnist:
> Probably in Docker compose should be used

Yes, It is probably a better idea to use docker-compose to link the containers. I'll work on a docker-compose file soon.

DavertMik:
I think I have one, You can try to test it:

```yaml
version: '2'
services:
  codeceptjs:
    image: codeception/codeceptjs
    depends_on:
      - firefox
      - web
    volumes:
      - .:/tests
  web:
    image: node
    command: node app/server.js
    volumes:
      - .:/app
  firefox:
    image: selenium/standalone-firefox-debug:2.53.0
    ports:
      - '4444'
      - '5900'

```

PR #406, opened 2017-02-06
Issue with 'grabAttributeFrom' method

robrkerr:
Needed to use 'getAttribute' to grab an attribute from an element - it was just returning 'null' for me otherwise.


robrkerr:
Sorry - still getting the hang of contributing like this on Github - see new pull request #407

PR #407, opened 2017-02-06
Issue with 'grabAttributeFrom' method

robrkerr:
Needed to use 'getAttribute' to grab an attribute from an element - it was just returning 'null' for me otherwise.


DavertMik:
We need a test for this. In #408 I added instructions for test execution. Could you write a test for this feature? Thanks

robrkerr:
Added a test that this pull request fixes. 

In doing so, I realised that the issue was only with custom attributes - that's why the existing test was passing but I was still having issues. This stackoverflow answer explains what was going on: http://stackoverflow.com/questions/10280250/getattribute-versus-element-object-properties/10280487#10280487

DavertMik:
Thanks for the deep investigation and the fix. Very nice to have this edge cases fixed!

robrkerr:
No problem :)

Issue #295, opened 2016-11-17
UnhandledPromiseRejectionWarning error when selecting the dropdown using "selectOption" 


magimenon:
#### What are you trying to achieve?
I'm trying to select an option from the dropdown. Name of my <select>  is "search" and I select a value "new". I get a below error:

UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Failed to execute 'iterateNext' on 'XPathResult': The document has mutated since the result was returned.

One thing to be noted here is when I select the first dropdown there are other dropdown gets populated (ajax). I googled and couldn't figure out what's going wrong here?

#### What do you get instead?

UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Failed to execute 'iterateNext' on 'XPathResult': The document has mutated since the result was returned.

sushan088:
I am having the same problem. Previously it works fine with Webdriverio and phantomjs but after using nightmare, not working in that as well.

DavertMik:
Could you provide HTML which reproduces this error? 

Also what helper is that? Nightmare or webriverio?

magimenon:
Im using nightmare & node version 6.9.1

The HTML is - 

<form name="ContactUsForm">
    <p class="required"><span>*</span> = required</p>
    <fieldset repeat="content.form.fieldsets" name="messageInfo">
        <legend if="legend">Message</legend>
        <div each="fields" template="">
            <label class="  ">
                <span>Subject</span>
                <input type="text" data-which-field="" name="contact_subject" value="" maxlength="100" class="valid">
            </label>
            <label class="  ">
                <span>Message Text</span>
                <textarea name="customer:comments" wrap="" maxlength="2000"></textarea>
            </label>
        </div>
    </fieldset>
    <fieldset repeat="content.form.fieldsets" name="contactInfo">
        <legend if="legend">Contact Information</legend>
        <div each="fields" template="">
            <label class=" required ">
                <span>First Name</span>
                <input type="text" data-which-field="" name="contact:name:first" value="" maxlength="100" required="">
            </label>
            <label class=" required ">
                <span>Last Name</span>
                <input type="text" data-which-field="" name="contact:name:last" value="" maxlength="100" required="">
            </label>
            <label class="required ">
                <span if="label">Preferred Contact</span>
                <select name="preferred_contact" data-which-field="" required="" fieldtype="">
                    <option value="email" repeat="values">Email</option>
[...]

PR #408, opened 2017-02-06
Possible fix for a 'selectOption' issue

robrkerr:
I had what appears to be the same issue as in #295, where the select element was correctly found and the correct option selected but then this error was thrown: 
```
Failed to execute 'iterateNext' on 'XPathResult': The document has mutated since the result was returned.
```

So I changed the implementation slightly so that it no longer cycles through different options - it just selects the first one that matches. 

Note: this won't work for selecting multiple options - it will only work for a dropdown select. But given this is the more common one I thought it was much more crucial it work for this. Obviously a better solution would be for it to work on both but I don't know enough about what's going on to work that out.


DavertMik:
Looks like it broke current `selectOption` tests. To reproduce them and fix run:

```
# start a server with demo app
# yes, you need PHP for that
php -S 127.0.0.1:8000 -t test/data/app
mocha tests/helpers/Nightmare_test.js --grep "selectOption"

```

sqmgh:
The tests are likely failing because he changed/broke current behavior.

The error that is being worked around is pretty 'standard' for iterators. Just pushing the results into an array, and then looping through that instead would fix things.

Hacked together like:
```
      var current = null;
      var items = [];
      while (current = found.iterateNext()) {
        items.push(current);
      }
      for (var i = 0; i < items.length; items++) {
        current = items[i];
```

Of course, I don't know the exact behavior that is expected by the test. This method of fixing things doesn't really mimic the exact behavior of a user in all cases. 
There could be a scenario where clicking one option could remove another from the select. Removing the user's ability to actually click the second item. If that's important, the fix would need to be slightly more complicated. 

Forgive me for not posting a proper PR for this. I stumbled here searching for the error message while previewing CodeceptJS to see if it was suitable for my project needs. I have neither the time or inclination to get php and a testing environment set up to do things properly. However, the code change I have posted did unblock me from running my specific sample test.

robrkerr:
Ok - I worked out that the error was occurring due to changes to the DOM due to an 'onchange' event listener. I've recreated this in a test and following the suggested code changes from @sqmgh (thanks!) both this new test and the existing ones are passing.

DavertMik:
Thank you for the deep investigation and a good fix :+1:

PR #414, opened 2017-02-10
[Command error] error: unknown option `--kind'

battlesteel:
To create different kinds of pageobject by running the command "codeceptjs go --kind step" output the following error:
error: unknown option `--kind'

The option "--kind" should be use "--type", eg: "codeceptjs go --type step"


DavertMik:
Thanks for noticing!

PR #416, opened 2017-02-12
Fixing typos

kostasx:


DavertMik:
Thanks for the fixes! The `gh-pages` branch is not updated manually but taken from the documentation at `master` branch. Guides are listed in `docs` directory and documentation from helper is taken from docblocks of source files.

So could you update following files instead?

https://github.com/Codeception/CodeceptJS/blob/master/docs/nightmare.md
https://github.com/Codeception/CodeceptJS/blob/master/lib/helper/Nightmare.js

Thanks

kostasx:
Sure!

PR #419, opened 2017-02-13
Fixed codecept.js is undefined

dtaylor7:
A bug was occurring: `codeceptjs is undefined` when running tests.

The reason for this was because the javascript injected into the page was happening after codecept attempted to run tests which required it.

This is a temporary fix which adds a global flag which is set to True when the Nightmare browser has loaded the page and injected the codecept javascript. The tests are delayed until both of these conditions are met.

This is not a full proof fix ( and only solves issue for nightmare ) and the only reason we have used a Global variable to fix the issue was due to time.

However hopefully it illustrates the issue and points to a possible solution.

### DIFF
--- a/lib/actor.js
+++ b/lib/actor.js
@@ -55,6 +55,40 @@ function recordStep(step, args) {
   let task = `${step.name}: ${step.humanizeArgs()}`;
   let val;
 
+  // check to ensure browser is loaded
+  recorder.add('wait-for-browser-ready', function(){
+    return new Promise((resolve) => {
+
+      if (step.helper.isReady || step.helper.isReady === undefined) {
+        resolve();
+      } else {
+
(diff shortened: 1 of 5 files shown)


DavertMik:
Thanks. Unfortunately this breaks all the stuff (accroding to Travis) but at least I can try to dig deeper into the problem and investigate it.

dtaylor7:
Yea i didn't look at tests... I have rewritten it as we had some problems, but this is a little cleaner now... and moving towards a production solution. I would suggest hooking into your events for a fully integrated solution.

Rob-Doherty:
Has there been any conclusion with this? Would be a great help for us too

APshenkin:
Not actual after 1.0 release

PR #420, opened 2017-02-13
Update Protractor.js

Jeroenhuisjes:
This will fix the issue of the Protractor helper always sending Firefox as the browser to Browserstack. Before this even if you said browserName: 'Chrome' it would startup Firefox in Browserstack. With this change it will boot the right browser.


DavertMik:
Thank you!

DavertMik:
However, your fix breaks tests. Could you look into them and fix? https://travis-ci.org/Codeception/CodeceptJS/jobs/201278629

browser name should be converted to string

Jeroenhuisjes:
Ah yes I see! Let me check tomorrow at work 👍 For now this works form me, but let's make it work for real.

Issue #423, opened 2017-02-15
Get error when using I.pressKey


romanovma:
#### What are you trying to achieve?

run tests with `I.pressKey`

#### What do you get instead?

Get the error:

```bash
Cannot read property 'Key' of undefined
```
* CodeceptJS version: 0.5.0
* Protractor: 5.1.1
* Mac OS

```js
helpers: {
    'Protractor': {
      url: 'http://localhost:4200/',
      driver: 'hosted',
      browser: 'chrome',
      rootElement: 'app-root',
      useAllAngular2AppRoots: true,
      restart: true,
      timeout: 50000
    }
  },

DavertMik:
thanks, i will check and fix that

PR #431, opened 2017-02-23
revert SeleniumWebDriver init inside Protractor.js helper from commit…

romanovma:
… 3987de3e5145335147b713fe256fede3369e2483

#423

### DIFF
--- a/lib/helper/Protractor.js
+++ b/lib/helper/Protractor.js
@@ -87,6 +87,14 @@ class Protractor extends SeleniumWebdriver {
     Runner = require('protractor/built/runner').Runner;
     By = require('protractor').ProtractorBy;
     this.isProtractor5 = !require('protractor').wrapDriver;
+
+    try {
+      // get selenium-webdriver
+       this.webdriver = requireg('selenium-webdriver');
+     } catch (e) {
+       // maybe it is installed as protractor dependency?
+       this.webdriver = requireg('protractor/node_modules/selenium-webdriver');
+     }
   }
 
   static _checkRequirements() {


romanovma:
Or may be we can just call
```super._init()``` 
from Protractor's _init method.
Could you check and let me know if I should update the pull request.

**Comment by DavertMik** [comment]
Sorry, could you explain how it is related to `I.pressKey` issue?

**Comment by romanovma** [comment]
Yep. Protractor helper inherits `pressKey` from SeleniumWebDriver so the following code from 'SeleniumWebdriver.js' is run:
```
pressKey(key) {
    let modifier;
    if (Array.isArray(key) && ~['Control', 'Command', 'Shift', 'Alt'].indexOf(key[0])) {
      modifier = this.webdriver.Key[key[0].toUpperCase()];
      key = key[1];
    }

    // guess special key in Selenium Webdriver list
    if (this.webdriver.Key[key.toUpperCase()]) {
      key = this.webdriver.Key[key.toUpperCase()];
    }

    let action = new this.webdriver.ActionSequence(this.browser);
    if (modifier) action.keyDown(modifier);
    action.sendKeys(key);
    if (modifier) action.keyUp(modifier);
    return action.perform();
  }
```

Once it hits any of rows with `this.webdriver.Key` it throws `Cannot read property 'Key' of undefined` which is because `this.webdriver` is `undefined`. 

Then we can see that `this.webdriver` is defined inside `SeleniumWebDrive`'s `_init` method so we can define `this.webdriver` by one of the following options: 1) doing it manually inside `Protractor`'s `_init` method (which was done before update to Protractor ver5 btw) or 2) just calling parent `_init` method which is `SeleniumWebDrive`'s `_init` method. First is more safe as I don't know the whole picture that's why I proposed it rather than the second one.

DavertMik:
> . First is more safe as I don't know the whole picture that's why I proposed it rather than the second one.

Yes, it's more safe. I was thinking maybe a webdriver instance can also be retirived from Protractor library, but now sure, as Protractor regularely changes its API

DavertMik:
Thanks. I think this is the best solution for now

PR #439, opened 2017-03-06
Run several instances of codeceptJS

APshenkin:
New feature is most for CI run. Usual, You want to run several suites for several browsers with different window size. Previously for this you have to create several jobs in your CI, or create bash script to run several suites (but they will performed sequentially).
Now there is several options to run suites in parallel with one command.

Now there is `run:multiple --multiple` option to set what suites to run.

First, create section in your codeceptjs conf like this:
```
multiple: {
        chrome: {
            smoke: {
                grep: '@smoke28',
                windowSizes: ['maximize','1200x840']
            },
            regression: {
                grep: '@smoke29'
            }
        },
        firefox: {
            smoke: {
                grep: '@smoke28',
                windowSizes: ['maximize','1200x840']
            },
            regression: {
                grep: '@smoke29'
            }
        }
    },
```

for Each browser you can configure suites. First param in Array is `grep` parameter to search tests, second one is windowsSizes to launch tests (e.g. if you set two window sizes for smoke, then this suite will launch 2 times, if not set - will set default value from config)

now you can try to run several test cases. Here are examles to run:

`codeceptjs run:multiple --multiple 'chrome: smoke; firefox: regression'` - will launch smoke suite in chrome and regression in firefox
`codeceptjs run:multiple --multiple 'chrome: smoke; regression'` - will launch smoke suite in chrome and regression in firefox and chrome
`codeceptjs run:multiple --multiple 'all'` - will launch all suites
`codeceptjs run:multiple --multiple 'smoke, regression'` - will launch smoke and regression in firefox and chrome

You can combine launch parameters as you want.

When running multiple you will see suite tag like [ browser_suite_windowSize ] in console for each test (recomended to set steps to false, to provide readable logs)

Also output folder for screenshots will be different for suites: [output_param_form codeceptjs_conf]/browser_suite_windowSize.

Also if you use mochawesome html reporter, the output folder will be also changed


DavertMik:
JUST WOW :+1:

lib/command/run.js
```diff
@@ -18,17 +19,156 @@ module.exports = function (suite, test, options) {
...
+    }
+  }
+
+
+  function runnerHelper(browser, suite, config) {
+    if (Object.prototype.toString.call(config.multiple[browser][suite]) === '[object Array]') {
+      let suiteConf = config.multiple[browser][suite];
+      let sizes = (suiteConf[1] === undefined) ? ['default'] : suiteConf[1].replace(/ /g, "").split(',')
+      // override browser
+      let override_br = replaceValue(new getConfig(testRoot, configFile), 'browser', (browser === 'ie') ? 'internet explorer' : browser)
```
DavertMik: please give a better name to this variable
`overriddenConfig` would be ok

lib/command/run.js
```diff
@@ -18,17 +19,156 @@ module.exports = function (suite, test, options) {
...
+
+        }
+      } else if (browserSuites.length == 1 && browserSuites[0] != '') {
+
+        //run suite in all browsers
+        let suites = browserSuites[0].split(',');
+
+
+        for (let j = 0; j < suites.length; j++) {
+          for (let l = 0; l < Object.keys(config.multiple).length; l++) {
```
DavertMik: not sure, but can this be written using forEach or `for..in` which looks nicer to me for languages other than C :)

APshenkin:
Hi @DavertMik !

All changes after review are in the branch. Please check it 🚀

DavertMik:
Thanks!

DavertMik:
However, I will do some changes; I think the API can be improved. 

I'd prefer this syntax:

```
codeceptjs run:multiple 'chrome: smoke, firefox: regression'
codeceptjs run:multiple 'chrome: smoke, regression'
codeceptjs run:multiple all
codeceptjs run:multiple 'smoke, regression'
```

* no additional `--multiple` tag
* using comma as separator for suites

PR #441, opened 2017-03-08
Make function parameter name parsing more robust

### LINKED ISSUE #127: Problem when using arrow functions callbacks with no parentheses syntax
If I use this syntax (no parentheses surrounding the 'I' object)

``` javascript
Scenario('test if the welcome is shown', I => {
  I.amOnPage('/');
  I.see('Hello');
});
```

When using the 'SeleniumWebdriver' plugin the tests simply don't run, they get stucked on line that shows the first Feature's name. No errror is shown.

When using the 'WebDriverIO' plugin the tests run, but I got an 'Object of type '/' is not defined in container' error.

If I surround the 'I' object with parenthesis everything run just fine.

rincedd:
Use goatslacker/get-parameter-names to read function parameter names in a more robust way. With this, single-param arrow functions that omit the parentheses do not produce an error anymore.

Fixes #127

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Feature('CodeceptJS demo');
 Scenario('check Welcome page on site', (I) => {
   I.amOnPage('/');
   I.see('Welcome');
-}
+});
 ```
 
 Codeception tests are:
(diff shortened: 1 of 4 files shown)


DavertMik:
Thanks for pointing to the proper library! :+1:

PR #446, opened 2017-03-20
Exit with non-zero code if init throws an error

lennym:
If an error is thrown by the `init` function on startup then the error is caught and the process exits with a zero exit code. This means that CI environments will yield a passing build, when in fact the test runner has suffered a catastrophic error.

Resolve this by exiting with a non-zero code if `init` throws an error.


barryosull:
Just ran into the same issue, very happy to see a fix for it.

DavertMik:
Thanks, very obvious but still was not there. Good catch

PR #455, opened 2017-04-05
Add seeNumberOfVisibleElements

elarouche:
Give the ability to not only see how many times an element appears in the DOM but how many times an element is actually visible.

Had to build this function as a custom helper for my project. I believe this should be implemented. Very useful, especially when writing acceptance tests.

### DIFF
--- a/lib/helper/WebDriverIO.js
+++ b/lib/helper/WebDriverIO.js
@@ -770,7 +770,7 @@ class WebDriverIO extends Helper {
     return this.browser.isVisible(withStrictLocator(selector))
       .then(function(res) {
         res = res.filter((val) => val == true);
-        return assert.equal(res.length, num);
+        return truth(`elements of ${locator}`, 'to be seen').assert.equal(res.length, num);
       });
   }
 
--- a/lib/helper/WebDriverIO.js
+++ b/lib/helper/WebDriverIO.js
@@ -757,6 +757,18 @@ class WebDriverIO extends Helper {
         return assert.equal(res.value.length, num);
       });
   }
+  
+  /**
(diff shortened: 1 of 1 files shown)


elarouche:
In other words, it is a combination of `seeElement` and `seeNumberOfElements` from the WebDriverIO helper

lib/helper/WebDriverIO.js
```diff
@@ -757,6 +757,18 @@ class WebDriverIO extends Helper {
...
+  
+  /**
+   * asserts that an element is visible a given number of times
+   * Element is located by CSS or XPath.
+   */
+  seeNumberOfVisibleElements(selector, num) {
+    return this.browser.isVisible(withStrictLocator(selector))
+      .then(function(res) {
+        res = res.filter((val) => val == true);
+        return assert.equal(res.length, num);
```
DavertMik: please add assertion failure message specific for this case

lib/helper/WebDriverIO.js
```diff
@@ -757,6 +757,18 @@ class WebDriverIO extends Helper {
         return assert.equal(res.value.length, num);
       });
   }
+  
+  /**
+   * asserts that an element is visible a given number of times
```
DavertMik: Please add longer documentation with code example; similar to other methods.

elarouche:
Not a 100% sure about that `add assertion failure message` commit. Went with ``return truth(`elements of ${locator}`, 'to be seen').assert.equal(res.length, num);``, is that the best way to go about it?

DavertMik:
Yes, looks good now

PR #456, opened 2017-04-05
Click only visible elements

imvetri:
Just a POC to throw error if the element is not visually present.

Idea here is to get the coordinates of the target element. Find the element in the document at the same coordinates. If the element doesnt match, its very likely that its not present on the view.

Based on the review comments, refactoring will be taken care.

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -110,8 +110,11 @@ AfterSuite((I) => {
 ## Within
 
 To specify the exact area on a page where actions can be performed you can use `within` function.
-Everything executed in its context will be narrowed to context specified by locator:
+Everything executed in its context will be narrowed to context specified by locator: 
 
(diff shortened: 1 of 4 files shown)


satsukikorin:
I haven't looked at your code yet, but does it send the click event to somewhere in the _middle_ of the element, not just to the top left corner of the bounding rectangle?

DavertMik:
Travis reports an error in Nightmare tests:

```
 the string "el.getBoundingClientRect is not a function" was thrown
```

DavertMik:
please don't include your within patches into this PR

imvetri:
Closing this PR. Will raise a new one as it has other irrelevant commits.

imvetri:
@satsukikorin it sends to (top, left) for now. And we can have control on where the click should be triggered.

PR #457, opened 2017-04-05
Within to maintain context after failure

imvetri:
within doesn't restore the context back to the root if the fn() execution fails because of any run time exception.

Fixed it with try catch finally block


imvetri:
#457

DavertMik:
Please remove all changes which don't belong to this PR.

imvetri:
@DavertMik yeah thanks. that was a mistake

imvetri:
Closing the PR as it contains irrelevant commits.

PR #465, opened 2017-04-10
Fix on seeNumberOfVisibleElements

elarouche:
Just remembered that isVisible returns true or false if can't find multiple elements.
With this fix it will indeed return 0 if no elements found and 1 if only 1 element found.


DavertMik:
THanks for the update

PR #471, opened 2017-04-13
Fix Multiple option

APshenkin:
1. Fix issue, that reports mixed in output
2. Fix issue, that setting different configurations for one browsers not working.
3. Add windowsSize to tag in console


lib/command/run-multiple.js
```diff
@@ -10,19 +10,20 @@ const runner = path.join(__dirname, '/../../bin/codecept');
...
 module.exports = function (suites, options) {
   // registering options globally to use in config
   process.profile = options.profile;
   let configFile = options.config;
   let codecept;
-
   let testRoot = getTestRoot(configFile);
   config = getConfig(configFile);
+  let outputFolder = config.output;
+  let defaultWindowSize = searchValueInObj('windowSize', config.helpers)
```
DavertMik: please remove defaultWindowSize. 
if no `windowSize` option in config, this option should be ignored.

lib/command/run-multiple.js
```diff
@@ -46,47 +47,59 @@ module.exports = function (suites, options) {
     suite = suite.split(':');
 
     let suiteName = suite[0];
-    let browser = suite[1];
+    let browserConfig = {browser: suite[1], windowSize: suite[2]};
```
DavertMik: windowSize should not be hardcoded anywhere in this file.

lib/command/run-multiple.js
```diff
@@ -46,47 +47,59 @@ module.exports = function (suites, options) {
...
     let suiteConf = configMultiple[suiteName];
 
     if (!suiteConf) {
       throw new Error(`Suite ${suiteName} was not configured in "multiple" section of config`);
     }
 
-    runSuite(suiteName, suiteConf, browser);
+    runSuite(suiteName, suiteConf, browserConfig, outputFolder, defaultWindowSize);
```
DavertMik: please don't add extra params; just add output to suiteConf

```js
suiteConf.output = outputFolder
```

and keep old signature

lib/command/run-multiple.js
```diff
@@ -46,47 +47,59 @@ module.exports = function (suites, options) {
...
   });
 };
 
-function runSuite(suite, suiteConf, browser) {
-  let suiteBrowsers = suiteConf.browsers.map((b) => typeof b === 'object' ? b.browser : b);
+function runSuite(suite, suiteConf, configuration, outputFolder, defaultWindowSize) {
```
DavertMik: keep 3 arguments. Do not make functions longer than 3 arguments

lib/command/run-multiple.js
```diff
@@ -46,47 +47,59 @@ module.exports = function (suites, options) {
...
-  if (!browser) {
+  if (!configuration.browser) {
     // run all browsers in this suite
-    suiteBrowsers.forEach((b) => runSuite(suite, suiteConf, b));
+    suiteBrowsers.forEach((configuration) => runSuite(suite, suiteConf, configuration, outputFolder, defaultWindowSize));
     return;
   }
 
-  let overriddenConfig = config;
+  if (!configuration.windowSize) {
```
DavertMik: this block should be removed

Issue #454, opened 2017-04-04
When a failure happens inside a `within` (using Nightmare), all subsequent scenarios retain the selector context


jkronika:
#### What are you trying to achieve?
* Scenario 1 fails inside a `within` block
* Scenario 2 executes with a clean context, and messaging reflects this

#### What do you get instead?
* Scenario 1 fails inside the `within` block
* Scenario 2 executes in the context from the first scenario's `within`, or at least the messaging seems to indicate as much. Some tests continue to pass, while others fail even though a manual run works fine.
* When a subsequent scenario has its own `within`, the new context is used until that block closes, then we return to the selector context from the failed scenario.

The following examples are contrived, but should demonstrate the issue...

##### Example Output
```bash
 Page Feature 1
 • I am on page "...url..."
 • I see element ".some .selector .on .page"
   Within .some .selector .on .page form:
   • Within .some .selector .on .page form: I see element "label select[name="foo"] option:nth-child(2)"
 ✖ FAILED in 2354ms

 Page Feature 2
   • Within .some .selector .on .page form: I am on page "...url..."
   • Within .some .selector .on .page form: I see element "main > .second .scenario .wrapper"
 ✖ FAILED
```

##### Example Code
```js
let url = `...url...`; // actual URL used is internal and cannot be shared, but shouldn't matter

Scenario(`Page Feature 1`, (I) => {
    let WRAPPER = `.some .selector .on .page`;

    I.amOnPage(url);
    I.seeElement(WRAPPER);

    within(`${WRAPPER} form`, () => {
        I.seeElement(`label select[name="foo"] option:nth-child(2);
        I.say(`We never get this far because the above line fails.`);
    });

    I.say(`Likewise we never execute anything down here, due to the earlier failure inside the within`);
});

Scenario(`Page Feature 2`, (I) => {
    // New scenario, obviously not in the previous one's within block
    let WRAPPER = `main > .second .scenario .wrapper`;

    I.amOnPage(url);
    I.seeElement(WRAPPER);
    // why did the above say they were within the previous scenario's form?!?
});
```
### Details

* CodeceptJS version: 
* NodeJS Version: 6.9.1
* Operating System: ^0.5.0
* Protractor || WebDriverIO || Nightmare version (if related): Nightmare ^2.7.0
* Configuration file:

```js
var config = {
    tests: "./lib/functional/run.functional.suite.js",
    timeout: 20000,
    output: "./test-reports",
    helpers: {
        Codecept: {
[...]

PR #492, opened 2017-04-26
wrap within callback in try catch

imvetri:
Issue that will be fixed by the PR : https://github.com/Codeception/CodeceptJS/issues/454

### DIFF
--- a/lib/within.js
+++ b/lib/within.js
@@ -21,18 +21,24 @@ function within(context, fn) {
       if (helpers[helper]._withinBegin) recorder.add('start within block', () => helpers[helper]._withinBegin(context));
     });
 
-    fn();
+    try{
+      fn();
+    }
+    catch(e){
+      //need suggestion how to report error with respect to codeceptJS
+    }
+    finally{
+      Object.keys(helpers).forEach((helper) => {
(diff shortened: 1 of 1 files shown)


lib/within.js
```diff
@@ -21,18 +21,24 @@ function within(context, fn) {
       if (helpers[helper]._withinBegin) recorder.add('start within block', () => helpers[helper]._withinBegin(context));
     });
 
-    fn();
+    try{
+      fn();
+    }
+    catch(e){
+      //need suggestion how to report error with respect to codeceptJS
```
imvetri: I would like to report it to user if callback fails on exception. Not sure how to do it in codecept way, hoping for guidance here.

imvetri: @DavertMik travis tests are failing and I couldn't understand why. Could you please share some guidance on that ?

DavertMik: Yes, looks like test failure is completely unrelated. I restarted the job

DavertMik: Please use this to add error to promise chain:

```js
 recorder.throw(err); // catch error in test and rethrow into promise
```

DavertMik:
Outdated

PR #494, opened 2017-04-27
Add basic REST services functionalities

Kalostrinho:
Added _REST.js_ helper to fulfill the needs for testing REST services through the use of the **UNIREST** library.

### DIFF
--- a/lib/helper/REST.js
+++ b/lib/helper/REST.js
@@ -1,0 +1,121 @@
+'user strict';
+const Helper = require('../helper');
+const requireg = require('requireg');
+let unirest = requireg('unirest');
+
+/**
+*REST helper
+*--------
+*
+*Gives **CodeceptJs** the ability to execute RESTful API calls, helps to create web service tests also to combine API calls with WebDriverIO / SeleniumWebDriver tests.
+*Currently under development phase.
+*Please visit our [demo repo](https://github.com/DigitalOnUs/codeceptjs-poc/tree/master/tests/REST#using-rest-api-on-codeceptjs) to know how to use this Helper.
+*
+*
+*### GET:
+*
+*There are two ways to use this wrapper, by setting an endpoint or giving an absolute URL.
+*
+*#### Absolute URL:
+*```javascript
+*let Request = I.sendRequest('GET','http://hostname/user/1')
+*```
+*
+*#### Endpoint hit:
+*
+*For this you need to specify on the module configuration the endpoint is going to be used.
+*```javascript
+*"REST": {
+*  "endpoint": "http://api.mysite.com/v1"
+*  }
+*```
+*
+*Then use the Codecept wrapper
+*```javascript
+*let Request = I.sendRequest('GET','/users')
+*```
+*
+*We are using [UNIREST](http://unirest.io/nodejs.html#request) library to execute API calls, so a `Request` object is returned, check out the documentation
+*to get more details on how it works.
+*
+*#### Example using chainable unirest methods.
+*
+*This example comes from unirest library but we adapted to allow you to check the behavior using Codecept.
+*
+*```javascript
+*let Request = I.sendRequest('POST','http://mockbin.com/request')
+*Request
+*.headers({'Accept': 'application/json', 'Content-Type': 'application/json'})
+*.send({ "parameter": 23, "foo": "bar" })
+*.end(function (response) {
+*  console.log(response.body);
+*});
+*```
+*
+*### Available methods:
+*
+*GET, HEAD, POST, PUT, PATCH, DELETE.
+*
+*Extra funcionality Cookie Jar, [read unirest doc about it](https://github.com/Mashape/unirest-nodejs#unirestjar)
+*```javascript
+*let cookie = I.createACookieJar('http://mockbin.com/request', 'key=value', '/');
+*```
+**/
+
+class REST extends Helper {
+
+  constructor(config) {
+    super(config);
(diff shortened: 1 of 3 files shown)


lib/helper/REST.js
```diff
@@ -0,0 +1,110 @@
...
+const Helper = require('../helper');
+const requireg = require('requireg');
+let unirest;
+
+/**
+* REST helper to execute RESTful API calls, this is currently under development phase.
+* Please visit our demo repo to know how to use this Helper.
+* [https://github.com/DigitalOnUs/codeceptjs-poc/tree/master/tests/REST#using-rest-api-on-codeceptjs]
+* Usage:
+*      let response = yield I.sendGet('https://randomuser.me/api/')
```
DavertMik: please markdown code formatting tags.
and move this example to `sendGet` method

lib/helper/REST.js
```diff
@@ -0,0 +1,110 @@
...
+                .end(function (response){
+                  if (response.code >= 400) {
+                    return reject(response);
+                  }
+                  return resolve(response);
+                });
+    });
+  }
+
+  sendPost(url, headers = {}, payload = {}){
```
DavertMik: We need a docblock documentation for EACH public method with example.
So you will need to write some docs about sendPost, sendGet, sendDelete, etc

lib/helper/REST.js
```diff
@@ -0,0 +1,110 @@
...
+                  }
+                  return resolve(response);
+                });
+    });
+  }
+
+  sendPut(url, headers = {}, payload = {}){
+    let request = unirest.put(url);
+    request.timeout(this.options.timeout);
+    return new Promise(function (resolve, reject){
```
DavertMik: Looks like sendXXX methods have common logic for sending requests.
How about create a general `sendRequest` method which will be used by all of them?

Maybe we can even make it public and use like this:
```js
I.sendRequest('GET', 'http://xxx/yyy');
```

So we could make also PATH, HEAD, LINK and other strange requests )

lib/helper/REST.js
```diff
@@ -0,0 +1,110 @@
...
+*      }
+*/
+
+class REST extends Helper {
+
+  constructor(config) {
+    super(config);
+    unirest = requireg('unirest');
+    this.options = {
+      timeout: 3000
```
DavertMik: I'd like to include `endpoint` config here. It will work similarly to `url` config option as it is done in browser helpers.

So if I set:
```
endpoint: http://api.mysite.com/v1
```

I could make requests like:

```js
I.sendGet('/users');
```

If I provide a full URL the request should go as it is now

lib/helper/REST.js
atrevino: Thanks for the review, we actually thought about creating a generic _sendRequest_ public method, but we wanted you to review this way first... anyhow, will apply now the changes you ask for and wait for next review, thanks a lot!

Let

atrevino:
One of the problems we detected is that when REST and Webdriver modules are active, browser is always lunched, any idea to not start browser until a Webdriver command is executed? @DavertMik

Kalostrinho:
Thanks for these observations @DavertMik! We'll work on these right away!

DavertMik:
@atrevino we have `manualStart` option for this. http://codecept.io/helpers/WebDriverIO/#configuration
So you can create your own command and start browser at the point you want. 

But probably you can implement the logic you proposed: start browser only when the fiest browser command is called. Maybe that would be even more useful

atrevino:
Hey @DavertMik I did some code changes to use a different approach to wrap unirest library and be able to execute API calls. Here are "the documentation" I generated to use it, let me know if that approach is ok and is good enough to push the code changes.


REST helper
--------

Gives **CodeceptJs** the ability to execute RESTful API calls, helps to create web service tests also to combine API calls with WebDriverIO / SeleniumWebDriver tests.
Currently under development phase.
Please visit our [demo repo](https://github.com/DigitalOnUs/codeceptjs-poc/tree/master/tests/REST#using-rest-api-on-codeceptjs) to know how to use this Helper.


### GET:

There are two ways to use this wrapper, by setting an endpoint or giving an absolute URL.

#### Absolute URL:
```javascript
let Request = I.sendRequest('GET','http://hostname/user/1')
```

#### Endpoint hit:

For this you need to specify on the module configuration the endpoint is going to be used.
```javascript
"REST": {
  "endpoint": "http://api.mysite.com/v1"
  }
```

Then use the Codecept wrapper
```javascript
let Request = I.sendRequest('GET','/users')
```

We are using [UNIREST](http://unirest.io/nodejs.html#request) library to execute API calls, so a `Request` object is returned, check out the documentation 
to get more details on how it works. 

#### Example using chainable unirest methods.

This example comes from unirest library but we adapted to allow you to check the behavior using Codecept.

```javascript
let Request = I.sendRequest('POST','http://mockbin.com/request')
Request
.headers({'Accept': 'application/json', 'Content-Type': 'application/json'})
.send({ "parameter": 23, "foo": "bar" })
.end(function (response) {
  console.log(response.body);
});
```

### Available methods:

GET, HEAD, POST, PUT, PATCH, DELETE.

Extra funcionality Cookie Jar, [read unirest doc about it](https://github.com/Mashape/unirest-nodejs#unirestjar)
```javascript
let cookie = I.createACookieJar('http://mockbin.com/request', 'key=value', '/');
```

Kalostrinho:
Closing to fix commit structure first, code improvements afterwards

PR #504, opened 2017-05-05
REST helper update, fix asynchronicity to get response from APIs

atrevino:


DavertMik:
Thanks for the update. However I don't like that there is no easy way to create request and you need to use unirest API to do that. This breaks CodeceptJS principles. 

What I'd like to see is sending request like

```js
I.sendPost('/users', { name: 'davert' }, { headers: { auth: 11111111 }, cookies: {} });
```

or set headers for all requests:

```js
I.haveRequestHeader({ auth: 111111 });
I.sendPost('/users', { name: 'davert' });
```

so the idea is that most of requests to be available via simple sendGet, sendPost, etc. So you won't need method chaining and so on. Sure there should be an option to start request manually with `generateRequest` but this should not be the default option

atrevino:
You want a specific method to return the response from API call? 
Like:
```javascript
I.haveRequestHeaders({auth:111});
I.sendPost('/users');

let response = yield I.getResponse();
```

or directly from `I.sendPost('/user');`
Like
```javascript
I.haveRequestHeaders({auth:111});
let response = yield I.sendPost('/users');
```

DavertMik:
Yes, I like this a bit more

```js
I.haveRequestHeaders({auth:111});
let response = yield I.sendPost('/users');
```

Thanks :)

atrevino:
Sorry for the delay, I'm doing tests about this. Probably will push only for review.

DavertMik:
Cool. Waiting for the update!

DavertMik:
Thanks. I will work on improve it and maybe add some tests

Issue #505, opened 2017-05-05
Minimum version of node required is not documented


AndyGaskell:
#### What are you trying to achieve?
Install CodeceptJS on Ubuntu

#### What do you get instead?
Can't get it installed.

### Details
I think the issue is really just that a minimum node.js version is not documented.  This is more of an problem as the latest Ubuntu includes node.js 4.7.2, which does not work with CodeceptJS.  The simplest change here would be just an update to the read-me.  A better options would be checks in the npm install logic, with either a warning or an error.  CodeceptJS is a great system, but it is a real shame if some people struggle to get started.  Although it's really Ubuntu's fault for not shipping newer versions of node, we can probably help people.  I hope it is helpful to raise a issue for this.  This came from issues that I first posted on StackOverflow at http://stackoverflow.com/questions/43679010/installation-problems-with-codeceptjs and there is more info and background there.  Thanks to @cwhite92 for helping me.

* CodeceptJS version: v0.6.2
* NodeJS Version: 4.7.2
* Operating System: Ubuntu 17.04 x86_64 and Ubuntu 16.04.1 LTS

PR #507, opened 2017-05-08
Adding minimum node version note

AndyGaskell:
Just a note on node.js minimum version, re #505 Minimum version of node required is not documented

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,8 @@ codeceptjs def .
 
 Later you can even automagically update Type Definitions to include your own custom [helpers methods](docs/helpers.md).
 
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks. Our test suite is still running on 4.x so according to Travis CodeceptJS I supposed that it should work with 4.x...
I will update the minimum version in docs in next release

AndyGaskell:
Hi, thanks for looking at my PR :) 
I think it was mainly the spread operator, seemed like it wasn't supported in 4.7.2, though it might be ok in 4.8.3.  What's it running on in the Travis set-up?  
It's definitely fine in 5.12.
http://node.green/#ES2015-syntax-spread-------operator

PR #510, opened 2017-05-10
Update to docker files

ianfinch:
I tried to run as a docker image, but this failed with an error saying "unable to find config file".

This seems to be because "run.sh" executes in "/" (and so does "codecept.js"), but the config file is in "/tests/".  I've made a couple of small changes (to the Dockerfile and docker/run.sh) to cd to the /tests/ directory and also to allow a CODECEPT_ARGS rather than GREP env var (I'm assuming the docker file is based on an earlier codecept.js syntax).

Happy to discuss.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM 	node:6.9.5
 # LTS
 
 		# Set grep as an ENV variable
-ENV 	GREP=" "
+ENV 	CODECEPT_ARGS=""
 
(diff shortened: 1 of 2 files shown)


DavertMik:
Yes. Thanks for the update

SuneRadich:
Ah, this has me pulling my hair. My first time trying out docker images, and could not figure out what was wrong. (I downloaded the docker image this morning :-)

Its working now.

PR #547, opened 2017-06-06
docs: tidy up the advanced usage

clarkie:
- make the config consistent 
- reorder the options so that `--all` is top of the list. My assumption but this is probably the most common usage
- add `sh` to code blocks
- fix typo


DavertMik:
Yes, thanks I like the way you reordered information. 
And sorry for the late response

PR #549, opened 2017-06-07
Fix saveScreenshot() for PhantomJS

HughZurname:
#### Changes:
- Updated other helpers to use options for configuring fullPageScreenshots.
#### Fixes:
- Resolves #540 based on suggestion.

PR #557, opened 2017-06-16
[WIP] Database Helper - Initial work

### LINKED ISSUE #540: saveScreenshot() doesn't work for failed scenarios using PhantomJS

If a scenario is failed the screenshot will be saved as expected.

**What do you get instead?**

Error (s. the attachement)
save_scren_failed

### Details

related to #523

### LINKED ISSUE #551: [Enhancement] - Database Helper

Currently I use [lokijs ](http://lokijs.org/) in a custom helper of sorts to handle some data driven test scenarios. 

I find it pretty useful in few cases: 
- When you're randomizing some data in a test and want to validate against it at some point in the future as part of an e2e run. 
- Data driven testing becomes easier by creating sets that you combine through queries and can access these values form throughout the test execution.
- Writing values out to the database can be useful for retrospective debugging of failed tests.

### Details
I've started work on a helper for this, so this is both a statement of intent and an opportunity for any suggestions, feedback or requests before I begin in earnest.

HughZurname:
## What does this contain?
- Initial work on an in memory database helper. I used [lokijs ](http://lokijs.org/) for this.
- Seed data import mechanism.
- Basic Create, Remove, Update, Delete functionality.
- Documentation. This needs to be expanded, but the initial work is there.
- Resolves #551 
- Rolls up changes in #549 (so I could remove it?), which resolves #540.

## What's outstanding?
- Unit tests
- More documentation and examples.


DavertMik:
Thanks for the pull request. However, I need to understand it a bit more
I never used LokiJS so I don't understand the use cases for it. Could you explain how do you use it? 
Why Loki?

HughZurname:
Hey 😄. Loki is just a mongoDB like in memory document store, the main reason I chose it is because it's really lightweight and provides a lot of functionality for it's footprint. 

The reason I ended up using it was because I needed to effectively store some variables between tests to validate against in a later scenario. At the time I figured I could just use the filesystem helper, but static files proved insufficient for the task. 

I think the best explanation would be to give you a more concise version of how I've used it in my own tests. I've added commentary inline:
```js
let user, claimId;

Scenario("Enter user Details", function* (I) {
    
    //Here I am retrieving some user form data that I seeded before the test execution.
    //The arguments are the document store I am querying and the query object itself.
    //The first matching value is selected, but this can just as easily be used to randomise the selected user.
    user = yield I.findOne("users", {email: "someguy@email.com"});

    //testId is generated elsewhere per run
    I.fillField("orderNumber", testId);
    I.dontSee("Document number is required");

    I.fillField("firstName", user.firstname);
    I.dontSee("First name is required");

    I.fillField("surname", user.surname);
    I.dontSee("Surname is required");

    I.fillField("addressLine1", user.address1);
    I.dontSee("Address1 is required");

    I.fillField("postCode", user.postcode);
    I.dontSee("Postcode is required");

    I.fillField("phoneNumber", user.number);
    I.dontSee("Telephone number is required");

    I.fillField("email", user.email);
    I.dontSee("Must be a valid email");
});

Scenario("Save & Continue",  function* (I) {
    I.click("Save & Continue");

    //Custom helper function to grab the order id from the url.
    claimId = yield I.grabUrlPart(4);
});

AfterSuite((I) => {
    //Here I create a new document store for orders before inserting relevant data from the tests I've just completed.
    //This becomes queryable like users was above in later scenarios allowing me to validate against known data.
    I.addCollection("orders");
    I.insert("orders", {
        orderId: claimId,
        orderNumber: testId,
        email: user.email,
        fullName: `${user.firstname} ${user.surname}`,
    });
});
```
I hope this makes sense?  If not, just let me know and I'll answer it as best I can.

DavertMik:
I was thinking on this and concluded that this module is too specific.
I'm not sure how useful it can be for wide audience. 
So I'd propose to publish this helper as standalone package and I will add a link to it into Data Management chapter of documentation (which I'm currently writing)

HughZurname:
Sure 😁, how would I go about doing that? Are there any examples of this having been done for codecept that I could use as a rough guide? I saw something about meta-packages, but I'm guessing that's not quite what I'm aiming for.

DavertMik:
Not meta-packages, they are just for simpler installations.

So you extract your helper, then and create npm package, named like: 'codeceptjs-loki'.
Then anyone can install this package and include your helper like:

```js
helpers: {
  Loki: {
    require: "codeceptjs-loki"
  }
}
```

CodeceptJS will execute `require('codeceptjs-loki')` to load the helper.
That's all!

HughZurname:
Hi 😄 

So I've extracted it to it's own package and published it to npm [here](https://www.npmjs.com/package/codeceptjs-loki) and it seems to work as mostly you've described above. 

The only difference is that it's necessary to give it an explicit path for some reason. 
```
"helpers": {
  "Loki": {
    "require": "node_modules/codeceptjs-loki"
  }
}
```
I've tried to look at the config creation scripts to try and track down why this may happen, but I've not managed to follow it through.

Issue #548, opened 2017-06-07
Multiple run and XML report


enes-ergul:
#### What are you trying to achieve?
I want to use the new multiple run feature and export the report as xml file

> $(npm bin)/codeceptjs run-multiple --all --reporter mocha-junit-reporter

#### What do you get instead?
Only one of the parallel runs is present in the XML output

### Details

* CodeceptJS version: 0.6.3
* NodeJS Version: 6.9.2
* WebDriverIO Version: 4.8.2
* Mocha JUnit Reporter Version: 1.13.0

shrikanthkm07:
Even I am facing the same issue. Actually I don't even see the steps also. Is these any solution for this ?

APshenkin:
@enes-ergul @shrikanthkm07 
As I understand, You use this npm package?
https://www.npmjs.com/package/mocha-junit-reporter
If so, than we have to add new override in multiple-run for `mochaFile` parameter. https://github.com/Codeception/CodeceptJS/blob/master/lib/command/run-multiple.js#L95
and pass reporter folder via codeceptJS config.

You can add PR for this or wait, when I'll be able to take this in work

APshenkin:
@enes-ergul @shrikanthkm07 
I have added new override in multiple-run. Reports files will have `report.xml` name. This changes will come with 1.0 release

Issue #543, opened 2017-05-31
After block not properly executed if Scenario fails


SuneRadich:
#### What are you trying to achieve?
Run logout logic in an After block for each Scenario, also on Scenarios that fail

#### What do you get instead?
My After block logic is not run for the test that fails. It console.logs my debug statement, so the After block is called, but the steps are not executed. Maybe because Teardown has already been started?

> Provide console output if related. Use `--verbose` mode for more details.

```bash
codeceptjs run --verbose --debug
CodeceptJS v0.6.3
Using test root "/Users/srv/Documents/Projects/nlpgmc"

POC --
   [1] Starting recording promises
   Emitted | suite.before ([object Object])
   [1] Queued | hook Protractor._beforeSuite()
 test something #1
 > [Session] Starting singleton browser session
   [2] Starting recording promises
   Emitted | test.before
   [2] Queued | hook Protractor._before()
LOGIN!
   Emitted | step.before (I am on page "/#/welcome")
   [2] Queued | amOnPage: "/#/welcome"
   Emitted | step.after (I am on page "/#/welcome")
   [2] Queued | return step result
   Emitted | step.before (I click ".button--primary")
   [2] Queued | click: ".button--primary"
   Emitted | step.after (I click ".button--primary")
   [2] Queued | return step result
   Emitted | step.before (I click {"css":"titlebar .align-left button"})
   [2] Queued | click: {"css":"titlebar .align-left button"}
   Emitted | step.after (I click {"css":"titlebar .align-left button"})
   [2] Queued | return step result
   Emitted | step.before (I click "Log på med testbruger")
   [2] Queued | click: "Log på med testbruger"
   Emitted | step.after (I click "Log på med testbruger")
   [2] Queued | return step result
   Emitted | step.before (I fill field ".ghostInput", "0801541089")
   [2] Queued | fillField: ".ghostInput", "0801541089"
   Emitted | step.after (I fill field ".ghostInput", "0801541089")
   [2] Queued | return step result
   Emitted | step.before (I click "Next")
   [2] Queued | click: "Next"
   Emitted | step.after (I click "Next")
   [2] Queued | return step result
   Emitted | step.before (I click "No, enter this later")
   [2] Queued | click: "No, enter this later"
   Emitted | step.after (I click "No, enter this later")
   [2] Queued | return step result
   Emitted | step.before (I wait for text "Savings", 5, "h1")
   [2] Queued | waitForText: "Savings", 5, "h1"
[...]

PR #568, opened 2017-06-22
Fix work with promise chains

APshenkin:
Our goals:
- [x] Fix https://github.com/Codeception/CodeceptJS/issues/543 - After block not properly executed if Scenario fails
- [x] Expected behavior in promise chains: `_beforeSuite` hooks from helpers -> `BeforeSuite` from test -> `_before` hooks from helpers -> `Before` from test - > Test steps -> `_failed` hooks from helpers (if test failed) -> `After` from test -> `_after` hooks from helpers -> `AfterSuite` from test -> `_afterSuite` hook from helpers.
- [x] if during test we got errors from any hook (in test or in helper) - stop complete this suite and go to another
- [x] if during test we got error from Selenium server - stop complete this suite and go to another
- [x] if `restart` option is false - close all tabs expect one in `_after` (ready for webdriverIO, SeleniumWebdriver and Protractor. For Nightmare it will be hard to add (I found only this way https://github.com/rosshinkley/nightmare-window-manager)
- [x] Complete `_after`, `_afterSuite` hooks even After/AfterSuite from test was failed
- [x] Don't close browser between suites, when `restart` option is false. We should start browser only one time and close it only after all tests.
- [x] Close tabs and clear local storage, if `keepCookies` flag is enabled

Bonus:
- [x] Add step execution time in debug mode
- [x] Fix https://github.com/Codeception/CodeceptJS/issues/548


DavertMik:
Please fix `test/runner/interface_test.js`, it tests the order of events

DavertMik:
Wow, awesome work. We are getting closer and closer to 1.0 release!

lib/interfaces/bdd.js
```diff
@@ -45,6 +49,12 @@ module.exports = function (suite) {
         suites.shift();
       }
       if (!opts) opts = {};
+
+      afterAllHooks = [];
```
DavertMik: this was already defined at line 27

DavertMik:
Please try to use interactive shell to check that nothing was broken for it.
Also, execute one test with `within` block to be sure this changes to recorder doesn't break them.

APshenkin:
@DavertMik 
During these changes I found one interaction that may be not correct. I want to ask your advice about it.
Now we have clear promises chain as described in topic. It will break only after fail. Also if we get error in `BeforeSuite`, `Before` or `After` hooks we will skip all tests in this suite (looks it's ok, because there is no sense to try repeat `BeforeSuite/Before/After` if it's not working). But there is an interesting ways with `After` and `AfterSuite` hooks.
If `After` hook from test will fail, then `_after` hooks from helpers will not be executed.
The same thing with `AfterSuite` - if it fails, then `_afterSuite` hooks from helpers will not be executed. It looks normal, but this can cause an issue, that after fails in these hooks, browser will not close or we will not clear environment for next tests.

I think, that maybe we can track this and force close/clear browser in this cases.

What do you think? Should we track this, or it should be resolved from user side?

DavertMik:
> If After hook from test will fail, then _after hooks from helpers will not be executed.

Yes, this makes a big problem for closing browser sessions... 
I have 2 ideas about this:

1. run `_after` hook for Helper even test has failed in `After`. Same for `AfterSuite`.
2. add new `finalize` hook which will be executed no matter what happened.

But this is important to solve it in one way or another.

lib/interfaces/bdd.js
APshenkin: fixed

APshenkin:
@DavertMik please review the changes. `pause()` and within block works as expected

lib/event.js
```diff
@@ -11,6 +11,14 @@ module.exports = {
     passed: 'test.passed',
     failed: 'test.failed',
   },
+  hook: {
```
APshenkin: we can't use test.after/suite.after, because they are already defined. I suggest test.afterError, suite.afterError

lib/helper/SeleniumWebdriver.js
```diff
@@ -668,6 +675,27 @@ class SeleniumWebdriver extends Helper {
   }
 
   /**
+   * Close all tabs expect for one.
+   *
+   * ```js
+   * I.closeTabsExceptForOne();
+   * ```
+   */
+  closeTabsExceptForOne() {
```
APshenkin: fixed

lib/scenario.js
```diff
@@ -90,21 +96,30 @@ module.exports.injected = function (fn, suite) {
...
 module.exports.setup = function () {
-  recorder.start();
+  if (!recorder.isRunning()) {
+    recorder.start();
+  }
   event.emit(event.test.before);
 };
 
 module.exports.teardown = function () {
+  if (!recorder.isRunning()) {
```
APshenkin: fixed

lib/interfaces/bdd.js
```diff
@@ -88,7 +95,18 @@ module.exports = function (suite) {
         fn = opts;
         opts = {};
       }
-
+      if (!afterAllHooksAreLoaded) {
+        afterAllHooks.forEach(function (hook) {
```
APshenkin: I added comment in test.
FYI, we collect all hooks in right order in array and then push them to suite. We start collecting hooks from helpers and after from testfile. this cause hooks reordering 
(expected: `After` from test -> `_after` hooks from helpers -> `AfterSuite` from test -> `_afterSuite` hook from helpers.
actual:  `_after` hooks from helpers -> `After` from test -> `_afterSuite` hook from helpers. -> `AfterSuite` from test )

This change fix this chain

lib/step.js
```diff
@@ -22,6 +22,7 @@ class Step {
   }
 
   run() {
+    this.startTime = new Date();
```
APshenkin: If we place it in `steps.js` then we can't pass this value properly between states. Or I don't know how to do it. If you have any suggestion, it will be awesome

lib/event.js
DavertMik: > test.afterError, suite.afterError

good

Issue #573, opened 2017-06-26
I.seeNumberOfVisibleElements() throws an error 'locator is not defined'


BorisOsipov:
#### What are you trying to achieve?
I'm trying to use seeNumberOfVisibleElements() for testing purpose

#### What do you get instead?
I am getting an error:

_ReferenceError: locator is not defined
at Object.<anonymous> (/usr/lib/node_modules/codeceptjs-webdriverio/node_modules/codeceptjs/lib/helper/WebDriverIO.js:782:37)_

> Provide console output if related. Use `--verbose` mode for more details.

```bash
CodeceptJS v0.6.3
Using test root "~/petclinic"

....
-- FAILURES:
  1) Some important feature: Some important test:
     locator is not defined  
  Scenario Steps:  
  - I.seeNumberOfVisibleElements({"css":"#someInvalidId"}, 2) at Test.Scenario (tests/Create_owner_test.js:10:7)
  
  ReferenceError: locator is not defined
      at Object.<anonymous> (/usr/lib/node_modules/codeceptjs-webdriverio/node_modules/codeceptjs/lib/helper/WebDriverIO.js:782:37)
      at Object.exec (/usr/lib/node_modules/codeceptjs-webdriverio/node_modules/webdriverio/build/lib/helpers/safeExecute.js:28:24)
      at Object.resolve (/usr/lib/node_modules/codeceptjs-webdriverio/node_modules/webdriverio/build/lib/webdriverio.js:189:29)
      at /usr/lib/node_modules/codeceptjs-webdriverio/node_modules/webdriverio/build/lib/webdriverio.js:538:32
      at _fulfilled (/usr/lib/node_modules/codeceptjs-webdriverio/node_modules/q/q.js:854:54)
      at self.promiseDispatch.done (/usr/lib/node_modules/codeceptjs-webdriverio/node_modules/q/q.js:883:30)
      at Promise.promise.promiseDispatch (/usr/lib/node_modules/codeceptjs-webdriverio/node_modules/q/q.js:816:13)
      at /usr/lib/node_modules/codeceptjs-webdriverio/node_modules/q/q.js:624:44
      at runSingle (/usr/lib/node_modules/codeceptjs-webdriverio/node_modules/q/q.js:137:13)
      at flush (/usr/lib/node_modules/codeceptjs-webdriverio/node_modules/q/q.js:125:13)

  FAIL  | 0 passed, 1 failed   // 7s
   Emitted | global.result ([object Object])
```
[...]

PR #574, opened 2017-06-26
#573 I.seeNumberOfVisibleElements() throws an error 'locator is not defined' 

BorisOsipov:
Hi folks!

I found issue and tries to fix it, please, have look.
I am a little bit new in JS and may do it in a wrong way.

Thank you


DavertMik:
Looks good to me. Thank you!

Issue #560, opened 2017-06-19
       "windowSize": "maximize" can't support  chrome 59 version


tony204:
"windowSize": "maximize" can't support  chrome 59 version

if you user chrome 59 version ,  maximize or other windowsize can't support
@DavertMik

DavertMik:
Yes, I think this is a limitation of ChromeDriver and Chrome 59
Not sure how to overcome it. You can only pass the `window-size` argument to `args` for ChromeOptions to set the initial size for Chrome

APshenkin:
@DavertMik @tony204 
I think we can try this hack:
Execute script in browser to get information about screen resolution:
```
return screen.width;
return screen.height;
```
Then pass them to windowHandleSize function

Here is a demo how this scripts works in browser: https://jsfiddle.net/8q67em27/1/

tony204:
@DavertMik  @APshenkin 
i update the source code,upload the 3 files WebDriverIO.js , SeleniumWebdriver.js, Appium.js that you guys commit to master, but it also didn't fix the chrome 59 maximize problem.
Can you confirm is it any steps maybe i missed?

ninpop25:
@DavertMik @APshenkin : I am facing issue with *Window.Maximize()*
below is the ticket
https://github.com/Codeception/CodeceptJS/issues/892
please help me,as its failing my automation build first test and thus results in failed build

PR #575, opened 2017-06-27
Fix maximize

APshenkin:
Fix https://github.com/Codeception/CodeceptJS/issues/560

### DIFF
--- a/lib/helper/Appium.js
+++ b/lib/helper/Appium.js
@@ -125,8 +125,8 @@ class Appium extends WebdriverIO {
         return this.browser.windowHandleSize({
           width: res.value[0],
           height: res.value[1]
-        })
-      })
+        });
+      });
     } else if (this.options.windowSize && this.options.windowSize.indexOf('x') > 0) {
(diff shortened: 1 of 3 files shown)


lib/helper/Appium.js
```diff
@@ -121,9 +121,13 @@ class Appium extends WebdriverIO {
     }
 
     if (this.options.windowSize === 'maximize') {
-      this.browser.windowHandleMaximize(false);
-    }
-    if (this.options.windowSize && this.options.windowSize.indexOf('x') > 0) {
+      this.browser.execute('return [screen.width, screen.height]').then((res) => {
```
DavertMik: I think this works only in browsers...

DavertMik:
I will merge it after my PR because we probably have conflicts here

lib/helper/Appium.js
APshenkin: Yes, if you set windowSize in config we think, that you plan to use browsers to test

PR #581, opened 2017-06-29
Nightmare restart fix

APshenkin:
Fix 
`Error: .evaluate() fn should be a function` in after block for Nightmare when `restart:false`


DavertMik:
Thanks!

PR #591, opened 2017-07-05
Refactored steps and hooks improvements

DavertMik:
Simplified steps improvements and hooks handling by @APshenkin 

* removed additional events for after and afterSuite hooks
* added `step.passed` and `step.failed` sync events similar to test.passed and test.failed
* separated steps time calculation and printing


lib/scenario.js
```diff
@@ -81,9 +81,8 @@ module.exports.injected = function (fn, suite, hookName) {
     }
     recorder.catch((err) => {
       event.emit(event.test.failed, suite, err); // emit
-      if (hookName === 'after' || hookName === 'afterSuite') {
-        event.emit(event.hook[hookName].failed, suite);
-      }
+      if (hookName === 'after') event.emit(event.test.after, test);
```
APshenkin: Should be if (hookName === 'after') event.emit(event.test.after, suite);. Test is undefined

APshenkin: Also there is an issue in bdd.js
now
context.After = function (fn) {
      afterEachHooks.unshift(['After', scenario.injected(fn, suites[0]), 'after']);
    };

should be
context.After = function (fn) {
      afterEachHooks.unshift(['After', scenario.injected(fn, suites[0], 'after')]);
    };

Because of this we don't get name of hook in this place for After

lib/reporter/cli.js
```diff
@@ -66,9 +66,9 @@ class Cli extends Base {
         }
       });
 
-      event.dispatcher.on(event.step.started, function (step) {
-        output.step(step, global.suiteDetails);
-      });
+      event.dispatcher.on(event.step.started, (step) => output.step(step));
+      event.dispatcher.on(event.step.passed, (step) => output.stepExecutionTime(step));
```
DavertMik: Yes, output can get broken in --verbose and --debug mode (if something is printed during debug)
That may happen. And I don't have any solution to that (

Let's have a different format then.

In `--debug`: 

```
• I am on page "https://github.com/404" (9.716 sec)
```

in `--verbose`:

```
 • I am on page "https://github.com/404"
   Emitted | step.passed (I am on page "https://github.com/404")
   Finished in 9.716 sec
```

DavertMik:
@APshenkin please check my updates

lib/reporter/cli.js
APshenkin: 👍

PR #592, opened 2017-07-06
Use mkdirp to create output directory

vkramskikh:
fs.mkdirSync fails if config.output has subdirectories


DavertMik:
This package is not listed in dependencies...

vkramskikh:
It is https://github.com/Codeception/CodeceptJS/blob/master/package.json#L37

DavertMik:
Oh, sorry, can't keep them all in my head. :upside_down_face: 
Let's get it merged then!

PR #600, opened 2017-07-15
Refactored data driven tests

DavertMik:
* documentation updated
* specify dataset using `Data()`
* creating data tests moved out from `bdd.js` to keep it simple
* added `test.inject` property for not to hardcode data injection in `scenario.js`
* data driven test can accepts DataTable, array, function, generator
* renamed `scenarioContext` to `current` for better readbility

```js
// define data table inside a test or load from another module
let accounts = new DataTable(['login', 'password']); //
accounts.add(['davert', '123456']); // adding records to a table
accounts.add(['admin', '123456']);

// Pass dataTable to Data()
// Use special param `current` to get current data set
Data(accounts).Scenario('Test Login', (I, current) => {
  I.fillField('Username', current.login); // current is reserved!
  I.fillField('Password', current.password);
  I.click('Sign In');
  I.see('Welcome '+ current.login);
});
```


lib/data/context.js
```diff
@@ -0,0 +1,50 @@
...
+        data.forEach((dataRow) => {
+          opts.data = dataRow;
+          let test = context.Scenario(`${title} | ${dataRow}`, opts, fn);
+          test.inject.current = dataRow;
+        });
+      }
+    };
+  };
+
+  context.xData = function (dataTable) {
```
APshenkin: What solution will be for Data.only?

test/data/sandbox/ddt_test.ddt.js
```diff
@@ -1,18 +1,24 @@
+Feature('DDT');
```
APshenkin: Previously user was be able to set datatable for Feature. This implementation doesn't allow to do this. Is this OK?

lib/data/context.js
```diff
@@ -0,0 +1,50 @@
...
+    return {
+      Scenario: function (title, opts, fn) {
+        if (typeof opts === 'function' && !fn) {
+          fn = opts;
+          opts = {};
+        }
+        data.forEach((dataRow) => {
+          opts.data = dataRow;
+          let test = context.Scenario(`${title} | ${dataRow}`, opts, fn);
+          test.inject.current = dataRow;
```
APshenkin: now `current` is not able to use as pageObject name. We have to change pageObject validation here https://github.com/Codeception/CodeceptJS/blob/master/lib/container.js#L142

to block using `current` and allow `scenarioContext`

DavertMik: It will be overridden. For non-data tests you can use `current`...

test/data/sandbox/ddt_test.ddt.js
DavertMik: You can set datatable for feature by setting the same datatable for each Scenario inside a test.
This is much more usable:

```js
Data(accounts).Scenario('do smth with users')
Data(accounts).Scenario('do smth else')
```

lib/data/context.js
DavertMik: sorry, I didn't get this

lib/data/context.js
DavertMik: Added this to documentation but removed check from container.

We need separate concerns, so container should no have hardcoded elements inside it which are related to the different parts of CodeceptJS
the only thing responsible for data management is located in `data/context`.

APshenkin: agreed

test/data/sandbox/ddt_test.ddt.js
APshenkin: agreed

lib/data/context.js
APshenkin: Looks like it's important to have ability to launch only test that contains `Data`

DavertMik: You mean a specific dataset? Sorry I have no ideas for now how to do it (

PR #606, opened 2017-07-17
New Webdriverio methods

APshenkin:
new Methods:
* grabCssPropertyFrom
* seeTitleEquals
* seeTextEquals
* seeCssPropertiesOnElements
* seeAttributesOnElements
* grabNumberOfVisibleElements
* waitInUrl
* waitUrlEquals
* waitForValue
* waitNumberOfVisibleElements
* switchToNextTab
* switchToPreviousTab
* closeCurrentTab
* openNewTab
* refreshPage
* scrollPageToBottom
* scrollPageToTop
* grabBrowserLogs

add url decoding before assertion

### DIFF
--- a/lib/helper/WebDriverIO.js
+++ b/lib/helper/WebDriverIO.js
@@ -993,7 +993,9 @@ class WebDriverIO extends Helper {
       return client.unify(commands, {
         extractValue: true
       }).then((props) => {
-        let values = Object.keys(cssProperties).map(function(key){return cssProperties[key]});
+        let values = Object.keys(cssProperties).map(function (key){
+          return cssProperties[key];
+        });
         if (!Array.isArray(props)) props = [props];
         let chunked = chunkArray(props, values.length);
         chunked = chunked.filter((val) => {
@@ -1001,7 +1003,7 @@ class WebDriverIO extends Helper {
             if (val[i] !== values[i]) return false;
           }
           return true;
-        })
+        });
         return assert.ok(chunked.length === elemAmount,
           `Not all elements (${locator}) have CSS property ${JSON.stringify(cssProperties)}`);
       });
@@ -1031,7 +1033,9 @@ class WebDriverIO extends Helper {
       return client.unify(commands, {
         extractValue: true
       }).then((attrs) => {
-        let values = Object.keys(attributes).map(function(key){return attributes[key]});
+        let values = Object.keys(attributes).map(function (key){
+          return attributes[key];
+        });
         if (!Array.isArray(attrs)) attrs = [attrs];
         let chunked = chunkArray(attrs, values.length);
         chunked = chunked.filter((val) => {
@@ -1039,7 +1043,7 @@ class WebDriverIO extends Helper {
             if (val[i] !== values[i]) return false;
           }
           return true;
-        })
+        });
         return assert.ok(chunked.length === elemAmount,
           `Not all elements (${locator}) have attributes ${JSON.stringify(attributes)}`);
       });
--- a/lib/helper/WebDriverIO.js
+++ b/lib/helper/WebDriverIO.js
@@ -1549,22 +1549,22 @@ class WebDriverIO extends Helper {
     sec = sec || this.options.waitForTimeout;
     context = context || this.root;
     return client.waitUntil(function () {
-        return client.elements(withStrictLocator(context)).then(function (res) {
-          if (!res.value || res.value.length === 0) {
-            return false;
+      return client.elements(withStrictLocator(context)).then(function (res) {
+        if (!res.value || res.value.length === 0) {
+          return false;
+        }
+        let commands = [];
+        res.value.forEach((el) => commands.push(this.elementIdText(el.ELEMENT)));
+        return this.unify(commands, {
+          extractValue: true
+        }).then((selected) => {
+          if (Array.isArray(selected)) {
+            return selected.filter(part => part.indexOf(text) >= 0).length > 0;
           }
-          let commands = [];
-          res.value.forEach((el) => commands.push(this.elementIdText(el.ELEMENT)));
-          return this.unify(commands, {
-            extractValue: true
-          }).then((selected) => {
-            if (Array.isArray(selected)) {
-              return selected.filter(part => part.indexOf(text) >= 0).length > 0;
-            }
-            return selected.indexOf(text) >= 0;
-          });
+          return selected.indexOf(text) >= 0;
         });
-      }, sec * 1000,
+      });
+    }, sec * 1000,
       `element (${context}) is not in DOM or there is no element(${context}) with text "${text}" after ${sec} sec`);
   }
 
@@ -1579,22 +1579,22 @@ class WebDriverIO extends Helper {
     let client = this.browser;
     sec = sec || this.options.waitForTimeout;
     return client.waitUntil(function () {
-        return findFields(client, context).then(function (res) {
-          if (!res.value || res.value.length === 0) {
-            return false;
+      return findFields(client, context).then(function (res) {
+        if (!res.value || res.value.length === 0) {
+          return false;
+        }
+        let commands = [];
+        res.value.forEach((el) => commands.push(this.elementIdAttribute(el.ELEMENT, "value")));
+        return this.unify(commands, {
+          extractValue: true
+        }).then((selected) => {
+          if (Array.isArray(selected)) {
+            return selected.filter(part => part.indexOf(value) >= 0).length > 0;
           }
-          let commands = [];
-          res.value.forEach((el) => commands.push(this.elementIdAttribute(el.ELEMENT, "value")));
-          return this.unify(commands, {
-            extractValue: true
-          }).then((selected) => {
-            if (Array.isArray(selected)) {
-              return selected.filter(part => part.indexOf(value) >= 0).length > 0;
-            }
-            return selected.indexOf(value) >= 0;
-          });
+          return selected.indexOf(value) >= 0;
         });
-      }, sec * 1000,
+      });
+    }, sec * 1000,
       `element (${context}) is not in DOM or there is no element(${context}) with value "${value}" after ${sec} sec`);
   }
 
@@ -1741,9 +1741,10 @@ class WebDriverIO extends Helper {
       .waitUntil(function () {
         return this.getTabIds().then(function (handles) {
           return this.getCurrentTabId().then(function (current) {
-            if (handles.indexOf(current) + num + 1 <= handles.length) return this.switchTab(
+            if (handles.indexOf(current) + num + 1 <= handles.length) {
+              return this.switchTab(
               handles[handles.indexOf(current) + num]);
-            else return false;
+            } else return false;
           });
         });
       }, sec * 1000, `There is no ability to switch to next tab with offset ${num}`);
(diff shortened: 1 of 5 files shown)


lib/helper/WebDriverIO.js
```diff
@@ -924,7 +1002,71 @@ class WebDriverIO extends Helper {
...
+        return assert.ok(selected.length > 0,
+          `Element ${locator} hasn't CSS property ${cssProperty} with value ${value}`);
+      });
+    });
+  }
+
+  /**
+   * Checks that there is an Element with given attribute.
+   */
+  seeAttributeOnElement(locator, attr, value) {
```
DavertMik: actually, this can be merged with `seeElement`, with optional second parameter.
signature can be simplified:

```js
I.seeElement('a', ['href']); // seeks for an element with href attribute
I.seeElement('a', ['href', 'target']); // seeks for an element with href and target
I.seeElement('a', { target: '_blank' } ); // element with attribute target=_blank
```

lib/helper/WebDriverIO.js
```diff
@@ -924,7 +1002,71 @@ class WebDriverIO extends Helper {
...
+        selected = selected.filter((val) => val === value);
+        return assert.ok(selected.length > 0, `Element ${locator} hasn't attribute ${attr} with value ${value}`);
+      });
+    });
+  }
+
+  /**
+   * Grab number of visible elements by locator
+   */
+  grabNumberOfVisibleElements(locator) {
```
DavertMik: should reuse code from `seeNumberOfVisibleElements`
documentation with examples should be added

lib/helper/WebDriverIO.js
```diff
@@ -924,7 +1002,71 @@ class WebDriverIO extends Helper {
...
+        return assert.equal(selected.length, num,
+          `expected number of visible elements (${locator}) is ${num}, but found ${selected.length}`);
+      });
+    });
+  }
+
+  /**
+   * Checks that there is an Element with given CSS property.
+   */
+  seeCssPropertyOnElement(locator, cssProperty, value) {
```
DavertMik: I propose the following syntax: 

```js
seeCssProperty('.nav', { background: '#000', width: '100%' });
```
code examples should be added

lib/helper/WebDriverIO.js
```diff
@@ -724,6 +765,26 @@ class WebDriverIO extends Helper {
   }
 
   /**
+   * Return attribute value from random element, if there is several elements with the same locator
+   */
+  grabAttributeFromRandomElement(locator, attr) {
```
DavertMik: what is the purpose of that? 0.o
I don't see test should rely on something random! 
I'm **completely against this method**.
Please rethink the implementation. You can use *random* in your own test but we should not provide this in core

**Remove this method**

lib/helper/WebDriverIO.js
```diff
@@ -704,6 +726,25 @@ class WebDriverIO extends Helper {
   }
 
   /**
+   * Grab CSS property for given locator
+   */
+  grabCssPropertyFrom(locator, cssProperty) {
```
DavertMik: please add an example

lib/helper/WebDriverIO.js
```diff
@@ -668,6 +670,26 @@ class WebDriverIO extends Helper {
   }
 
   /**
+   * Return text from random element, if there is several elements with the same locator
+   */
+  grabTextFromRandomElement(locator) {
```
DavertMik: no random elements, please. 

you can get the number of visible elements, so use Math.random() inside user code to get the element.
I'd propose to pull out this method from the core as well as grabAttributeFromRandomElement.
We should avoid recommending to use something randomly in tests.

**Remove this method**

lib/helper/WebDriverIO.js
```diff
@@ -1367,6 +1509,55 @@ class WebDriverIO extends Helper {
     }, sec * 1000, `element (${locator}) still not present on page after ${sec} sec`);
   }
 
+
+  /**
+   * Waiting for the part of the URL to match the expected
+   */
```
DavertMik: Please mention that this is very useful for SPA.
JS example is required

lib/helper/WebDriverIO.js
```diff
@@ -1422,6 +1639,29 @@ class WebDriverIO extends Helper {
   }
 
   /**
+   * Waits for a specified number of elements on the page
+   */
+  waitNumberOfVisibleElements(locator, num, sec) {
```
DavertMik: I think the same can be achieved with `waitForVisible`, you just pass XPath with the number of elements....

lib/helper/WebDriverIO.js
```diff
@@ -1496,6 +1736,53 @@ class WebDriverIO extends Helper {
     });
   }
 
+  /**
+   * Switch focus to a particular tab by its number. It waits tabs loading and then switch tab
+   */
+  changeTab(num, amountOfTabs = null, sec = null) {
```
DavertMik: sorry, the behavior is not simple to get. WHy should I declare the `amountOfTabs`? It is not even documented properly.
I'd recommend to use 2 methods:

```js
I.switchToNextTab(); // switch to next tab
I.switchToNextTab(2); // switch to tab +2 from current
```
In code:

```js
switchToNexTab(num = 1, sec = null) {
```

In the same manner `switchToPreviousTab` can be implemented

lib/helper/WebDriverIO.js
```diff
@@ -1496,6 +1736,53 @@ class WebDriverIO extends Helper {
...
+      }, sec * 1000, "The number of tabs hasn't changed to " + amountOfTabs)
+      .getTabIds().then(function (handles) {
+        return this.switchTab(handles[num - 1]);
+      });
+  }
+
+  /**
+   * Close current tab
+   */
+  closeCurrentTab() {
```
DavertMik: please add `openNewTab`. I recommend using 

```js
return this.executeScript(`window.open('about:blank','_blank');`).then(() => this.switchToNextTab());
```

lib/helper/WebDriverIO.js
APshenkin: fixed

lib/helper/WebDriverIO.js
APshenkin: The reason to add this method was that sometimes you have a large amount of similar elements and you want to test that they works as expected. But you don't want to test each element, because of time execution. Because of this you want to use any random element in your test.

Will remove this method.

lib/helper/WebDriverIO.js
APshenkin: example added

lib/helper/WebDriverIO.js
APshenkin: Ok, will remove this method

lib/helper/WebDriverIO.js
APshenkin: I think we should not allow to test several properties in one step. Current implementation takes all elements with given locator and request cssProperty. If all elements will not have expected cssproperty value, then we will throw the error. If we will give opportunity to validate several props in one steps there may appear false positive errors

APshenkin: example was added

lib/helper/WebDriverIO.js
APshenkin: fixed

lib/helper/WebDriverIO.js
```diff
@@ -1496,6 +1736,53 @@ class WebDriverIO extends Helper {
...
+   */
+  refreshPage() {
+    let client = this.browser;
+    return client.refresh();
+  }
+
+  /**
+   * Scroll page to the bottom
+   */
+  scrollPageToBottom() {
```
APshenkin: fixed

lib/helper/WebDriverIO.js
APshenkin: fixed

lib/helper/WebDriverIO.js
APshenkin: fixed

lib/helper/WebDriverIO.js
APshenkin: Yes, you can use XPath for this, but it will be not so descriptive as this method. Also using special XPath force to store two locators for one element (looks like bad practice)

lib/helper/WebDriverIO.js
APshenkin: I don't think that mixing this is good idea. This will lead to large methods which in the future will be more difficult to maintain

lib/helper/WebDriverIO.js
```diff
@@ -1396,6 +1569,36 @@ class WebDriverIO extends Helper {
   }
 
   /**
+   * Waits for the specified value to be in value attribute
+   *
+   * ```js
+   * I.waitForValue("GoodValue", '//input');
+   * ```
+   */
+  waitForValue(value, context, sec = null) {
```
DavertMik: this is inconsistent with `seeInField` or `fillField`. First is locator, second is value. Because locator is required.

lib/helper/WebDriverIO.js
```diff
@@ -1396,6 +1569,36 @@ class WebDriverIO extends Helper {
...
+    let client = this.browser;
+    sec = sec || this.options.waitForTimeout;
+    return client.waitUntil(function () {
+      return client.elements(withStrictLocator(context)).then(function (res) {
```
DavertMik: should use `findFields`

lib/helper/WebDriverIO.js
```diff
@@ -1496,6 +1726,117 @@ class WebDriverIO extends Helper {
...
+  /**
+   * Open new tab and switch to it
+   *
+   * ```js
+   * I.openNewTab();
+   * ```
+   */
+  openNewTab() {
+    let client = this.browser;
+    return client.newWindow('about:blank');
```
DavertMik: this opens a new Window, not a new Tab!

lib/helper/WebDriverIO.js
```diff
@@ -913,9 +963,74 @@ class WebDriverIO extends Helper {
...
+  }
+
+  /**
+   * Checks that there is an Element with given CSS property.
+   *
+   * ```js
+   * I.seeCssPropertyOnElement('h3', 'font-weight', "bold");
+   * ```
+   */
+  seeCssPropertyOnElement(locator, cssProperty, value) {
```
DavertMik: I still think this should accept cssProperties and values as object.
If user needs o check several CSS properties he will need to call `seeCssPropertyOnElement` several times, which 
will do extra job locating an element. 

Also, I don't think that we should check all possible elements on a page for having CSS property.
If I search for `h3 font-weight:bold` I want to ensure that h3 headings on page are bold. But I can also locate `h3.bold` element which adds boldness to text. Then it makes my test false-positive. I thought my h3s are bold but they are not! 

So this should go to the first element only!
So users would prepare more exact locators.

lib/helper/WebDriverIO.js
```diff
@@ -704,6 +706,29 @@ class WebDriverIO extends Helper {
   }
 
   /**
+   * Grab CSS property for given locator
+   *
+   * ```js
+   * I.grabCssPropertyFrom('h3', 'font-weight');
+   * ```
+   */
+  grabCssPropertyFrom(locator, cssProperty) {
```
DavertMik: Same comment as for `seeCssPropertyOnElement`. 
This behavior is absolutely unpredictable in current implementation. `font-weight` of which element I receive when calling ` I.grabCssPropertyFrom('h3', 'font-weight');` ?

lib/helper/WebDriverIO.js
APshenkin: It opens a new Tab, You can test this :)

lib/helper/WebDriverIO.js
APshenkin: fixed

lib/helper/WebDriverIO.js
APshenkin: fixed

lib/helper/WebDriverIO.js
APshenkin: Now it will check that all elements have given properties

lib/helper/WebDriverIO.js
APshenkin: We leave it as it is now implemented. If there is several h3 on page it will returns array with h3 `font-weight` values

PR #608, opened 2017-07-18
Modernize JS a bit

shvaikalesh:


DavertMik:
Thanks. Good to know that some dirty hacks are needed no longer.

PR #612, opened 2017-07-20
Smart Waits

DavertMik:
* fixed returning promise in `defineTimeout`
* fixed respecting context in `click`, `doubleClick`, `checkOption`
* added context test for `click`
* SmartWaits introduced. [What are SmartWaits](http://codeception.com/docs/03-AcceptanceTests#SmartWait)

If locator is defined and element is expected on page, we can wait for few more seconds before failing a test.

```js
I.click('#click');
I.click('Home', '#nav');
I.seeElement('#user');
```

Doesn't work with wide scope locators like:

```js
I.click('Hello');
```

Smart wait is disabled by default.

### DIFF
--- a/lib/helper/WebDriverIO.js
+++ b/lib/helper/WebDriverIO.js
@@ -1076,7 +1076,7 @@ class WebDriverIO extends Helper {
    * ```
    */
   grabNumberOfVisibleElements(locator) {
-    return this.browser.elements(locator).then(function (res) {
+    return this.browser.elements(locator).then((res) => {
       if (!res.value || res.value.length === 0) {
         return 0;
       }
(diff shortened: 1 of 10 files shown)


lib/helper/WebDriverIO.js
```diff
@@ -1702,4 +1712,18 @@ function isFrameLocator(locator) {
   return locator[key];
 }
 
+function prepareLocateFn(context) {
+  if (!context) return this._locate.bind(this);
+  let el;
```
APshenkin: Don't understand, where we define `el`? Could you please give some comments about it?

DavertMik: Sure, el is defined few lines below:

```js
return this.browser.elementIdElements(el = res.value[0].ELEMENT, l);
```
This is done to cache the context element and reuse it if it was already found

lib/helper/WebDriverIO.js
```diff
@@ -380,8 +383,18 @@ class WebDriverIO extends Helper {
...
-    return this.browser.elements(withStrictLocator(locator));
+  _locate(locator, smartWait = false) {
+
+    if (!this.options.smartWait || !smartWait) return this.browser.elements(withStrictLocator(locator));
+    let els;
+
+    return this.defineTimeout({implicit: this.options.smartWait})
+      .then(() => this.debugSection('SmartWait', `Locating ${locator} in ${this.options.wait}`))
+      .then(() => {
+        return els = this.browser.elements(withStrictLocator(locator));
```
DavertMik: I agree to this point but implicit wait doesn't check for visibility. 
If an element is on a page but is not visible better to fail with error "Element is invisible".
Otherwise, user may think that SmartWait just doesn't work.

PR #614, opened 2017-07-20
Web specific locators for Appium

DavertMik:
To have one test for mobile and web app.

* added `runInWeb` method to execute code only in web
* added `web` locator type similarly to `android` and `ios`

### DIFF
--- a/lib/helper/Appium.js
+++ b/lib/helper/Appium.js
@@ -9,6 +9,9 @@ const path = require('path');
 const requireg = require('requireg');
 const fs = require('fs');
 const recorder = require('../recorder');
+const isGenerator = require('../utils').isGenerator;
+const resumeTest = require('../scenario').resumeTest;
+
 
 const mobileRoot = '//*';
 const webRoot = 'body';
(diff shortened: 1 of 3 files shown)


lib/helper/Appium.js
```diff
@@ -276,6 +276,26 @@ class Appium extends WebdriverIO {
...
+   * ```js
+   * I.runInWeb(() => {
+   *    I.waitForElement('#data');
+   *    I.seeInCurrentUrl('/data');
+   * });
+   * ```
+   *
+   * @param {*} fn
+   */
+  runInWeb(fn) {
```
APshenkin: will this works with generators? Think, that we should change generators to awync/await, when Node 8 will be LTS

lib/helper/Appium.js
```diff
@@ -276,6 +276,26 @@ class Appium extends WebdriverIO {
...
+    if (!this.isWeb) return;
+    recorder.session.start('Web-only actions');
+    fn();
+    recorder.add('restore from Web session', () => recorder.session.restore());
```
APshenkin: I think, we should have tests for this cool features like different locators for different platforms, execution only on specified platform. This is killer feature and we have to be sure that it's works after each update

DavertMik: Yes, there are unit tests for Android / iOS. I will add few more for Web

lib/helper/Appium.js
DavertMik: Oh, generators... I will try to apply them.

PR #615, opened 2017-07-20
Fix the "waitForTimeout" typo in configuration.md

Nierrrrrrr:
It should be capital 'F' instead of 'f' since the #387


DavertMik:
Yes, you are right )

PR #626, opened 2017-07-23
Fix misprint in smart wait log message

BorisOsipov:
Fix misprint in smart wait log message -
[SmartWait] Locating //*['test'] in **undefined**


DavertMik:
Thanks.
Btw, how do you find smart waits?

BorisOsipov:
@DavertMik I use my fork as dependency in my tests in order to get bugfixes earlier. So I read git log after update my fork to get know what is new :)

PR #643, opened 2017-08-02
Adding the French language

Gynidark:
Hello

I hope you are well, I just noticed that there was no French language, so I allowed myself to do a translation in French. :blush:

Good day to you !

Gynidark.


DavertMik:
Thanks. However, translation should provide JavaScript-valid names for functions

The intention is to write code as this:

```js
Je.clique();
Je.doubleClique();
// not Je['Double clique'] - which looks a bit ugly :|
```
depending on language you can vary camel case or snake case but you should choose one.

DavertMik:
No updates :(
I think I should close this PR. Please resend it when you change the naming

PR #645, opened 2017-08-02
Attempting to re-add Istanbul coverage to repo

KennyRules:


DavertMik:
We have mixed setup with unit / acceptance tests and istanbul would be able to cover only unit tests. However, that would be useful to see current Helpers coverage. The only thing I'm not aware of - how to use it? 😉

KennyRules:
Sweet - right now, it actually will always run on the test gulp task.

It creates a `coverage` folder which is already being .gitignore'd in the repo which has all the report files.

The console output for the test will also show the coverage info :)

I'd be thrilled to add any more relevant documentation/tasks, just not sure what you have in mind!

DavertMik:
Is there a way to see the graphical representation of the coverage?
Metrics by themselves are not so useful but seeing which parts should be improved is a must

KennyRules:
For sure! - Here is the console output after the tests run:
console-output

And here is the index.html that istanbul generates in coverage/lcov-report/index.html 

coverage-report-index

I do want to note this is the default Istanbul / Gulp behavior, so as to keep the PR as unopinionated as possible!

DavertMik:
Nice. Thanks, that's might be pretty helpful!

Issue #528, opened 2017-05-19
Cannot read property 'url' of undefined in Nightmare.js


bendiggle:
#### What are you trying to achieve?
I am trying go to another page using `I.amOnPage(url)`.
In my test my steps are:
```
    I.amOnPage(url); // this amOnPage works fine
    I.checkOption(element);
    I.amOnPage(anotherUrl); // this one doesn't
```

#### What do you get instead?
There error I get is 
`Cannot read property 'url' of undefined`

I have narrowed it down to the `amOnPage` function in Nightmare.js file in `codeceptjs/lib/helper/Nightmare.js`.
```
    return this.browser.goto(url, headers).then((res) => {
      this.debugSection('URL', res.url);
      this.debugSection('Code', res.code);
      this.debugSection('Headers', JSON.stringify(res.headers));
    });
```
the `res` being returned from the promise is coming back as `undefined`.

implico:
Is `url == anotherUrl`? If so, a temporary workaround would be to navigate to an intermediate page, e.g. `I.amOnPage('about:blank')`

bendiggle:
@implico Yes, url is equal to anotherUrl! (Sorry, I should have mentioned). All I really want to do is refresh the page but could not find a way to do it so was trying to just go to that page again.
Thanks, I will try that.

implico:
Yes, I was 99% sure it is. This will work, not an elegant solution though, to be fixed later.

APshenkin:
@bendiggle You can create custom helper with refresh method https://github.com/segmentio/nightmare#refresh

bendiggle:
Thanks @APshenkin, I will do that!

DavertMik:
We need to get `refresh` method in Nightmare helper as well.
Who will send a pull request? @bendiggle ?

awshanks:
This is now in release 1.0.1.  Can this now be closed?

APshenkin:
@awshanks yep, will close it

petrisorionel:
Hello, I found this problem on version 1.0.2

lougreenwood:
I'm also still seeing this bug:

```
Error: Evaluation timed out after 30000msec.  Are you calling done() or resolving your promises?
 ✖ "before each" hook: Before for "test logout" in 30034ms

-- FAILURES:

  1) Login page: "before each" hook: Before for "test login form":
     Cannot read property 'url' of undefined

  Scenario Steps:

[...]

PR #650, opened 2017-08-06
Adding refresh function to Nightmare Helper - Issue 528

awshanks:
Added refresh function requested in Issue 528
https://github.com/Codeception/CodeceptJS/issues/528

### DIFF
--- a/test/helper/Nightmare_test.js
+++ b/test/helper/Nightmare_test.js
@@ -137,6 +137,18 @@ describe('Nightmare', function () {
           });
       });
     })
-  })
+  });
+
+  describe('refresh page', () => {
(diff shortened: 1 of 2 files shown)


test/helper/Nightmare_test.js
```diff
@@ -137,6 +137,18 @@ describe('Nightmare', function () {
...
     })
-  })
+  });
+
+  describe('refresh page', () => {
+    it('should refresh the current page', function*() {
+      I.amOnPage(site_url);
+      let url = yield browser.url();
+      assert.equal(site_url + '/', url);
+      I.refresh();
```
DavertMik: should be `yield I.refresh` here; we need to handle promises manually here

Issue #647, opened 2017-08-03
I.click using inner text does not working properly


Jotohanes:
#### What are you trying to achieve?
I would like to click certain text in particular locator.
#### What do you get instead?
 Clickable element JETZT KAUFEN inside .mscom-link.c-call-to-action.c-glyph was not found by text|CSS|XPath

I am pretty sure it was not the case on previous version of codeceptjs since I got this error recently after I update my codeceptjs.

```shells
  Clickable element JETZT KAUFEN inside .mscom-link.c-call-to-action.c-glyph was not found by text|CSS|XPath
```

> Provide test source code if related

```js

Scenario('test something', (I) => {

    I.amOnPage('https://www.microsoft.com/de-de/');
    I.waitForVisible('.mscom-link.c-call-to-action.c-glyph');
    I.click('JETZT KAUFEN','.mscom-link.c-call-to-action.c-glyph');

});
    
```

### Details

* CodeceptJS version: 1.0.0
* NodeJS Version:  6.11.0
* Operating System: Windows 7 Pro, 2009
* WebDriverIO 
* Configuration file:

```js
{
  "tests": "./*_test.js",
  "timeout": 10000,
  "output": "./output",
  "helpers": {
    "WebDriverIO": {
      "url": "http://localhost",
      "browser": "chrome"
    }
  },
  "include": {
    "I": "./steps_file.js"
  },
  "bootstrap": false,
  "mocha": {},
  "name": "test"
}
```

APshenkin:
@Jotohanes so, the problem is that `.mscom-link.c-call-to-action.c-glyph` element is link itself with  necessary text. When we add context in click, then the flow is:
1. Recieve all elements `context` and get first of them.
2. Get first inner element with text, using these locators:
```
let narrowLocator = xpathLocator.combine([
    `.//a[normalize-space(.)=${literal}]`,
    `.//button[normalize-space(.)=${literal}]`,
    `.//a/img[normalize-space(@alt)=${literal}]/ancestor::a`,
    `.//input[./@type = 'submit' or ./@type = 'image' or ./@type = 'button'][normalize-space(@value)=${literal}]`
  ]);

or
let wideLocator = xpathLocator.combine([
      `.//a[./@href][((contains(normalize-space(string(.)), ${literal})) or .//img[contains(./@alt, ${literal})])]`,
      `.//input[./@type = 'submit' or ./@type = 'image' or ./@type = 'button'][contains(./@value, ${literal})]`,
[...]

PR #652, opened 2017-08-07
[WebdriverIO] fix inner click

APshenkin:
fix https://github.com/Codeception/CodeceptJS/issues/647


APshenkin:
@DavertMik I had fixed this issue. The reason of issue was that when we tried to get all elements from context element with necessary text then we got context element itself and inner element with text. Then we choose first element and tried to click on it. But it was not element with link just `p` element. So I moved validating self element to the end to fix this issue

Issue #636, opened 2017-07-28
Enhancement: Allow file instead of string for --override flag


nathand8:
#### What are you trying to achieve?
I want to be able to store codecept config overrides in a separate json **file**.
```bash
codeceptjs run --override codecept-custom.json
```

codecept-custom.json
```json
{
    "helpers": {
        "Protractor": {
            "url": "http://build.example.com"
        }
    }
}
```

#### What do you get instead?
JSON parse error

### Details

* CodeceptJS version: 1.0.0
* NodeJS Version: v8.1.4
* Operating System: Mac
* Protractor
* Configuration file:

```js
{
  "output": "./output",
  "helpers": {
    "Protractor": {
      "url": "http://build04.example.com:8009",
      // "url": "http://localhost",
      "driver": "hosted",
      "browser": "chrome",
      "rootElement": "body"
    }
  },
  "mocha": {},
  "bootstrap": false,
  "teardown": null,
  "hooks": [],
  "tests": "./test/*/*.js",
  "timeout": 10000,
  "name": "CodeceptTests"
}
```

DavertMik:
Right now you can use different config file, with `--config` option
You can have different `codecept.conf.js` files and use them depending on a situation. 
Also you can use their parts by moving common logic to shared file and requiring it as a module.
Anyway, JS-based configs are so flexible that you may not need `--override` option:

See http://codecept.io/configuration/#dynamic-configuration

nathand8:
Excellent suggestion. We can create an "override file" and import that to `codecept.conf.js`. Then we can use [extend](https://www.npmjs.com/package/extend) to overwrite those values.

This essentially gives us the same effect. I'll close this ticket because that workaround is good enough for me.

If anyone needs a pattern to follow, here's my `codecept.conf.js`
```javascript
var extend = require('extend');

var defaultConfig = {
  // Default config
};

var configOverrides = require('./custom-config') || {};

exports.config = extend(defaultConfig, configOverrides);
```

PR #655, opened 2017-08-08
Add documentation for --config

StephanBijzitter:
I was looking for this option on the website and I could not find it. Then I went to Google and I still could not find it. Finally I started looking through Github's issue tracker and I finally found the information I needed in issue #636.

I believe this information is important to have on the configuration page, as it's quite literally in the name.


DavertMik:
Yes, thanks! Looks like I forgot to mention that in nice way

Issue #206, opened 2016-08-10
Before/After don't support generators


joefitter:
Passing a generator to `Before` or `After` doesn't have the same behaviour as if used as a `Scenario`. It seems to skip over the `Before` and immediately run the tests

DavertMik:
Yes, it is not supported and documented for Before/After. I didn't even think of such usage.
To be honest, it is pretty hard to handle generators inside tests so I don't think this feature is easy to implement.

Nighthawk14:
👍

PR #663, opened 2017-08-11
Hooks generators

APshenkin:
* https://github.com/Codeception/CodeceptJS/issues/206 Add generators support in scenario hooks BeforeSuite/Before/AfterSuite/After;
* Add Scenario Steps output for hooks, when they fails

### DIFF
--- a/lib/event.js
+++ b/lib/event.js
@@ -15,6 +15,10 @@ module.exports = {
     before: 'suite.before',
     after: 'suite.after'
   },
+  hook: {
+    started: 'hook.start',
+    passed: 'hook.passed'
+  },
   step: {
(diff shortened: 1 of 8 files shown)


DavertMik:
Can we just not call them hooks? The term is [already in use](codecept.io/hooks/)

APshenkin:
@DavertMik This term is from Mocha. Yep, we can change it, lets try to figure out suitable term

DavertMik:
Ok, I couldn't think anything better :(
Let's have them as hook. If PR is ready we can merge it.

APshenkin:
@DavertMik Yes, It's ready

Issue #447, opened 2017-03-20
Tests fail silently if they have the same name as other tests


lennym:
We have a suite which has a number of tests named "Happy Path".

We recently discovered that if some of them fail, then the test run still exits with a zero exit code, and our CI build passes.

This seems to be a result of this code-
 https://github.com/Codeception/CodeceptJS/blob/master/lib/listener/exit.js#L19 - which removes a test from the failures list if a test with the same name later passes.

I'd recommend doing something a bit stricter than purely name matching in this case (possibly name + file, or generating a unique id for a suite/test) so that it is not possible for test failures to be ignored.

Happy to PR if you have a preferred approach.

PR #675, opened 2017-08-23
Fix unique names issue

APshenkin:
* Fix https://github.com/Codeception/CodeceptJS/issues/447 - Tests fail silently if they have the same name as other tests. Now we will track test uuid to monitor failed/passed tests
* Now we are using test uuid in screenshot names, when `uniqueScreenshotNames: true`
* Remove obsolete util `HashCode`

### DIFF
--- a/lib/helper/Mochawesome.js
+++ b/lib/helper/Mochawesome.js
@@ -57,9 +57,8 @@ class Mochawesome extends Helper {
       fileName = clearString(test.title);
     }
     if (this.options.uniqueScreenshotNames) {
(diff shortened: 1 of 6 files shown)


DavertMik:
UUID is provided by mocha? 
Does it generating every new UUID for each new test run?

APshenkin:
@DavertMik Yes, Mocha generates UUID for each tests and hook. If you set `retries` for tests then every retry will have the same UUID. Mocha generates new UUID for test each new test run

Issue #458, opened 2017-04-06
amOutsideAngularApp is needed before all lines


kristho91:
#### What are you trying to achieve?
I am trying to test a non-angular page with CodeceptJS and Protractor. I expect amOutsideAngularApp to hold it's value until I run amInsideAngularApp. But it does not - it seems.
It works if I add amOutsideAngularApp before the I.see(xxx) line too.

> Provide console output if related. Use `--verbose` mode for more details.

```bash
CodeceptJS v0.5.1
Using test root "C:\****"

Testing something basic --
   [1] Starting recording promises
   Emitted | suite.before
   [1] Queued | hook Protractor._beforeSuite()
 test something
 > [Session] Starting singleton browser session
   [2] Starting recording promises
   Emitted | test.before
   [2] Queued | hook Protractor._before()
   Emitted | test.start ([object Object])
   Emitted | step.before (I am outside angular app )
   [2] Queued | amOutsideAngularApp: 
   Emitted | step.after (I am outside angular app )
   [2] Queued | return step result
   Emitted | step.before (I am on page "/")
   [2] Queued | amOnPage: "/"
   Emitted | step.after (I am on page "/")
   [2] Queued | return step result
   Emitted | step.before (I see "Login")
   [2] Queued | see: "Login"
   Emitted | step.after (I see "Login")
   [2] Queued | return step result
   [2] Queued | fire test.passed
   [2] Queued | finish test
   Emitted | step.start (I am outside angular app )
 * I am outside angular app 
   Emitted | step.start (I am on page "/")
 * I am on page "/"
   Emitted | step.start (I see "Login")
 * I see "Login"
   [2] Error | Error: Error while waiting for Protractor to sync with the page: "window.angular is undefined.  This could be either because this is a non-angular page or because your test involves client-side navigation, which can interfere with Protractor's bootstrapping.  See http://git.io/v4gXM for details"
   [2] Starting <teardown> session
   Emitted | test.failed ([object Object])
   [2] <teardown> Queued | hook Protractor._failed()
   [2] <teardown> Queued | () => done(err)
   [2] <teardown> Stopping recording promises
 > Screenshot has been saved to C:\****\test_something.failed.png
 ├ù FAILED in 5372ms

   Emitted | test.after
   [2] <teardown> Queued | hook Protractor._after()
 > [Session] cleaning cookies and localStorage
   [3] Starting recording promises
   Emitted | suite.after
   [3] Queued | hook Protractor._afterSuite()

[...]

PR #676, opened 2017-08-23
fix Protractor reinject outside app

APshenkin:
Fix https://github.com/Codeception/CodeceptJS/issues/458 - amOutsideAngularApp is needed before all lines when test non-Angular app


DavertMik:
Thanks!

Issue #559, opened 2017-06-18
"codeceptjs run --debug-brk" does nothing


mo:
"./node_modules/.bin/codeceptjs run --debug-brk" runs my test as usual without stopping on an initial breakpoint and no debugger URL is printed either.

DavertMik:
I'm not sure why should have it worked the way you wanted. There is no `--debug-brk` option in command

Please tell more about how you want it to debug

mo:
I found it in "--help" (see screenshot)

mo:
![image](https://user-images.githubusercontent.com/7117/27400407-49bce8f2-56c0-11e7-8644-4c757fc5b5f8.png)

DavertMik:
Oh, looks like this is inherited from mocha framework but it is not used :(
Sorry

PR #681, opened 2017-08-24
Add nodeJs debugger support

APshenkin:
fix https://github.com/Codeception/CodeceptJS/issues/559 - "codeceptjs run --debug-brk" does nothing

Now there is two options --debug-brk and --inspect that passes to NodeJs and allow you to run NodeJs inspectors

### DIFF
--- a/bin/_codeceptjs
+++ b/bin/_codeceptjs
@@ -1,0 +1,108 @@
+#!/usr/bin/env node
+'use strict';
+
+var program = require('commander');
+var path = require('path');
+var Config = require('../lib/config');
+var Codecept = require('../lib/codecept');
+var print = require('../lib/output');
+var fileExists = require('../lib/utils').fileExists;
+var fs = require('fs');
+
+program.command('init [path]')
+  .description('Creates dummy config in current dir or [path]')
+  .action(require('../lib/command/init'));
+
+program.command('shell [path]')
+  .alias('sh')
+  .description('Interactive shell')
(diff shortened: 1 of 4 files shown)


DavertMik:
How stable is such approach?
Will process won't stuck and can it be interrupted by Ctrl-C?

DavertMik:
I think this feature needs some testing before going to release

APshenkin:
@DavertMik 
You can interrupt process by Ctrl-C in debug mode. Process will stuck on first line of code to give user opportunity to set breakpoints in test code.

I take implementation of this feature from mocha framework https://github.com/mochajs/mocha/blob/master/bin/mocha. So I think it's save to use this feature

APshenkin:
@DavertMik You are right, we have to test it deeply. These changes break debug mode in Webstorm. I'll investigate this

Vorobeyko:
When will this feature be implemented?
I am really want this feature(

jlewisCode:
any updates to this or does this need to be picked up by someone?

DavertMik:
@APshenkin are you still on this?
Maybe someone could take it?

DavertMik:
According to reports in Slack, debugger works even without this PR :v:

pstoica:
You can't pass `debug-brk` to the codecept command itself. My goal was to reuse my existing npm scripts with VSCode's debug launchers, but you have to rewrite the commands to use `node $NODE_DEBUG_OPTIONS codeceptjs`. Perhaps this is worth revisiting?

Issue #682, opened 2017-08-24
Introduce waitWhileSelector 


sabau:
#### What are you trying to achieve?
I would like to be able to wait while an element is present. 
For example I usually need to wait while a spinner is present to go on.

#### What do you get instead?
I have to reply a feature I was using in casperJs:
```js
Casper.prototype.waitWhileSelector = function waitWhileSelector(selector, then, onTimeout, timeout) {
    "use strict";
    this.checkStarted();
    timeout = getTimeoutAndCheckNextStepFunction(timeout, then, 'waitWhileSelector', this.options.waitTimeout);
    return this.waitFor(function _check() {
        return !this.exists(selector);
    }, then, onTimeout, timeout, {
        selector: selector,
        waitWhile: true
    });
};
```

### Details

* CodeceptJS version: 1.0.1
* NodeJS Version: 6.11.2
* Operating System: Windows 7 (sadly)
* Nightmare version: 2.10.1
* Configuration file:

```js
{
  "output": "logs",
  "helpers": {
    "Nightmare": {
      "url": "http://127.0.0.1:8080",
      "width": 1900,
      "height": 1200,
      "waitTimeout": 20000,
      "waitForAction": 2000,
      "waitForTimeout": 20000
    }
  },
  "include": {
    "I": "./steps_file.js",
    "loginPage": "./pages/Login.js",
    "mainPage": "./pages/Main.js",
    "mainMenuFragment": "./fragments/MainMenu.js",
    "navbarFragment": "./fragments/Navbar.js"
  },
  "mocha": {},
  "bootstrap": false,
  "teardown": null,
  "hooks": [],
  "tests": "tests/*_test.js",
  "timeout": 90000,
  "name": "codecept"
}
```

sivnz:
+1

sabau:
Feature introduced under the name waitUntilExists with [pull request](https://github.com/Codeception/CodeceptJS/pull/683)

PR #683, opened 2017-08-25
waitWhileElement

sabau:
Hi, 
as described in the issue [waitWhileElement](https://github.com/Codeception/CodeceptJS/issues/682) I was trying to replicate a feature I was used to. I hope to have covered everything from helpers to docs update to tests coverage.

I hope it can be a useful feature. I often have to wait random amount of time due to slow DB, high loads, different environments to be tested, usually some graphical elements disappears (modals or spinner).

Best regards,
Karoly

### DIFF
--- a/docs/helpers/Nightmare.md
+++ b/docs/helpers/Nightmare.md
@@ -752,14 +752,14 @@ Element can be located by CSS or XPath.
 -   `locator`  element located by CSS|XPath|strict locator
 -   `sec`  time seconds to wait, 1 by default
 
-## waitWhileElement
+## waitUntilExists
 
 Waits for element not to be present on page (by default waits for 1sec).
 Element can be located by CSS or XPath.
 
 ```js
-I.waitWhileElement('.btn.continue');
-I.waitWhileElement('.btn.continue', 5); // wait for 5 secs
+I.waitUntilExists('.btn.continue');
+I.waitUntilExists('.btn.continue', 5); // wait for 5 secs
 ```
 
 **Parameters**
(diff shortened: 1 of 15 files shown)


DavertMik:
Thanks, for the comprehensive implementation and tests
However, I don't find the naming really good. Wait whileElement sounds to abstract to me, I'd never guess what it waits for before I go to docs. I think it should be much more descriptive. How about 
`waitToDisappear` or `waitForElementToDisappear`? Maybe you could have another good ideas? 

Also please fix the SeleniumWebdriver implementation. Thanks

sabau:
True, is just took it as it was. Disappear maybe can be confused with the concept of visibility, what about `waitForElementRemoval` or `waitUntilElementRemoved`?

Sorry, I missed that SeleniumWebdriver bug, I'll fix it as soon as I can.

APshenkin:
@sabau @DavertMik There is already implementation for WebdriverIO - `waitForInvisible` and `waitForStalenessOf`
I think we can use this names for methods

sabau:
Hi @APshenkin, 
I was looking at them, they are very similar, I think the main differences are:

- `waitForStalenessOf` is expecting the element to be present and disappear after a while, but we can't be sure everytime that the element is there.
- `waitForInvisible` is satisfied both when all elements are removed or when all elements become invisible, it covers this new call, but have a wider scope.

But looking at WebdriverIO/jquery I was missing the classical name `waitUntilExists`

DavertMik:
Yes, `waitUntilExists` looks better to me

sabau:
Hi @DavertMik, sorry for the long waiting. 
Ok, then I'll fix the naming ASAP, hopefully today.

DavertMik:
Thank you!

PR #688, opened 2017-08-28
Lint fixes

KennyRules:
Tightened up existing code per linter warnings - For example, using triple equality checks or wrapping anonymous function in brackets in certain circumstances.

Touched up some typos.

Added some unit tests for data table and output.

Fixed a bug with output class where level could be undefined.


DavertMik:
Wow! Thank you very very much

DavertMik:
@APshenkin no objections?

Issue #571, opened 2017-06-23
Run-multiple - Output directory path appended to test root path by codecept, So unable to create output directories


shrikanthkm07:
#### What are you trying to achieve?

> After upgrading to codeceptjs@0.6.3 for run-multiple

I am using **run-multiple** command for parallel execution. I am using mocha-reporter to generate output xml file. So I pass my custom directory path in **`output: /Users/madhusus/hydra-dev/src/node_modules/consumer-automation/test-reports`**
I am expecting the report file to be saved in this path.

#### What do you get instead?
Internally codecept is appending the "test root path" which is  `/Users/madhusus/hydra-dev/src/node_modules/nextgen-test/` to the path which I have passed in **`output`**.

So the end result for output directory will be **`'/Users/madhusus/hydra-dev/src/node_modules/nextgen-test/Users/madhusus/hydra-dev/src/node_modules/consumer-automation/test-reports/desktop_browser_chrome_1'`** which is invalid.
> Provide console output if related. Use `--verbose` mode for more details.

```bash
creating output directory: /Users/madhusus/hydra-dev/src/node_modules/nextgen-test/Users/madhusus/hydra-dev/src/node_modules/consumer-automation/test-reports/desktop_browser_chrome_1
fs.js:922
  return binding.mkdir(pathModule._makeLong(path),
                 ^

Error: ENOENT: no such file or directory, mkdir '/Users/madhusus/hydra-dev/src/node_modules/nextgen-test/Users/madhusus/hydra-dev/src/node_modules/consumer-automation/test-reports/desktop_browser_chrome_1'
    at Error (native)
    at Object.fs.mkdirSync (fs.js:922:18)
    at Command.module.exports (/Users/madhusus/hydra-dev/src/node_modules/nextgen-test/node_modules/codeceptjs/lib/command/run.js:26:8)
    at Command.listener (/Users/madhusus/hydra-dev/src/node_modules/nextgen-test/node_modules/commander/index.js:301:8)
    at emitTwo (events.js:106:13)
    at Command.emit (events.js:191:7)
    at Command.parseArgs (/Users/madhusus/hydra-dev/src/node_modules/nextgen-test/node_modules/commander/index.js:615:12)
    at Command.parse (/Users/madhusus/hydra-dev/src/node_modules/nextgen-test/node_modules/commander/index.js:458:21)
[...]

PR #689, opened 2017-08-29
Absolute output fix

APshenkin:
Fix https://github.com/Codeception/CodeceptJS/issues/571 - Unable to pass absolute path for output

### DIFF
--- a/lib/command/run.js
+++ b/lib/command/run.js
@@ -3,6 +3,7 @@ const getConfig = require('./utils').getConfig;
 const getTestRoot = require('./utils').getTestRoot;
 const deepMerge = require('./utils').deepMerge;
 const fileExists = require('../utils').fileExists;
+const detectAbsolutePath = require('../utils').detectAbsolutePath;
 const path = require('path');
 const mkdirp = require('mkdirp');
 const Config = require('../config');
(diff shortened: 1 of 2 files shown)


lib/command/run.js
```diff
@@ -21,7 +21,10 @@ module.exports = function (test, options) {
     config = Config.append(JSON.parse(options.override));
   }
 
-  if (!fileExists(outputDir = path.join(testRoot, config.output))) {
+  if (config.output.startsWith('/') || config.output.startsWith('\\')) outputDir = config.output;
```
DavertMik: windows? ;)
Please also move this part to `utils`, like `detectAbsolutePath`

APshenkin: @DavertMik Yep, this checks absolute path for Windows and Unix systems. Will move this to utils

lib/utils.js
```diff
@@ -36,6 +36,11 @@ module.exports.fileExists = function (filePath) {
   return true;
 };
 
+module.exports.detectAbsolutePath = function (filePath) {
+  if (filePath.startsWith('/') || filePath.startsWith('\\')) return true;
```
DavertMik: what about `C:\Users\...` ?

PR #697, opened 2017-09-01
BeforeSuite addMochawesome Context support and DDT improve

APshenkin:
BeforeSuite addMochawesome Context support
Add `xadd` for datatables to ignore this tests in run
Add `Data(table).only.Scenario` to give ability to launch only `Data` tests

### DIFF
--- a/lib/data/table.js
+++ b/lib/data/table.js
@@ -31,7 +31,7 @@ class DataTable {
       tempObj.toString = () => JSON.stringify(tempObj);
       arrayCounter++;
     });
-    this.rows.push({ skip: true, data: tempObj})
+    this.rows.push({ skip: true, data: tempObj});
   }
 
 }
--- a/lib/data/table.js
+++ b/lib/data/table.js
@@ -19,7 +19,19 @@ class DataTable {
       tempObj.toString = () => JSON.stringify(tempObj);
       arrayCounter++;
     });
-    this.rows.push(tempObj);
+    this.rows.push({ skip: false, data: tempObj});
+  }
+
+  xadd(array) {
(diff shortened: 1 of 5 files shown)


lib/helper/Mochawesome.js
```diff
@@ -67,6 +68,7 @@ class Mochawesome extends Helper {
   }
 
   addMochawesomeContext(context) {
+    if (currentTest = "") currentTest = { test: currentSuite.ctx.test }
```
jploskonka: Hmm, shouldn't this be `currentTest ==`?  Or am I missing something? :)

APshenkin: Yep will fix this)

lib/helper/Mochawesome.js
```diff
@@ -32,6 +32,7 @@ class Mochawesome extends Helper {
 
   _beforeSuite(suite) {
     currentSuite = suite;
+    currentTest = ""
```
DavertMik: `;` - in the end :)

lib/data/table.js
```diff
@@ -19,7 +19,19 @@ class DataTable {
       tempObj.toString = () => JSON.stringify(tempObj);
       arrayCounter++;
     });
-    this.rows.push(tempObj);
+    this.rows.push({ skip: false, data: tempObj});
+  }
+
+  xadd(array) {
```
DavertMik: where is xadd used?
probably it should be documented, right?

APshenkin: Updated.

DavertMik:
Thanks! Looks good to me

PR #699, opened 2017-09-02
Typings

KennyRules:
* Added TypeScript compiler / configs to check the JavaScript.
* Fixed all resulting compiler warnings. This was mostly just optional parameter edge-cases, moving class property initializations to constructors, and pre-pending `global` or `window` when appropriate.
* Added `index.d.ts` to project and package.json. Currently just contains some global values and types for Mocha.
* Added `type-check` gulp task to run the compiler - executable via `npm run gulp type-check`. Task is not run by default, but can easily be added if so desired 😃 
* Touched up spelling when encountered.

I intend to work on some type definitions in a subsequent PR - I did not want to combine substantial type definitions and compiler fixes in a single PR, just in case.

Thanks so much for the awesome framework, @DavertMik & co!

### DIFF
--- a/lib/helper/Protractor.js
+++ b/lib/helper/Protractor.js
@@ -228,37 +228,37 @@ class Protractor extends SeleniumWebdriver {
   /**
    * {{> ../webapi/waitForElement }}
    */
-  waitForElement(locator, aSec = null) {
-    let sec = aSec || this.options.waitForTimeout;
+  waitForElement(locator, sec = null) {
+    let aSec = sec || this.options.waitForTimeout;
     let el = global.element(guessLocator(locator) || global.by.css(locator));
-    return this.browser.wait(EC.presenceOf(el), sec * 1000);
+    return this.browser.wait(EC.presenceOf(el), aSec * 1000);
   }
 
   /**
    * Waits for element to become clickable for number of seconds.
    */
-  waitForClickable(locator, aSec = null) {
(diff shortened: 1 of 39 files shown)


lib/helper/Protractor.js
```diff
@@ -227,36 +228,36 @@ class Protractor extends SeleniumWebdriver {
   /**
    * {{> ../webapi/waitForElement }}
    */
-  waitForElement(locator, sec = null) {
-    sec = sec || this.options.waitForTimeout;
-    let el = element(guessLocator(locator) || by.css(locator));
+  waitForElement(locator, aSec = null) {
```
DavertMik: Why do you prefer `aSec` instead of `sec`?

DavertMik:
Thanks for the awesome work!
This PR should dramatically improve quality of our codebase.

lib/helper/Protractor.js
KennyRules: This showed up thanks to the TS compiler warnings for JS files ! :joy: 

If the parameter is null and waitForTimeout is undefined, null will get multiplied by 1000 which can cause super funky bugs.

DavertMik: Can you revert name of param back to `sec` because it is already documented as `sec`.
You can use `aSec` as an internal variable name )
Thanks!

DavertMik:
Thank you!

PR #709, opened 2017-09-12
Fix "[object Object] was not found by.." error message

BorisOsipov:
I always use objects in locators.
In many methods I got errors like "[object Object] was not found by.."
I tried to fix it.

### DIFF
--- a/lib/helper/WebDriverIO.js
+++ b/lib/helper/WebDriverIO.js
@@ -11,6 +11,7 @@ const fileExists = require('../utils').fileExists;
 const clearString = require('../utils').clearString;
 const decodeUrl = require('../utils').decodeUrl;
 const chunkArray = require('../utils').chunkArray;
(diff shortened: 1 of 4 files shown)


lib/helper/WebDriverIO.js
```diff
@@ -452,6 +452,14 @@ class WebDriverIO extends Helper {
     return findFields.call(this, locator).then((res) => res.value);
   }
 
+  static _throwNotFoundError(locator, prefixMessage = "Element",
```
DavertMik: can this be handled by new NotFoundError class instead of a custom method?
Something like

```js
throw new ElementNotFound(locator, msg);
```
It would be much cleaner this way

BorisOsipov:
Thank you for review.  I will change this.

DavertMik:
Thank you!

PR #712, opened 2017-09-13
Ignore local storage clean error, if we try to clean on data urls

APshenkin:
Ignore local storage clean error, if we try to clean localStorage on data urls


BorisOsipov:
@APshenkin thank you! I ran into this issue several times, but didn't have time to report\fix it.

PR #733, opened 2017-09-28
Fixed 2 failing tests from master.

KennyRules:
Root cause seems to be a "data" property for DataTable row objects needing to be accessed.


DavertMik:
Thanks!

PR #756, opened 2017-10-26
Allow providing arbitrary objects in config includes

rlewan:
Currently `config.includes` expects a map with `require` paths as values. This change allows providing object instances as well, e.g.:

```
exports.config = {
  includes: {
    loginPageObject: './load/me/from/here',
    homePageObject: new HomePageObject()
  }
}
```


DavertMik:
Yes, it was from implementation where we had JSON only config.
Thanks for implementation. Could you update the documentation to show this feature?

rlewan:
Sure, will do 👍

DavertMik:
Thanks!

PR #760, opened 2017-10-30
Add return default width and height into _startBrowser

sirreus:
Previously, there were failures when trying to start a scenario with the original window size, before in the previous scenario the window size was reduced.

For example:
`
Before((I, ownerOfferPage) => {
    I.amOnPage(ownerOfferPage.url);
    ownerOfferPage.specialOfferPageShouldBeOpen();
});

Scenario('some scenario headline',
    function* (I, ownerOfferPage) {
        I.resizeWindow(1290, 780);
        yield* ownerOfferPage.viewContentBlocksAfterResize();
    });

Scenario('some scenario headline',
    function* (I, ownerOfferPage) {
        yield* ownerOfferPage.checkThatBillboardIsComeOnPage(); <--- failed here*
        yield* ownerOfferPage.checkThatRightBannerIsPresent();
        yield* ownerOfferPage.checkThatSuperFooterComeOnPage();
    });
`

* because in Before specialOfferPageShouldBeOpen() check default windows size.

### DIFF
--- a/lib/helper/Appium.js
+++ b/lib/helper/Appium.js
@@ -178,6 +178,8 @@ class Appium extends WebdriverIO {
       }
       if (this.options.windowSize === 'maximize' && !this.platform) {
         promisesList.push(this.browser.execute('return [screen.width, screen.height]').then((res) => {
+          this.options.defaultWidth = res.value[0];
+          this.options.defaultHeight = res.value[1];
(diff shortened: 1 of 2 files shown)


DavertMik:
Sorry, I forgot about this PR and it appeared it is still valid by idea but the source files were refactored so it can't be accepted as is.

PR #747, opened 2017-10-13
Dockerized test environment

jploskonka:
I found it quite tricky to figure out how to run all codecept tests. I would really love to have it possible to be able to run all tests with one command. Also manually running php server for helpers tests was little bit irritating for me.

I use Docker a lot in development and don't want to install Selenium locally. @DavertMik I would like to get your opinion on that if it's something you'd consider merging and suggesting as simpler way to contribute.

If you like the idea I can cleanup the code and finish it sometime soon. There's still issue with Appium tests (I didn't have time to fit it yet).

## Todo:
- [x] easy way to run single helper tests
- [x] Cleanup code
- [x] green travisCI and Semaphore
- [x] rest test
- [x] move docker-compose to test dir
- [x] acceptance test
- [x] support node 6 & node 8
- [x] Update travis config to use dockerized tests
- [x] README info about using it for development :)

PR #771, opened 2017-11-09
Remove leftover wercker config files

jploskonka:
Removing `wercker.yml` config files, those are leftovers after experiments with #747 so nobody gets confused that we use WerckerCI ;)


DavertMik:
Yes, thanks

PR #777, opened 2017-11-15
fix nan sec

APshenkin:
Don't display (NaN sec) for skipped steps in tests


lib/output.js
```diff
@@ -94,11 +94,15 @@ module.exports = {
 
     if (outputLevel === 2) {
       newline = true;
-      process.stdout.write(`${styles.debug(" (" + time + ")\n")}`);
+      if (time !== 'NaN sec') {
```
DavertMik: this check should be for integer values and not for strings...
please move this few lines upper where the `time` value is calculated

DavertMik:
I think this is incorrect. Reproduced in Chrome console:

```
step = { startTime: 3, endTime: 5 }
{startTime: 3, endTime: 5}

 time = (step.endTime - step.startTime) / 1000 + " sec"
"0.002 sec"

isNaN(time)
true
```
isNaN check should happen before it was transformed to string

APshenkin:
@DavertMik 
You're right! My bad. Now it's fixed

DavertMik:
Ok, I will fix that manually

PR #783, opened 2017-11-17
Async await support

APshenkin:
Goals:

- [x] Support async/await in Feature hooks
- [x] Support async/await in Scenarios
- [x] Add tests for async/await
- [x] Update documentation

User benefits, when using async await:

if you want to use `grab` methods in pageObject, you have to create generator function in pageObject and then call this function with specific `yield*` in scenario

e.g.:

```
// pageObject:

goToTestPage: function* () {
  const url = yield I.grabTextFrom('mylocator');
  I.amOnPage(url);
}

// in Test:
Scenario('test', function* (myPageObject) {
  yield* myPageObject.goToTestPage()
});
```
because of this `yield`'s tests will look ugly. And also you have to remember is your function in pageObject a generator or not.

With async/await you should pass await only in case, when you want to get returned value.

e.g.:

```
// pageObject:

async goToTestPage() {
  const url = await I.grabTextFrom('mylocator');
  I.amOnPage(url);
}

// in Test:
Scenario('test', (myPageObject) => {
  myPageObject.goToTestPage()
});

// or example with return:
// pageObject:

async goToTestPage() {
  const url = await I.grabTextFrom('mylocator');
  I.amOnPage(url);
  return(url)
}

// in Test:
Scenario('test', async (myPageObject) => {
  const test = await myPageObject.goToTestPage();
  console.log(test)
});
```

### DIFF
--- a/lib/recorder.js
+++ b/lib/recorder.js
@@ -6,7 +6,7 @@ let errFn;
 let next;
 let queueId = 0;
 let sessionId = null;
-
+let asyncErr = null;
 let log = require('./output').log;
 let tasks = [];
 
@@ -58,6 +58,7 @@ module.exports = {
     if (promise && running) this.catch();
     queueId++;
     sessionId = null;
+    asyncErr = null;
     log(currentQueue() + `Starting recording promises`);
     promise = Promise.resolve();
     oldPromises = [];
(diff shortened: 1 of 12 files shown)


DavertMik:
Thanks! That's a really good enhancelemnt

lib/recorder.js
```diff
@@ -154,6 +155,16 @@ module.exports = {
     });
   },
 
+  saveFirstAsyncError(err) {
```
DavertMik: I have some doubts that this should be a part of recorder...

APshenkin: @DavertMik Can you please give some more information?

How do you think, where we should save first Async error?

DavertMik:
Let's go!

PR #789, opened 2017-11-21
Prevent REST helper from mutating default headers 

### LINKED ISSUE #784: REST Helper doesn't respect resetHeaders setting
I was trying to write some tests around endpoint authentication and I noticed that using `I.haveRequestHeaders` caused subsequent scenarios to include those headers when making get requests.

#### What do you get instead?
I see that later tests inherit headers from earlier tests even when `resetHeaders` is true. 

```bash
$ codeceptjs run --steps
CodeceptJS v1.0.3
Using test root "/home/alex/workspace/codecept-headers"

Repro test --
 with headers
 • I have request headers {"Authorization":"Bearer validToken"}
 • I send get request "/ping"
request headers  { Authorization: 'Bearer validToken', 'content-length': 0 }
 ✓ OK in 28ms

 without headers
 • I send get request "/ping"
request headers  { Authorization: 'Bearer validToken', 'content-length': 0 }
 ✖ FAILED in 4ms


-- FAILURES:

  1) Repro test: without headers:
     'Bearer validToken' == undefined
  
  Run with --verbose flag to see NodeJS stacktrace


  FAIL  | 1 passed, 1 failed   // 48ms
error Command failed with exit code 1.
```

```js

Feature('Repro test');

Scenario('with headers', function *(I) {
    I.haveRequestHeaders({
        'Authorization': 'Bearer validToken'
    });

    const {request: {headers}} = yield I.sendGetRequest('/ping');
    console.log("request headers ", headers);
    
    assert.equal(headers.Authorization, 'Bearer validToken');
});

Scenario('without headers', function *(I) {
    const {request: {headers}} = yield I.sendGetRequest('/ping');
    console.log("request headers ", headers);

    assert.equal(headers.Authorization, undefined);
});
```

### Details

* CodeceptJS version: 1.0.3
* NodeJS Version: v8.8.0
* Operating System: Ubuntu 16.04 Xenial
* Unirest: 0.5.1 

* Configuration file:

```js
{
  "tests": "./*_test.js",
  "timeout": 10000,
  "output": "./output",
  "helpers": {
    "REST": {
      "endpoint": "http://localhost:3030",
      "resetHeaders": true
    }
  },
  "include": {},
  "bootstrap": "./bootstrap.js",
  "teardown": "./bootstrap.js",
  "mocha": {},
  "name": "codecept-headers"
}
```
I also have a repo with a tiny Hapi server that reproduces the issue and includes my bootstrap code: https://github.com/alexashley/codecept-headers

Running the test without the authorization header first fixed the issue, but I ultimately decided to pass the headers along with every `sendGetRequest` instead of using `haveRequestHeaders`.

alexashley:
Closes #784. I linked this locally and ran it against my [reproduction repo](https://github.com/alexashley/codecept-headers) and everything looked good. The root cause was the assignment in the constructor causes the default headers to be mutated with every `haveRequestHeaders`.

The downside is that any tests that relied on this behavior are now broken (possibly including any `Before` hooks, but I haven't tested this).

### DIFF
--- a/lib/helper/REST.js
+++ b/lib/helper/REST.js
@@ -32,7 +32,7 @@ class REST extends Helper {
       endpoint: ''
     };
     this.options = Object.assign(this.options, config);
-    headers = this.options.defaultHeaders;
+    headers = Object.assign({}, this.options.defaultHeaders);
   }
 
   static _checkRequirements() {
@@ -68,7 +68,7 @@ class REST extends Helper {
 
   _cleanRequestHeaders() {
     if (this.options.resetHeaders) {
-      headers = this.options.defaultHeaders;
+      headers = Object.assign({}, this.options.defaultHeaders);
     }
   }
 
--- a/test/rest/REST_test.js
+++ b/test/rest/REST_test.js
@@ -1,16 +1,12 @@
 'use strict';
 const TestHelper = require('../support/TestHelper');
-let REST = require('../../lib/helper/REST');
-let should = require('chai').should();
-let api_url = TestHelper.jsonServerUrl();
-let assert = require('assert');
-let path = require('path');
-let fs = require('fs');
-let fileExists = require('../../lib/utils').fileExists;
-let AssertionFailedError = require('../../lib/assert/error');
-let expectError = require('../../lib/utils').test.expectError;
+const REST = require('../../lib/helper/REST');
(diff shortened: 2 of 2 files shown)


DavertMik:
Looks like tests are passing so the PR is ok to me!
Thanks!

Issue #790, opened 2017-11-21
Using haveRequestHeaders before sendPostRequest give us an empty header type


petrisorionel:
#### What are you trying to achieve?

I was trying to write some tests and I noticed that using I.haveRequestHeaders before I.sendPostRequest caused an issue because in sendPostRequest method the payload is sent first and the headers is sent secondly.

#### What do you get instead?

I get this error : 'Invalid JSON: [type] is missing.' 
HEADERS: [object Object]
TYPE: undefined
HEADERS: {}

> Provide test source code if related

Scenario('Send POST - ',  (I) => {
    let resp = yield I.sendPostRequest('/rest',
    {
    "data": {
        "relationships": {
            "promotions": {
                "data": {
                    "type": "promotions",
                    "id": "2"
                }
            }
        },
        "attributes": {
            "prd": "937",
            "available_runtimes": "",
            "extra_runtime": "",
            "available_users": "3",
            "users": "1",
            "extra_users": "",
            "discount": "17.0000",
            "extra_params": ""
        },
        "type": "promotion-it"
      }
    },
});

### Details

* CodeceptJS version: 1.0.3
* NodeJS Version: v8.8.0
* Operating System: MAC OS Sierra
* Unirest: 0.5.1

reubenmiller:
Fixed in PR #791

PR #791, opened 2017-11-21
Fixing the issue #790

petrisorionel:
Using haveRequestHeaders before sendPostRequest give us an empty header type.


DavertMik:
Thanks!

PR #793, opened 2017-11-23
Circle ci dockerized tests

jploskonka:


jploskonka:
I opened it by mistake too early, wanted to do it to my fork so that's why I closed it.

PR #796, opened 2017-11-27
Circle ci: dockerized tests

jploskonka:
# Why?
#747 introduced dockerized test environment for easier usage in development. However TravisCI still use non-dockerized way of running it. Also Docker is fun and I really want to have Appium tests working in emulator :D 

# Differences with TravisCI
- builds against 3 node versions instead of 2 (6.9.5, 8.9.1 and 9.2.0)
- Running acceptance tests (we skip them on travis)
- Because of building docker containers and my lack of experience with optimizing it build takes a bit longer than TravisCI (~14min vs ~10min). However for one more node version and additional acceptance specs I'd say it's not bad for first version :)

# TODO
- [x] - move shell scripts to well... shell scripts instead of keeping them in yml :P 
- [x] - simplify `config.yml`
- [x] - Fix 2 examples failing on CircleCI

@DavertMik I have two WebDriverIO exmaples failing on CircleCI unfortunately, I wasn't able to reproduce it locally and it also used to work OK there - I have no idea yet what is going on here :D   If you have any suggestion what may be wrong please let me know, maybe you saw something like this before :)

### DIFF
--- a/.circleci/build.sh
+++ b/.circleci/build.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -e
+
+cd test
+
+docker-compose build \
+  --build-arg NODE_VERSION=$NODE_VERSION \
+  test-unit \
+  test-helpers \
+  test-acceptance.webdriverio \
+  test-acceptance.nightmare \
+  json_server
+
+docker-compose pull \
+  php \
+  selenium.chrome
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,35 @@
+version: 2
+
+defaults: &defaults
+  machine:
+    image: circleci/classic:201710-02
+    docker_layer_caching: true
+  steps:
+    - checkout
+    - run: .circleci/build.sh
+    - run: .circleci/test.sh
+
+jobs:
+  node 6.9.5:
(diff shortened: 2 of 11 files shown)


jploskonka:
I will have one more shot at those failing scenarios before merging this week - it's for sure some small mistake ;) Also  I cleaned up config, removed my "optimizations" (I way overdid it with circle workflows and custom caching :D) and went down to little bit over 10min build :)

As for Appium I haven't look into that yet but for sure will follow one day, I have pretty limited time for codecept unfortunately right now.

DavertMik:
> I have pretty limited time for codecept unfortunately right now.

Wow. your limited time looks like greater than mine ) 
Ok, will be waiting for final update then!
Btw, looks like we are going to remove Node 6 support so it can be dropped from versions

jploskonka:
@DavertMik looks ready to merge :D 

I'm not sure why travis build wasn't triggered for this one, it passed for my fork: https://travis-ci.org/jploskonka/CodeceptJS/builds/309036420 

Also I had problems with `expectError` utility, for some unknown to me reason examples with it started failing because expectations were applied against error thrown by it instead of expected. Always good to remove some code ;-)

PR #798, opened 2017-12-01
fix displaying undefined args in output

APshenkin:
undefined args in test break it.

example:

Scenario.only('test', (I) => {
  let b;
  I.assertEqual(b, 'undefined');
});

expected:
```
@Test --
 test
 • I assert equal undefined, "undefined"
 ✖ FAILED in 7ms


-- FAILURES:

  1) @Test: test:
     undefined == 'undefined'
  
  Scenario Steps:
  
  - I.assertEqual(undefined, "undefined") at Test.Scenario (tests/test.js:12:5)
```

actual:
```
@Test --
 test
 ✖ FAILED in 1ms


-- FAILURES:

  1) @Test: test:
     Cannot read property 'toString' of undefined
```

This PR will fix the issue

### DIFF
--- a/lib/step.js
+++ b/lib/step.js
@@ -57,6 +57,8 @@ class Step {
         return arg.toString();
       } else if (typeof arg === "object") {
         return JSON.stringify(arg);
+      } else if (typeof arg === "undefined") {
+        return `${arg}`;
       } else if (arg.toString) {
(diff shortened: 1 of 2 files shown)


DavertMik:
Thanks!

PR #800, opened 2017-12-02
[WIP] Refactoring

DavertMik:
* Update to NodeJS v8
* Refactor helpers to use async/await
* Merge Protractor & SeleniumWebDriver helpers


.eslintrc.json
```diff
@@ -34,7 +34,8 @@
     "default-case": 0,
     "import/no-extraneous-dependencies": 0,
     "no-cond-assign": 0,
-    "array-callback-return": 0,
-    "function-paren-newline": 0
+    "import/no-unresolved": 0,
```
DavertMik: Hound CI can't parse this rule. Also, it depends on current installation settings so strictly following it won't be helpful.

.eslintrc.json
```diff
@@ -34,7 +34,8 @@
...
+    "no-await-in-loop": 0,
```
DavertMik: we have loops where await is used following `if` conditions. There is no clean way to rework it in `Promise.all`

.eslintrc.json
```diff
@@ -34,7 +34,8 @@
...
+    "arrow-body-style": 0
```
DavertMik: this makes some linex extra long and unreadable

PR #810, opened 2017-12-04
Fix no-plusplus rule

galkin:


DavertMik:
I'd rather ignore this rule. It doesn't makes much sense, it affects readability more then it helps (never seen ++ issues anywhere)

PR #817, opened 2017-12-07
Add browser name to report file for multiple run

trollr:
The name of the report file for multiple runs is always report.xml. CI/CD Environments like Jenkins need unique names to recording test results


trollr:
How to fix this hound issue? -.-

DavertMik:
Just use ESLint :)

But no problems, I can fix it myself. Thanks

PR #820, opened 2017-12-08
[WIP] add merge feature

APshenkin:


DavertMik:
1. this is not related to Data context so the place where this function is added is incorrect
2. dependency injection doesn't happen automatically as in other cases so I think this is confusing

this feature can be used outside of a core, so you can just create `merge` function and use it in your projects. You don't need it in the core.

DavertMik:
I think this would look better if I could export scenarios as functions and then import them like this:

```js
require('./common').forEach((t) => t.injectDependencies());
// if you need to override dependnecies, you can pass them in
require('./common').forEach((t) => t.injectDependencies({ myData }));
```
I think this syntax much more verbose and less magical. 
Also it doesn't introduce a new concept.

Issue #474, opened 2017-04-13
Coffeescript support


laurentpayot:
Mocha allows me to use CoffeeScript (v2) files for my unit tests with:
```
mocha tests/unit/*_tests.coffee --compilers coffee:coffeescript/register
```
Would it be possible to have a similar option in codecept ?

PS: Thank you for your great framework that deserves more traction.

DavertMik:
There is the same `compilers` option for `codeceptjs run`. Did you try it?

laurentpayot:
I tried but it doesn't even recognize CoffeeScript function arrows:
```
$ codeceptjs run --compilers coffee:coffeescript/register
 
Unexpected token >
 
/home/laurent/myproject/tests/e2e/layout_tests.coffee:3
Before((I) ->
            ^
SyntaxError: 
    at createScript (vm.js:53:10)
    at Object.runInThisContext (vm.js:95:10)
    at Module._compile (module.js:543:28)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at /home/laurent/.config/yarn/global/node_modules/mocha/lib/mocha.js:222:27
```
My CS code is ok because I can run my CS tests if I compile them in a directory (.js) like so:
```
coffee -b -o ./tests/e2e/.js -c ./tests/e2e/*.coffee && codeceptjs run ./tests/e2e/.js --config ../codecept.json
```
(For page objects I also had to declare `I` before `module.exports`, something you hardly do in CS, but that's fine.)

That compilation directory workaround is quite dirty…

DavertMik:
Yes, I understand. Can you look around into source code to find how this can be solved? 
Anyway, CodeceptJS uses Mocha so the implementation for compilers should be pretty easy to add. 
I don't have much time to work on this feature but as you I'd really like to see CoffeeScript support.

laurentpayot:
Sure, I will try to look around asap.

laurentpayot:
I had a look but it's not obvious… Something to do with commander.js ?

laurentpayot:
Actually it seems that all mocha command line options other than `--reporter` are not taken into account. For instance `-C` (forcing no colors) has no effect.

laurentpayot:
[...]

Issue #370, opened 2017-01-14
[ Question ] How to get babel working to support imports?


jhartma:
#### What are you trying to achieve?
I want to use es2015 import syntax

#### What do you get instead?
I tried the ```--compilers js:babel-core/register``` option which works fine with mocha, but I get a console error ```Unexpected token import```

```bash
SyntaxError:                                                                                             
    at Object.exports.runInThisContext (vm.js:78:16)                                                     
    at Module._compile (module.js:543:28)                                                                
    at Object.Module._extensions..js (module.js:580:10)                                                  
    at Module.load (module.js:488:32)                                                                    
    at tryModuleLoad (module.js:447:12)                                                                  
    at Function.Module._load (module.js:439:3)                                                           
    at Module.require (module.js:498:17)                                                                 
    at require (internal/module.js:20:19)                                                                
    at .../node_modules/codeceptjs/node_modules/mocha/lib/mocha.js:222:27
```

#### Details
* CodeceptJS version: 0.4.13
* NodeJS Version: 7.4.0
* Operating System: Linux (Arch)

DavertMik:
Sorry, babel is not supported as well as any other compilers. I hope at some point NodeJS will implement imports natively.

Probably some support of transpilers can be implemented but I don't think I will do it in nearest future. Working with native code has its benefits.

sfedorin:
Hi,
The same question for me... I tried to set packages
```
{
  "name": "portal",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "npx codeceptjs run --steps"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "dependencies": {
    "@babel/cli": "^7.8.3",
    "@babel/core": "^7.8.3",
    "@babel/plugin-proposal-class-properties": "^7.8.3",
    "@babel/plugin-proposal-export-default-from": "^7.8.3",
    "@babel/polyfill": "^7.8.3",
    "@babel/preset-env": "^7.8.3",
    "@babel/preset-react": "^7.8.3",
    "@babel/register": "^7.8.3",
    "@codeceptjs/mailslurp-helper": "^1.0.2",
    "codeceptjs": "^2.3.6",
    "puppeteer": "^2.0.0",
    "mailslurp-client": "^6.7.3"
  }
}
```
.babelrc file has beed added as well. But I still get error `SyntaxError: Cannot use import statement outside a module`

evasyuk:
[...]

PR #834, opened 2017-12-19
[WIP] experimental implementation for enzyme and react

APshenkin:
What is a goal:
We want to have component tests for React in the same maner as selenium tests.

What's is included in this PR:
1. Support for mocha features `compilers` and `require`. This should solve these issues: https://github.com/Codeception/CodeceptJS/issues/474 https://github.com/Codeception/CodeceptJS/issues/370
2. Add option `--transpile`. When it's true, then transpiled code will process without issues. And also React components will displays correctly in test stdout.
3. Add Enzyme helper. Currently only mount methods. Have to add some manipulation methods

How to run it:
1. add these deps to package.json:
enzyme
enzyme-adapter-react-16
jsdom
react-element-to-string
sinon
babel-polyfill
babel-core
esprima
escodegen

and install them.

2. create `.babelrc`  file to store babel settings. If you are using webpack aliases in your project you also have to install  `babel-plugin-webpack-aliases`:
e.g.
```
{
  "presets": [ "react", "es2015", "stage-1" ],
  "plugins": [
    [ "babel-plugin-webpack-aliases", { "config": "./webpack.config.js" } ]
  ]
}
```

3. Write a test. E.g.:
```
import React from 'react';
import Button from '_shared/components/button';

Feature('test');

Scenario('simple test', async (I) => {
  const component = await I.mountComponent(<Button/>);
  console.log(component.html());
});
```

4. create run script in package.json. E.g.: `codeceptjs run --transpile --compilers js:babel-core/register,js:babel-polyfill`

5. In my case there are some css styles imports in project. They can't be traspiled by babel automaticly. In our project it's webpack job, so we have to emulate this imports. To do this I create a special js file:
```
// ./tests/helper.js
function donothing() {
  return null;
}

require.extensions['.css'] = donothing;
require.extensions['.less'] = donothing;
require.extensions['.scss'] = donothing;
require.extensions['.styl'] = donothing;

```
and add it to codeceptjs run script:
`codeceptjs run -R mocha-multi --transpile --compilers js:babel-core/register,js:babel-polyfill,css:./tests/helper`

So next steps for this:
- [ ] Test it with real project
- [ ] Add more Enzyme and sinon methods
- [ ] Write tests for these feature
- [ ] Create a boilerplate project for enzyme


DavertMik:
I think this PR should be split into two:

* transpilers/babel support
* enzyme

PR #843, opened 2017-12-26
add ability to require helpers from node_modules

APshenkin:
We use private npm registry to share custom helpers between several projects.

These changes will give ability to require custom helpers from node_modules


DavertMik:
yes, this is a MUST. Please fix the issues listed here and I'm ready to merge.

DavertMik:
Thanks!

PR #844, opened 2017-12-27
Better arrays humanize

APshenkin:
Current output for array of objects:
[[object Object],[object Object]]

This change will display array of objects correctly, e.g.:
[{"myfar": "foo"},{"codeceptjs": "is the best"}]


DavertMik:
Thanks!

PR #847, opened 2017-12-27
Support JSON objects as payload within REST

Kalostrinho:
#### REST helper enhancements:
- [X] Added **_resetRequestHeaders_** function to set headers back to default at any given time
  - Use case: When several requests with different headers are sent within the same scenario, the **_resetHeaders: true_** within the config is not useful
- [X] Added support for JSON objects since payload (as a JSON) was automatically converted into "URL query" type of parameters:
`// if payload object was ---> payload = { "one": "The default", "two": false};`
`// when this was sent to a POST, the request object was converted it into:`
`// <uri>?one=The%20default&two=false`

### DIFF
--- a/lib/helper/REST.js
+++ b/lib/helper/REST.js
@@ -41,6 +41,14 @@ class REST extends Helper {
     }
   }
 
+  /**
+   * Set timeout for the request
+   *
+   * ```js
+   * I.setRequestTimeout(10000); // In milliseconds
+   * ```
+   *
+   */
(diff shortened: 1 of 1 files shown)


Kalostrinho:
**_Failing on TravisCI due to a scenario on ApiDataFactory_test_** @DavertMik

DavertMik:
Yes, those tests are unstable. Let me restart

DavertMik:
Thanks!

PR #848, opened 2017-12-28
docs(FileSystemHelper): add a reminder for users to activate the helper

javigomez:
This pull adds a reminder in the documentation to avoid people having the trouble of FileSystem Helper functions not being available just because they forgot to add the Helper in the configuration file. For instance:`I.writeToFile is not a function`

It happened to me so I guess it will happen to others. 

Signed-off-by: javigomez <javier@typeform.com>


DavertMik:
Good idea. But the same as with Codeception this part is auto-generated, so this notice should be added to the docblock of a helper itself.

javigomez:
Ok Has sense

Issue #853, opened 2018-01-02
Missing uncheckOption()


brunobg:
#### What are you trying to achieve?
Ensure that a checkbox input is not filled

#### What do you get instead?
Nothing, there is checkOption but not uncheckOption.

### Details
WebdriverIO, any version. Am I missing something? I can provide a PR with uncheckBox. Right now the only way to do it seems to be with a executeScript.

DavertMik:
>  I can provide a PR with uncheckBox. Right now the only way to do it seems to be with a executeScript.

Yes, please send this PR
But I think the same effect can be achieved with `click`

brunobg:
Clicking would not ensure the final state. If the checkbox is originally checked, click only toggles it. Codeception has uncheckOption http://codeception.com/docs/modules/WebDriver#uncheckOption.

I'll send the PR later for WebdriverIO, but I think it will need further work for the other backends.

PR #856, opened 2018-01-04
Implements uncheckOption in WebDriverIO

brunobg:
PR for this issue: https://github.com/Codeception/CodeceptJS/issues/853

Code is extremely similar to checkOption(). I included docs but not tests.

### DIFF
--- a/lib/helper/WebDriverIO.js
+++ b/lib/helper/WebDriverIO.js
@@ -650,7 +650,7 @@ class WebDriverIO extends Helper {
   }
 
   /**
-   * {{> ../webapi/checkOption }}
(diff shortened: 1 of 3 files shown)


lib/helper/WebDriverIO.js
```diff
@@ -649,6 +649,24 @@ class WebDriverIO extends Helper {
     return this.browser[clickMethod](elem.ELEMENT);
   }
 
+  /**
+   * {{> ../webapi/checkOption }}
```
DavertMik: `uncheckOption` should be here

brunobg:
Sorry, doc is now fixed.

DavertMik:
Thanks!

Issue #858, opened 2018-01-06
run-multiple creates to long subfolders


tfiwm:
The sub directories which are created are to long because of line 109 in run-multiple.js:

```javascript
const outputDir = suite + JSON.stringify(browserConfig).replace(/[^\d\w]+/g, '_') + suiteId;
```

The problem here is that the the browserConfig gets stringified but in our case we are using browserstack and are setting a lot of desiredCapabilities which ends up in a really long JSON string and to an error (`Error: ENAMETOOLONG: name too long, mkdir`).

I can understand in the example that i want to see the browser + resolution. 
We need to find a way to give it a custom name as optional. Maybe as function property where the function returns the output name:

Here an example how it could look like:

```javascript
multiple: {
    login: {
      grep: '#auth #smokeTest - As an owner I Login & Logout',
      browsers: [
        {
          browser: 'Safari',
          desiredCapabilities: {
             'os': 'OS X',
             'os_version': 'High Sierra',
             'browser': 'Safari',
             'browser_version': '11.0',
             'resolution': '1600x1200'
          },
          outputName: function() { // dynamic version
             return 'MY_CUSTOM_NAME'
          }
          outputName: 'MY_CUSTOM_NAME' // string version
        }
      ]
    }
  }
```

Why should outputDir a function? Because it maid be possible that the browser list gets generated and a function will be flexible to return the right outputDir based on own algorithm.

As fallback a string should be possible too.

How do you like this proposal? Do you have better ideas? Or is there another way to solve it?

tfiwm:
@DavertMik Please can you take a look? THX!!

DavertMik:
Thanks for PR. Will be released in 1.1.2

PR #859, opened 2018-01-06
Custom output name for multiple browser run

tfiwm:
Added possibility to use custom output name for directory

#858


DavertMik:
Thanks!

PR #861, opened 2018-01-08
Compiler fixes

KennyRules:
Fixes
- Removed gulp execution from docker-compose for unit tests since it is no longer part of package.json.
- Fixed a bug where a waiter would report waiting for "null" seconds if waiting for an amount of time specified in helper options and thus not passing a time to waitForText invocation.
- Updated a failing test to pass by adding in quotation marks for object print.

Improvements
- Updated missing JSdoc notation around parameter types.
- Touched up some Markdown to be in line with existing documentation style.
- Put _beforeStep and _afterStep back in with associated acceptance test.
- Put async/await in functions missing it in puppeteer helper.
- Added build npm script to run TypeScript compiler.


DavertMik:
Thank you very much for fixes and improvements!

Issue #889, opened 2018-01-22
TypeError: msg.args.join is not a function


frantallukas10:
https://github.com/Codeception/CodeceptJS/blob/9cd3e50a82bf7f3447e3a250656ff5c0aa1b47da/lib/helper/Puppeteer.js#L217

"puppeteer": "^1.0.0"
"codeceptjs": "^1.1.2",

(node:11528) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): TypeError: msg.args.join is not
a function
(node:11528) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:11528) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): TypeError: msg.args.join is not
a function
(node:11528) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 3): TypeError: msg.args.join is not
a function

...

PR #891, opened 2018-01-22
Fix for #889 - Checking if console.log was called with array or not

KennyRules:


DavertMik:
What if an object was passed in? 
Maybe just convert it to JSON?

KennyRules:
Hm, I'm not seeing this comment on GitHub, so gonna respond via email!

For this situation, if a JSON object is passed, it would need to be handled
further down the chain.

According to the Puppeteer docs, args can only be an array of JSHandles for
the console callback:
https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#class-consolemessage

Unless there is something I missed?

KennyRules:
@DavertMik : After some trial-and-error, I found a potential way to resolve the Semaphore error!

If you think it's a good idea, I can open another PR to replace npm with yarn to keep this PR focused on the bugfix.

Alternatively, or perhaps additionally, there might be some Semaphore project settings that could be tweaked? I'm not familiar with the tool, though it seems that it can only be changed on the project page.

DavertMik:
> Hm, I'm not seeing this comment on GitHub, so gonna respond via email!

Yeah, I understood the way it works and deleted my comment. Sorry )

>  I can open another PR to replace npm with yarn to keep this PR focused on the bugfix.

I think both yarn lock and package lock should be removed from the repo. Anyway, they won't be taken into account when installed as a library and better to keep them out

PR #897, opened 2018-01-29
Integrate with nightmare-har-plugin

### LINKED ISSUE #413: Capture the network activity of an http request in HAR format using CodeceptJS
I'm working on getting the HAR report for the webpage I'm testing. This works with the help of "nightmare-har-plugin" using NightmareJS. I'm trying to use codeceptJS and looks like codeceptJS doesn't have the option to get the HAR. Is it something which the team will add in the near future?

Below is my code.
Scenario("Functional Test HomePage", function () {

	 har.install(Nightmare);
	 var nightmare = Nightmare(Object.assign(har.getDevtoolsOptions(), options));
	 nightmare.waitForDevtools()
		 .goto("http://www.google.com")
		 .getHAR()
		 .end()
		 .then(function (result) {
			 console.log(result)
		 })
		 .catch(function (error) {
		    console.error(error);
	    });
 });

mingfang:
fix #413

### DIFF
--- a/lib/helper/Nightmare.js
+++ b/lib/helper/Nightmare.js
@@ -227,7 +227,7 @@ class Nightmare extends Helper {
       this.browser = this.Nightmare(Object.assign(require('nightmare-har-plugin').getDevtoolsOptions(), this.options));
       await this.browser;
       await this.browser.waitForDevtools();
-    }else{
+    } else {
       this.browser = this.Nightmare(this.options);
       await this.browser;
     }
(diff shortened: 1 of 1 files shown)


lib/helper/Nightmare.js
```diff
@@ -232,6 +245,22 @@ class Nightmare extends Helper {
...
 
+  /**
+   * Get HAR
+   *
+   * ```js
+   * let har = yield I.grabHAR();
+   * fs.writeFileSync('sample.har', JSON.stringify({log: har}));
+   * ```
+   */
+  async grabHAR() {
```
DavertMik: can you make a function to saveHAR as well? 
Looks like that would be useful too

mingfang: added saveHar()

DavertMik:
Thanks!

PR #912, opened 2018-02-05
Remove unnecessary indentation breaking formatting

pavelloz:
Correct: https://codecept.io/helpers/#webdriverio-example
Wrong: https://codecept.io/helpers/#hook-usage-examples

Indentation is the only difference i noticed between those blocks of text.


DavertMik:
Thanks!

PR #915, opened 2018-02-06
Puppeteer helper, proceedClick should return _waitForAction

othree:
Since [click](https://github.com/Codeception/CodeceptJS/blob/master/lib/helper/Puppeteer.js#L625-L641) return the value from `proceedClick`.
I think `proceedClick` should return waitForAction just like other actions.


DavertMik:
In this case, this should not be a difference. 
The idea is the same: wait for action after the action is performed,

PR #917, opened 2018-02-07
requests to API fix

Spartans2017:
The latest changes from 1.1.2 are not working anymore for sendPostRequest .. if you are trying to send an object it's trying to convert it to string but the request itself is not successful. 

So with this change you can pass JSON object and also form urlencoded payload types.

example:
let payload = { username: "user1", password: "xxx" };
let payload2 = 'username=user1&password=xxx';

extra unit tests added + fixed old ones

### DIFF
--- a/lib/helper/REST.js
+++ b/lib/helper/REST.js
@@ -170,7 +170,6 @@ class REST extends Helper {
    * @param {object} headers
    */
   sendPostRequest(url, payload = {}, headers = {}) {
-    payload = typeof payload !== 'object' ? JSON.parse(payload) : payload;
     request = unirest.post(this._url(url)).headers(headers).send(payload);
     return this._executeRequest(request);
   }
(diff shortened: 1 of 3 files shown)


Kalostrinho:
I managed to replicate the scenario you're saying. Your fix is cool, although I'm not sure `why` would you send an object as a string in the first place. Probably it'll work for you with the current version, if you send `I.sendPostRequest(<endpoint>, JSON.parse(yourStringObject))`

**If this PR gets approved, I would recommend you also include the fixes for both _`PATCH`_ and _`PUT`_ accordingly, since they have the same validation for payload** 
👍

Spartans2017:
Thanks @Kalostrinho for your input, I fixed those methods, basically we don't need that line where we serialising the JSON. all requests can accept only JSON or form urlencoded payload types. Therefore I removed that line in all methods and fixed and introduced new unit tests for POST, PUT, PATCH methods.

DavertMik:
Looks good to me. I'm about to merge it. @Kalostrinho any objections?

PR #921, opened 2018-02-09
Puppeteer fixes

abidhahmed:
Fixed a bunch of issues with the Puppeteer helper.

- Exit gracefully after a test completes
- Add ability to set `user-agent` via config
- Add `keepCookies`,`keepBrowserState` and `restart` (feature parity between helpers)

Updated the documentation, but haven't added any tests for these. Wanted to fix the double-tab issue during startup, but see there's another PR already open for that from @reubenmiller.


DavertMik:
Thanks, looks good to me!

PR #922, opened 2018-02-09
add injectDependencies method

APshenkin:
As mentioned in https://github.com/Codeception/CodeceptJS/pull/820#issuecomment-352279338
Now test object has injectDependencies method that allow to inject custom deps to test:

usage:
```
require('./common').forEach((t) => t.injectDependencies());
```

### DIFF
--- a/lib/interfaces/bdd.js
+++ b/lib/interfaces/bdd.js
@@ -60,6 +60,11 @@ module.exports = function (suite) {
       }
       const test = new Test(title, fn);
       test.fullTitle = () => `${suite.title}: ${title}`;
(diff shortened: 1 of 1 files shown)


DavertMik:
Yeah, it's ok. Could you add documentation on how this can be used for reusing test scenarios? 

Also this can be used inside a single test case where you can inject a standalone module:

```js
this.injectDependencies({ myData: './data/users.js' });
// or dynamically load steps
this.injectDependencies({ mySteps: './data/steps.js' });
```

DavertMik:
Please run `npm run lint-fix` to fix the builds!

PR #933, opened 2018-02-14
Fix unhandled rejection

APshenkin:
Case when async function used without await and then we have other await function


DavertMik:
Ok, looks good!

PR #934, opened 2018-02-14
Fix waitUntilExists for WebdriverIO

abidhahmed:
Currently, `waitUntilExists` is using `waitForStalenessOf` in the background - but those two are completely opposite.

`waitForStalenessOf` waits for an element to **NOT** exist in a page, whereas `waitUntilExists` waits for an element to exist.

I've since updated the method to use [`[browser.waitForExist]`](http://webdriver.io/api/utility/waitForExist.html) directly from WebdriverIO to make this switch seamless and updated the documentation while I'm at it.


DavertMik:
yeah, makes more sense. Thanks!

DavertMik:
I'm reverting this PR and deprecating this method so no longer confusion happens.
Please use 

    * use 'waitForElement' to wait for element to be attached
    * use 'waitForDetached to wait for element to be removed

abidhahmed:
That works too. Thanks!

Issue #932, opened 2018-02-14
fillField append new value to textarea and clearField doesn't clear the textarea


suzuki11109:
#### What are you trying to achieve?
I want to set a value to textarea.

#### What do you get instead?
I use Vue. I use puppeteer.
I used fillField function to set value to textarea but it appears to append the value to the existing value instead of setting new one. 
So I tried to use clearField before fillField but the value still exist in textarea. 
I've tried to use other selector for textarea but the result is the same. 
And there is no any error so I'm sure the selector is correct.

> Provide console output if related. Use `--verbose` mode for more details.

```bash
# paste output here
```

> Provide test source code if related

```js
// paste test
```

### Details

* CodeceptJS version: 1.1.4
* NodeJS Version: 8.9.4
* Operating System: MacOS Sierra
* Protractor || WebDriverIO || Nightmare version (if related) : Puppeteer version 1.0.0
* Configuration file:

```js
# paste suite config here
{
  "tests": "./e2e/**/*_test.js",
  "timeout": 10000,
  "output": "./output",
  "helpers": {
    "Puppeteer": {
      "url": "http://localhost:4000",
      "show": false,
      "waitForTimeout": 5000
    }
  },
  "include": {
    "I": "./e2e/steps_file.js"
  },
  "bootstrap": false,
  "mocha": {},
  "name": "seminar.saito"
}

```

PR #935, opened 2018-02-14
[Bugfix #932] Clear value attribute instead of innerhtml for TEXTAREA elements

reubenmiller:
### Highlights
* Fixing #932. The TEXTAREA item wasn't being cleared because the innerhtml attribute was being cleared instead of the value attribute.
* Lint fix

### References
* [text area value vs innerhtml](https://stackoverflow.com/questions/5314186/javascript-get-textarea-input-via-value-or-innerhtml)


DavertMik:
Thanks!

PR #936, opened 2018-02-14
Retries API and fix for Puppeteer

DavertMik:
* [Puppeteer] Rerun steps failed due to "Cannot find context with specified id" Error.
* Added syntax to retry a single step:

```js
// retry action once on failure
I.retry().see('Hello');
// retry action 3 times on failure
I.retry(3).see('Hello');
// retry action 3 times waiting for 35 seconds
I.retry({ retries: 3, maxTimeout: 35000 }).see('Hello');
```


reubenmiller:
I quickly tested it out using the following.

```js
Feature('Retry step');

Scenario('Retry single step @WebDriverIO @Protractor @Nightmare @Puppeteer', (I) => {
  I.amOnPage('https://github.com/');
  I.click('About');
  I.retry(3).see('is how people build software');
});

Scenario('Retry single step (async) @WebDriverIO @Protractor @Nightmare @Puppeteer', async (I) => {
  I.amOnPage('https://github.com/');
  I.click('About');
  I.retry(3).see('is how people build software');
});
```

The first thing that came to mind is, can I add a delay before the next retry is attempted (or at least a comment in the docs).

So maybe another option like "delay" would be helpful for the user writing the test.
```js
I.retry({ retries: 3, delay: 500 }).see('is how people build software');
```

test/unit/actor_test.js
```diff
@@ -71,4 +87,5 @@ describe('Actor', () => {
         listeners.should.eql(3);
       });
   });
+
```
reubenmiller: Need to remove the blank line as it is causing it causes the lint test to fail.

DavertMik:
> So maybe another option like "delay" would be helpful for the user writing the test.

yes, there is an option `minTimeout` which is exactly what you want. I will add this to documentation

reubenmiller:
Ok. Sounds good

Issue #898, opened 2018-01-30
Failed assertion with chai-as-promised is not working


mingfang:
#### What are you trying to achieve?
I'm trying to use chai-as-promised for assertions.
It works when the assertion passes but it does not work when the assertion fails.

#### What do you get instead?
I get this error
```
(node:40795) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): AssertionError: expected { Object (_readableState, readable, ...) } to have property 'foo'
(node:40795) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```
and the test is marked as passed when though it should fail.

> Provide console output if related. Use `--verbose` mode for more details.

```bash
./node_modules/codeceptjs/bin/codecept.js run  --verbose
CodeceptJS v1.1.2

   [1] Starting recording promises
   Emitted | suite.before ([object Object])
   [1] Queued | hook Nightmare._beforeSuite()
   [1] Queued | hook Mochawesome._beforeSuite()
   [1] Queued | hook REST._beforeSuite()
   Emitted | test.before
   [1] Queued | hook Nightmare._before()
   [1] Queued | hook Mochawesome._before()
   [1] Queued | hook REST._before()
   Emitted | test.start ([object Object])
   Emitted | step.before (I send get request "http://google.com")
   [1] Queued | hook Nightmare._beforeStep()
   [1] Queued | hook Mochawesome._beforeStep()
   [1] Queued | hook REST._beforeStep()
   [1] Queued | sendGetRequest: "http://google.com"
   Emitted | step.after (I send get request "http://google.com")
   [1] Queued | hook Nightmare._afterStep()
   [1] Queued | hook Mochawesome._afterStep()
   [1] Queued | hook REST._afterStep()
   [1] Queued | step passed
   [1] Queued | return result
   [1] Queued | fire test.passed
   [1] Queued | finish test
   Emitted | step.start (I send get request "http://google.com")
 • I send get request "http://google.com"
   Emitted | step.passed (I send get request "http://google.com")
   Step finished in 0.06 sec
   Emitted | test.passed ([object Object])
 ✓ OK in 103ms

(node:40934) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): AssertionError: expected { Object (_readableState, readable, ...) } to have property 'foo'
[...]

Issue #786, opened 2017-11-19
Assert look likes doesn't work


anhvuong123:
I've defined a basic case to start working with Assert in CodeceptJS:
```
var assert = require('assert');
assert.equal('true', 'false');
```
Assert looks like doesn't work. The test consists code above was passed without alert relating to Fail Assert. Please help me via the case.

Thanks,

alexashley:
@anhvuong123 To my knowledge, Codecept doesn't ship with an assertion library, the examples use Node's [assert](https://nodejs.org/api/assert.html).

Also your example works as expected for me in the Node REPL:
```bash
λ> node
> const assert = require('assert')
undefined
> assert.equal('true', 'false')
AssertionError [ERR_ASSERTION]: 'true' == 'false'
    at repl:1:8
    at ContextifyScript.Script.runInThisContext (vm.js:50:33)
    at REPLServer.defaultEval (repl.js:240:29)
    at bound (domain.js:301:14)
    at REPLServer.runBound [as eval] (domain.js:314:12)
    at REPLServer.onLine (repl.js:441:10)
    at emitOne (events.js:120:20)
    at REPLServer.emit (events.js:210:7)
    at REPLServer.Interface._onLine (readline.js:282:10)
    at REPLServer.Interface._line (readline.js:631:8)
> assert.equal('true', 'true')
undefined
> 
```
Are you sure the tests are running? Could you provide more sample code?

mingfang:
Assertions hangs for 30s when the assertion fails.
This example
```
Scenario('test', async(I) => {
  var assert = require('assert');
  //assert.equal('foo', 'foo'); //this works
  assert.equal('foo', 'bar'); //this does not work
});
```
fails with this error
```
Error: Evaluation timed out after 30000msec.  Are you calling done() or resolving your promises?
```

reubenmiller:
The fix for this landed in v1.1.5.

PR #938, opened 2018-02-15
[Nightmare|Bugfix] Init browser with about:blank so browser does not hang if scenario fails before I.amOnPage() is called

reubenmiller:
#### Highlights

* Initialise Nightmare (in _startBrowser) on a blank page
* Fixes following issues, #786 & #898 

#### Root Cause

If a scenario fails before `I.amOnPage` is called, then the subsequent call to the _failed hook in the Nightmare helper, tries to take a screenshot of the browser. Before a screenshot is taken the height and width of the browser is retrieved using `browser.evaulate`, however if Nightmare has not had `.goto` called on it, then `.evaulate` will timeout.

### DIFF
--- a/lib/helper/Nightmare.js
+++ b/lib/helper/Nightmare.js
@@ -231,6 +231,7 @@ class Nightmare extends Helper {
       this.browser = this.Nightmare(this.options);
       await this.browser;
     }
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks!

Issue #732, opened 2017-09-28
[documentation] Wrong description for action 'grabSource' at WebDriverIO web-page.


amokmen:
There is wrong description for action [grabSource](http://codecept.io/helpers/WebDriverIO/#grabsource) at WebDriverIO web-page.

It seems that description was just copied from another action - **seeInSource**.

Right description should be like this:

> Retrieves a page source and returns it to test. Resumes test execution, so should be used inside a generator with yield operator.
> `let source = yield I.grabSource();`

Issue #658, opened 2017-08-09
Typo in within docs


APshenkin:
```
within('.form-signup-home', () => {
  I.fillField('user[login]', 'User');
  I.fillField('user[email]', 'user@user.com');
  I.fillField('user[password]', 'user@user.com');
  I.click('button');
});
```
within element is missing on github page. Should change it to `.js-signup-form`

Issue #656, opened 2017-08-09
Typo in acceptance doc


APshenkin:
http://codecept.io/acceptance/

There is no ability to locate elements this way: 
```
// locate element by text
I.seeElement({text: 'press me'});
```

Issue #490, opened 2017-04-25
Docs are misleading


imvetri:
### Details

* CodeceptJS version: 0.5 +
* NodeJS Version: v6
* Operating System: mac
* Nightmare version (if related)

following doesnt work
```js
I.amOnPage('/admin', [{'Authorization': 'Basic '+token}]);
```
following works
```js
I.amOnPage('/admin', {'Authorization': 'Basic '+token});
```

#### Conclusion
docs is misleading. If this real issue, I can send a PR for corrections.

DavertMik:
Yes, please update them. In this case you should edit this file https://github.com/Codeception/CodeceptJS/blob/master/lib/helper/Nightmare.js

PR #939, opened 2018-02-15
[Documentation] Small doc fixes

reubenmiller:
#### Highlights
* Fixes following documentation issues: #490, #658, #656, #732


docs/webapi/grabSource.mustache
```diff
@@ -0,0 +1,6 @@
+Retrieves page source and returns it to test.
+Resumes test execution, so should be used inside a generator with yield operator.
+
+```js
+let pageSource = yield I.grabSource();
```
DavertMik: better to use `await` now

reubenmiller: ok. done.

PR #940, opened 2018-02-15
[Tests] Improving reliability of the CI tests (Travis and Circle CI)

reubenmiller:
#### Highlights

* Added headless testing to test (seems more reliable than xvfb)
* Updated selenium/standalonechrome 3.9.1
* [Puppeteer] Added retries to Puppeteer tests
* [Puppeteer] Added some waits in test cases when switching/closing tabs
* Various Travis and Circle CI config changes (setting shm_size and DBUS_SESSION_BUS_ADDRESS)
* Refactored REST Helper tests to use async/await (for better readability)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,5 +33,4 @@ before_script:
 - sleep 10
 script:
 - npm build && npm test
-- './node_modules/.bin/mocha test/rest'
 - './node_modules/.bin/mocha test/helper/${HELPER}_test.js'
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ before_install:
 - docker pull selenium/standalone-chrome:3.9.1-actinium
 before_script:
 - docker run -d --net=host --shm-size=2g selenium/standalone-chrome:3.8
+- export DBUS_SESSION_BUS_ADDRESS=/dev/null
 - export DISPLAY=:99.0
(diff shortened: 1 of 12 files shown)


.travis.yml
```diff
@@ -21,9 +21,10 @@ addons:
 services:
 - docker
 before_install:
-- docker pull selenium/standalone-chrome:3.8
+- docker pull selenium/standalone-chrome:3.9.1-actinium
 before_script:
-- docker run -d --net=host selenium/standalone-chrome:3.8
+- docker run -d --net=host --shm-size=2g selenium/standalone-chrome:3.8
+- export DBUS_SESSION_BUS_ADDRESS=/dev/null
```
DavertMik: just wondering what's that?

reubenmiller: The `--shm-size=2g` option is as per the recommendation on the selenium/standalone-chrome repo.

As for the DBUS_SESSION_BUS_ADDRESS setting, the selenium docker images actually sets this in it's image due to a bug with selenium. I added it to the setting in case the environment used to run nightmare needed it...I ran the CI tests without the setting and it failed with a timeout every now and then, though after it, it seemed to help. Though this isn't exactly the most scientific approach.

DavertMik:
Tests are passing. Thank you!

PR #944, opened 2018-02-16
Fix incorrect description of waitUntilExists

JLRishe:
The description of `waitUntilExists` currently says (emphasis added) "Waits for element to **not be** present on page (by default waits for 1sec)." Prior to #934 this was (counterintuitively) true, but the method's behavior has now been changed to be consistent with its name.

### DIFF
--- a/docs/helpers/WebDriverIO.md
+++ b/docs/helpers/WebDriverIO.md
@@ -1385,7 +1385,7 @@ I.waitUntil(() => window.requests == 0, 5);
 
 ## waitUntilExists
 
-Waits for element not to be present on page (by default waits for 1sec).
+Waits for element be present on page (by default waits for 1sec).
 Element can be located by CSS or XPath.
 
 ```js


reubenmiller:
@JLRishe The docs for the helpers are actually generated from individual .mustache template files and not in the .md directly.

Though there has been some discussion on the dev slack channel about the `waitUntilExists` and it will probably be removed as the function is a duplicate of `waitForElement`. There was a mix up with the `waitForElement`, `waitUntilExists` and `waitForStaleness` that will be fixed in the next release which should be ready in the next few days.

Thanks for the PR :) though I don't think we'll be able to merge it as it will be superseded shortly when the new release drops.

JLRishe:
@reubenmiller Thanks for the clarification! I'll go ahead and close this PR since I modified the wrong file.

DavertMik:
I'm reverting this change and deprecating this method

PR #948, opened 2018-02-17
[AfterSuite|After] Allow AfterSuite and After test hooks to be defined after the first Scenario

### LINKED ISSUE #857: AfterSuite can't be defined after first Scenario

I expect `AfterSuite` to run if it is placed at the end of a test file.

#### What do you get instead?

It does not run. It only runs when you define `AfterSuite` before the first `Scenario`.

For some reason, [context.AfterSuite](https://github.com/Codeception/CodeceptJS/blob/master/lib/interfaces/bdd.js#L121-L123) does not register the hooks directly.  Instead, the hooks are only registered when [AddScenario](https://github.com/Codeception/CodeceptJS/blob/master/lib/interfaces/bdd.js#L46-L51) is called for the first time.

This means any AfterSuite defined after the first Scenario will not be registered. This is unexpected and undocumented behavior.

I haven't submitted a fix because there might be a good reason that `AfterSuite` does not register the hook directly.

reubenmiller:
### Highlights

Allow `AfterSuite` and `After` test hooks to be defined after the first Scenario. This allows for a more flexible and natural placement of these two test hooks. Previously these two hooks needed to be placed before the first `Scenario`. Fixes #857

#### Example
```js

Feature('My new feature');

BeforeSuite((I) => {
  console.log('Before suite');
});

Before((I) => {
  console.log('Before each test');
});

Scenario('Test Case 1', (I) => {
    I.amOnPage('/');
});

Scenario('Test Case 2', (I) => {
  I.amOnPage('/');
});

After((I) => {
  console.log('After each test');
});

AfterSuite((I) => {
  console.log('Bfter suite');
});
```

### DIFF
--- a/lib/interfaces/bdd.js
+++ b/lib/interfaces/bdd.js
@@ -39,22 +39,6 @@ module.exports = function (suite) {
         fn = opts;
         opts = {};
       }
-      /**
-       * load hooks from arrays to suite to prevent reordering
-       *
-       */
-      if (!afterAllHooksAreLoaded) {
-        afterAllHooks.forEach((hook) => {
-          suites[0].afterAll(hook[0], hook[1]);
-        });
-        afterAllHooksAreLoaded = true;
-      }
-      if (!afterEachHooksAreLoaded) {
-        afterEachHooks.forEach((hook) => {
-          suites[0].afterEach(hook[0], hook[1]);
-        });
-        afterEachHooksAreLoaded = true;
-      }
       if (suite.pending) {
         fn = null;
       }
@@ -124,7 +108,7 @@ module.exports = function (suite) {
     };
 
     context.AfterSuite = function (fn) {
-      suites[0].afterAll('AfterSuite', scenario.injected(fn, suites[0], 'afterSuite'));
+      afterAllHooks.unshift(['AfterSuite', scenario.injected(fn, suites[0], 'afterSuite')]);
     };
 
     context.Background = context.Before = function (fn) {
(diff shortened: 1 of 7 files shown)


lib/interfaces/bdd.js
```diff
@@ -124,15 +124,15 @@ module.exports = function (suite) {
     };
 
     context.AfterSuite = function (fn) {
-      afterAllHooks.unshift(['AfterSuite', scenario.injected(fn, suites[0], 'afterSuite')]);
+      suites[0].afterAll('AfterSuite', scenario.injected(fn, suites[0], 'afterSuite'));
```
APshenkin: this is revert changes of https://github.com/Codeception/CodeceptJS/commit/a79a7252b50b445f64fa2fd419badfc3a1cb7888#diff-70b2c1ced4ceeaeff1de3ac57c280b54

This was maked to set right ordering for after hooks.

Expected behavior should be : _beforeSuite hooks from helpers -> BeforeSuite from test -> _before hooks from helpers -> Before from test - > Test steps -> _failed hooks from helpers (if test failed) -> After from test -> _after hooks from helpers -> AfterSuite from test -> _afterSuite hook from helpers.

but if we will revert this changes then `_after` hooks from helpers will complete before `After` from test and `_afterSuite` hook from helpers will complete before `AfterSuite` from test.

reubenmiller: Ah ok. Thanks for the info. I’ll have another look at it then.

reubenmiller:
@APshenkin, I've found a way to solve it, though it uses internal api so it is not optimal.

The afterAll and After hooks are added each time the `Scenario()` is called or the `After()` or `AfterSuite()`, however before adding the hooks, it clears the existing hooks via (`suites[0]._afterAll = [];` (etc.).

Would that be ok? Or should I still look for a better solution?

Issue #943, opened 2018-02-16
waitForInvisible produces incorrect/misleading failure messages in certain situations. semantics are unintuitive


JLRishe:
I recently discovered that a number of my company's tests were failing on account of an incomplete understanding of how `waitForInvisible` works.

My tester writing the tests quite reasonably assumed that `waitForInvisible` would wait until the page contained _no visible elements_ matching the indicated selector.

After reading through the source code, I now realize that what it actually does is wait until there is _at least one non-visible element_ matching the indicated selector.

This means that if the page contains no elements matching the selector (e.g. if they had been removed from the page or never been there in the first place), `waitForInvisible` eventually indicates a failure with a factually incorrect message:

```js
I.waitForInvisible('#idonotexist');
```

The line above will wait and eventually fail with the message: `element (#idonotexist) still visible after 5 sec`. This is factually incorrect and misleading to someone reading the test results because `#idonotexist` is not "still visible". It doesn't exist at all!

Conversely, if the page contains a mixture of visible and non-visible elements matching a particular selector, `waitForInvisible` will succeed as long as at least one of them is non-visible.

```html
<div class="myclass" style="display: none">I am not visible</div>
<div class="myclass">I am visible</div>
```

```js
I.waitForInvisible('.myclass');   // succeeds
````

I realize that there is a `waitForStalenessOf` method which checks for the non-presence of elements in the DOM (at least this is true in the WebDriverIO helper; in the Protractor helper, it's just an alias for `waitForInvisible`). However, I would think that from a testing perspective, it would be more common and generally better practice to simply test whether an element is visible in the page or not, and not get mired down in the implementation details of hidden vs. not present in the DOM. But Codecept doesn't seem to provide any method for simply checking whether the page does or does not have a particular visible element.

So my questions are:

- Is `.waitForInvisible`'s behavior, as described and demonstrated above, really intentional? It seems strange and counterintuitive to me.
- If it _is_ intentional, can the failure message at least be improved to more clearly convey what it's checking for? E.g. `no invisible element (#idonotexist) found after 5 sec`

### Details

* CodeceptJS version: 1.1.4
* NodeJS Version:  8.9.4
* Operating System: Win10
* WebDriverIO version: 4.10.2
* Configuration file:  unchanged from default

reubenmiller:
You raise some valid points. I can investigate this and maybe look at adding some options which can wait for just 1 matching eememts or all of them. 

[...]

PR #950, opened 2018-02-17
waitForInvisible - implement treatNonPresentAsInvisible option (#943)

JLRishe:
In #943, I observed that `waitForInvisible`, rather than simply waiting for there to be no visible elements matching the selector, actually waits until there is _at least one_ invisible element matching the selector (at least in most implementations this is true; see below).

I posited that this was somewhat unintuitive from a testing perspective because when a test is waiting for an element to disappear, the test shouldn't care about whether the element is actually hidden or removed from the DOM.

I also indicated that the failure message "element (...) still visible after ... sec" was misleading, because it could say the element was "still visible" even in situations where the element didn't exist at all.

@reubenmiller said that the current behavior _is_ desirable as the default behavior, but suggested that I add a new option that would make `waitForInvisible` agnostic w.r.t. elements' presence or non-presence in the DOM, so that is what I have done in this pull request. I have also changed the failure messages to be consistent across all helpers, as "element (...) not invisible after ... sec".

One important note about these changes:

I noticed some inconsistency in `waitForInvisible`'s behavior across the different helpers in their implementation prior to my changes:

- In the Protractor helper, its behavior was to succeed either if the element was non-present or hidden, because that's what `EC.invisibilityOf(el)` does.
- In the Puppeteer helper, its behavior was to succeed either if the element was non-present or hidden, **but only when using a CSS selector**, because that's what `EC.invisibilityOf(el)` does. When using an XPath selector, its behavior was to wait for at least one hidden element, just like in the Nightmare and WebdriverIO helpers.

I have modified all helpers to be consistent and added unit tests to verify this, but this would imply a breaking change for anyone relying on the previously inconsistent behavior in the Protractor and Puppeteer helpers.

### DIFF
--- a/docs/webapi/waitForInvisible.mustache
+++ b/docs/webapi/waitForInvisible.mustache
@@ -7,3 +7,4 @@ I.waitForInvisible('#popup');
 
 @param locator element located by CSS|XPath|strict locator
 @param sec time seconds to wait, 1 by default
+@param treatNonPresentAsInvisible when true, succeeds when the element is either hidden or not present. otherwise, succeeds only if the element is hidden.
--- a/lib/helper/Nightmare.js
+++ b/lib/helper/Nightmare.js
@@ -42,6 +42,7 @@ let withinStatus = false;
  * * `waitForAction`: (optional) how long to wait after click, doubleClick or PressKey actions in ms. Default: 500.
  * * `waitForTimeout`: (optional) default wait* timeout in ms. Default: 1000.
  * * `windowSize`: (optional) default window size. Set a dimension like `640x480`.
+ * * `treatNonPresentAsInvisible`: (optional, default: false) - when true, waitForInvisible will succeed when no elements matching the selector are present
  *
  * + options from [Nightmare configuration](https://github.com/segmentio/nightmare#api)
  *
@@ -63,6 +64,7 @@ class Nightmare extends Helper {
       keepCookies: false,
       js_errors: null,
       enableHAR: false,
+      treatNonPresentAsInvisible: false,
     };
 
     this.isRunning = false;
@@ -854,17 +856,18 @@ class Nightmare extends Helper {
   /**
    * {{> ../webapi/waitForVisible }}
    */
-  waitForInvisible(locator, sec) {
+  waitForInvisible(locator, sec, treatNonPresentAsInvisible = null) {
     this.browser.options.waitTimeout = sec ? sec * 1000 : this.options.waitForTimeout;
+    const allowNonPresent = (treatNonPresentAsInvisible === null) ? this.options.treatNonPresentAsInvisible : treatNonPresentAsInvisible;
     locator = new Locator(locator, 'css');
 
-    return this.browser.wait((by, locator) => {
+    return this.browser.wait((by, locator, allowNonPresent) => {
       const el = window.codeceptjs.findElement(by, locator);
-      if (!el) return false;
+      if (!el) return allowNonPresent;
       return el.offsetParent === null;
-    }, locator.type, locator.value).catch((err) => {
+    }, locator.type, locator.value, allowNonPresent).catch((err) => {
       if (err.message && err.message.indexOf('.wait() timed out after') > -1) {
-        throw new Error(`element (${JSON.stringify(locator)}) still visible after ${sec} sec`);
+        throw new Error(`element (${JSON.stringify(locator)}) not invisible after ${sec} sec`);
       } else throw err;
     });
   }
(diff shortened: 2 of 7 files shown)


DavertMik:
Thanks! Yes, there is an inconsistency and I'd like to fix it. However, in our API we don't allow options in functions. That reduces readability of the scenario:

I wait for inivsible ".user", 2, true - does not produce any readable output

So there can be two options: implement a new function or switch everything to behavior you propose. I'd prefer the second one, as it makes more sense for testing. When element is hidden we can't be sure if it was hidden or it was destroyed: it's implementation detail which is known to developers but not known for users. Espcially when we do a wait not an assertion. 

So if you want to wait for element to be removed 

```js
I.waitForElementDetached()
```
If you want for element to be hidden or remvoed:

```js
I.waitForInvisible()
```
If you want element to be hidden but not removed:

```
I.waitForInvisible('#el');
I.seeElementInDOM('#el');
```

I think all this scenarios should be mentioned in documentation to solve possible confusion

JLRishe:
@DavertMik Thank you for your response.

I agree that having the new behavior be the default seems more sensible because, as you say and as I have said, it allows waiting for an element to go away without getting involved in implementation details.

But when I proposed this in #943, @reubenmiller told me that the current functionality is desirable because the new behavior would incorrectly succeed if the test writer used an incorrect selector.

He said:

> Personally I think it would be wise to have an option where the user can specify whether the assertion should fail if no matches are found or not. I think the default should fail if no matches are found, so that it ensures the user has the correct selector and the error is not silently passing.

He suggested that I implement this new behavior as a config option and as an optional parameter on the function:

> So the requirements of the feature would be something like:
>
> - Default setting in the helper config (should be implemented in each helper)
> - Function should take an additional option (overriding the default setting). If no option is specified then it takes the configured default

So that's how I implemented it.

It seems that you and he have different opinions on how this should be approached. I'm happy to implement it in any manner you the maintainers prefer and I can see that you are the owner of this project so your voice has higher weight, but is there any chance you could confer with him and come to a consensus?

reubenmiller:
Hi Guys,

@JLRishe Sorry for the misdirection. I didn't notice the lack of options on the helper functions and I was giving adviced based on causing the least amount of change (though that turned out to cause even more problems :D)


> When element is hidden we can't be sure if it was hidden or it was destroyed: it's implementation detail which is known to developers but not known for users. Espcially when we do a wait not an assertion.

Yeah I agree with this. I guess this was the original idea from @JLRishe, until I made the detour. The more that I think about it, the better this idea sounds.


> Personally I think it would be wise to have an option where the user can specify whether the assertion should fail if no matches are found or not. I think the default should fail if no matches are found, so that it ensures the user has the correct selector and the error is not silently passing.

Ok, maybe there is a better way which is more inline with the **CodeceptJS** way.

So what about the following guidelines:

1. `I.waitFor*` functions should not care if the waiting condition is already fulfilled when executing the function. If the user wants make sure that the item is visible or exists before the `waitFor*` function is called, then they should use an Assert like `I.seeElement` etc. And then the user is free to use asserts after the `waitFor*` function to check the expected state of the element (i.e. if it should be in the DOM still or removed completely)

1. `I.dontSee` and `I.dontSeeElement` should not differentiate between an item being not visible and not existing. For the user they don't care if the item is hidden or if it has been removed. In both these cases, the item is not visible to the user.


### Example

```js
// Optional: Check that the element is visible before waiting
I.seeElement('#item'); 

// Note: If the item is somehow already invisible or removed, then the function will continue without throwing an error.
I.waitForInvisible('#item');

// Optional: Use if the item should still be in DOM but not visible
I.seeElementInDOM('#item');
```


### Suggested Changes?

These are only my thoughts, so let me know what you think, and I can update them accordingly.

* Change `waitForInvisible` function so that it does not care if the item does not exist in the DOM, or it waits for the element to become invisible.
* Change I.dontSeeElement and I.dontSee functions (in all helpers) to not care if the element exists in the DOM or not. Does this make sense? @DavertMik
* Update docs to include information about the new behaviour

JLRishe:
Those suggested changes sound good to me.

I suspect that most of the `dontSeeElement` and `dontSee` functions already work that way, but I can make sure that the unit tests for them are covering all the bases and make updates to their implementation as needed.

reubenmiller:
@DavertMik Do you agree with my latest suggestions?

JLRishe:
@DavertMik Could you let me know if you are in agreement with @reubenmiller's proposal? I'm holding off on making any further changes until I have your approval to proceed with them.

DavertMik:
CodeceptJS API should not take any options as params. They make tests complicated and not readable. If you feel like some behavior is missing it's better to add another function than adding an option to current one.

Currently `waitForInvisible` waits for element to absent on page or not being visible
We also added `waitForDetached` function. 

So I think all needed behaviors are already there. Am I missing something?

reubenmiller:
Sorry @JLRishe I'll have to close this PR as I have submitted a new PR which fixes the problem that you reported. There was a big mix up regarding the existing behaviour of `I.waitForInvisible` and my misguided information definitely did not help.

Thanks for the effort by providing the PR but I'll have to close this now as it has been superseded by #972. I've only been a collaborator on this project for the last few weeks, so I'm still getting into the ethos of the project as a whole. 

Please don't let this discourage you from future PRs, your time and effort is valued :) I also mentioned you on the PR that I just submitted to recognise your efforts.

Happy coding :)

PR #955, opened 2018-02-20
grabNumberOfVisibleElements to return also 0 visible elements

michaltrunek:
Assertion removed from grabNumberOfVisibleElements method to accept also zero visible elements.
- this method should return also value if no elements are displayed on the page
- issue was discussed also on slack channel


DavertMik:
Yes! Thank you

PR #958, opened 2018-02-20
Removed comma after last argument

micron:
`npx codeceptjs init` crashes because of unnecessary comma in `./codeceptjs/lib/mocha_factory.js:42`

Stack trace:

    SyntaxError: Unexpected token )
      at createScript (vm.js:56:10)
      at Object.runInThisContext (vm.js:97:10)
      at Module._compile (module.js:549:28)
      at Object.Module._extensions..js (module.js:586:10)
      at Module.load (module.js:494:32)
      at tryModuleLoad (module.js:453:12)
      at Function.Module._load (module.js:445:3)
      at Module.require (module.js:504:17)
      at require (internal/module.js:20:19)
      at Object.<anonymous> 
      (/Users/miron/workspace/project/node_modules/codeceptjs/lib/container.js:4:22)

Node version `6.13.0`

Removing the last comma fixes this issue.

### DIFF
--- a/lib/mocha_factory.js
+++ b/lib/mocha_factory.js
@@ -38,7 +38,7 @@ class MochaFactory {
     if (reporterOptions['codeceptjs-cli-reporter'] && attributes) {
       Object.defineProperty(
         reporterOptions, 'codeceptjs/lib/reporter/cli',
-        attributes,
+        attributes
       );
       delete reporterOptions['codeceptjs-cli-reporter'];
     }


micron:
Trailing commas were introduced in node 8 so the README.md should be updated because the current codeceptjs version (1.1.5) does not work with node 6. Closing this pull request.

reubenmiller:
@micron Thanks for submitting a pull request. However this change is no longer necessary as v1.1 now requires nodejs 8.9.1 or newer. The change was required so that async/await could be natively supported which hugely simplified the code (IMO). In the newer nodejs versions the trailing comma is not an issue and in fact not having it will through a lint error (hence why the tests failed on the PR).

I would highly recommending running the linting before submitting (using `npm run lint`)as it will tell you if everything is formatted as expected before running the ci tests. 

Again thanks, PR requests are always welcome :) though I think I’ll have to close it unless anyone can persuade me otherwise.

reubenmiller:
:D i was too slow. That’s what I get for typing all of this on my phone.

PR #959, opened 2018-02-21
change library for get params

APshenkin:
Current library doesn't support syntax like this:

```
Scenario('test', async I => {
 ...
});

It returns error `Error: Object of type async I is not defined in container`
```

With updated library this will be fixed

### DIFF
--- a/test/data/sandbox/testscenario_test.testscenario.js
+++ b/test/data/sandbox/testscenario_test.testscenario.js
@@ -14,7 +14,8 @@ Scenario('Simple async/await test', async (I) => {
   console.log(text);
 });
 
-Scenario('Should understand async without brackets', async (I) => {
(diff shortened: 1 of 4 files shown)


DavertMik:
Could you add a test for this case?

APshenkin:
@DavertMik ready

test/data/sandbox/testscenario_test.testscenario.js
```diff
@@ -13,3 +13,8 @@ Scenario('Simple async/await test', async (I) => {
   const text = await I.stringWithScenarioType('async/await');
   console.log(text);
 });
+
+Scenario('Should understand async without brackets', async (I) => {
```
DavertMik: It was supposed to be
```
async I => {
```
right?

PR #961, opened 2018-02-24
[Puppeteer] Added waitForEnabled, waitForValue and waitNumberOfVisibleElements

reubenmiller:
#### Highlights

* Added waitForEnabled, waitForValue and waitNumberOfVisibleElements
* Added some extra tests to the WebDriverIO helper and change some error messages so they are the same as those generated by the Puppeteer helper (limited to the new Puppeteer functions).

Now I'm pretty sure the Puppeteer Helper is feature parity with the WebDriverIO helper  🎉 🍻

### DIFF
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -1235,6 +1235,97 @@ class Puppeteer extends Helper {
   }
 
   /**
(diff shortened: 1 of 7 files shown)


DavertMik:
That's great. Thank you!

Issue #678, opened 2017-08-24
Execute bootstrap script once in mutilple execution


kinlu:
Now I am integrating Browserstack tunnel setup in the bootstrap of conf.js.  When the multiple execution happens, the bootstrap script get called multiple times.  Only one tunnel for a group of test is up while the other tunnels fail as well as their tests.

I tried to use localIdentifier in the capabilities setting however it doesn't work.  The test instances seem can't match up the identifiers.

Only way to work is
- Single test instance with browserstack tunnel setup in the bootstrap
- Multiple test instances with browserstack tunnel setup in a separate command line window

Is it possible to execute bootstrap script once in mutilple execution to allow multiple test instances with browserstack tunnel setup in the bootstrap?

APshenkin:
@kinlu currently no, multiple execution creates several instances of codeceptjs. Each codeceptjs process launch bootstrap file. I think, we can create a new option that will launch bootstrap before launching codeceptJS instances and will skip bootstrap for these instances.

reubenmiller:
Instead of only calling the bootstrap script once for `run-multiple`, a new hooks `bootstrapAll` is proposed in PR #962. This hook is only executed once. Also there is a new `teardownAll` hook proposed in the PR.

reubenmiller:
Released in v1.1.6

PR #962, opened 2018-02-24
[Multiple-Run] Adding new bootstrapAll and teardownAll hooks

reubenmiller:
#### Highlights

* New `bootstrapAll` and `teardownAll` hooks which are used when running the `run-multiple` command. The `bootstrapAll` hook is trigger before any test suites have started, and `teardownAll` is triggered once all of the suites have finished. The two hooks also have event equivalents; `multiple.before` and `multiple.after`. #957, #678 

#### Background

Previously only `bootstrap` and `teardown` hooks were available. When using the `run-multiple` command, the `bootstrap` and `teardown` hooks trigger every time a suite starts and finishes. This makes them ill suited to perform general setup and cleanup tasks when using the [Multiple Execution](https://codecept.io/advanced/#multiple-execution) command.

### DIFF
--- a/lib/command/run-multiple.js
+++ b/lib/command/run-multiple.js
@@ -87,7 +87,6 @@ module.exports = function (suites, options) {
     // throw error if no configuration
     if (suiteConf.browsers.length === 0) throw new Error(`Browser ${browser} not found in multiple suite "${suiteName}" config`);
 
-    totalSubprocessCount += suiteConf.browsers.length;
     const suiteForks = runSuite(suiteName, suiteConf);
     if (Array.isArray(suiteForks)) {
       forksToExecute.push(...suiteForks);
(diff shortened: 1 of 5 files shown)


lib/command/run-multiple.js
```diff
@@ -71,28 +84,41 @@ module.exports = function (suites, options) {
         return b === browser;
       });
     }
-
     // throw error if no configuration
     if (suiteConf.browsers.length === 0) throw new Error(`Browser ${browser} not found in multiple suite "${suiteName}" config`);
 
-    runSuite(suiteName, suiteConf);
+    totalSubprocessCount += suiteConf.browsers.length;
```
APshenkin: looks like here this count is unnecessary. You reassign is later `totalSubprocessCount = forksToExecute.length;`

reubenmiller: yep, I'll remove it now. Oversight on my part.

PR #968, opened 2018-02-27
[Nightmare] Changed visibility detection

reubenmiller:
#### Highlights
* Changing detection of the Nightmare visibility detection to reflect that of the native Nightmarejs implementation (i.e. using the `offsetWidth` and `offsetHeight` attributes instead of `offsetParent`. Proposed in #614

#### References
* [Jquery note](https://makandracards.com/makandra/1339-check-whether-an-element-is-visible-or-hidden-with-javascript)
* [Nightmare implementation](https://github.com/segmentio/nightmare/blob/master/lib/actions.js#L72)


DavertMik:
Thanks!

Issue #976, opened 2018-03-02
[feature] Enable glob patterns in `config.test` / `Codecept.loadTests`


sveneisenschmidt:
Hello, first thank you for the work you have put in here!  
(This is the first part of a feature request. For the second part please go to #978.)

**Sidenote**
We at @trivago Leipzig use CodeceptJS for end-to-end testing and are working currently on a simple way to run multiple jobs in parallel without using grep and tags.

**Feature**
A first step into having a working solution is to enable CodeceptJS to load tests not only by grep/tags but also by the already working but limited glob support in  `Codecept.loadTests`. A simple modification by introducing a non-breaking change to how `glob.sync` is called would enable everyone to use glob patterns in their configuration.

The true power with this feature lies in combination with #978. If both features get accepted it will be possible do dynamically create `run-multiple` configuration by the amount of overall scenarios and distribute them evenly between a defined amount of runs. For a working examples please have a look [here](https://github.com/sveneisenschmidt/codeceptjs-boilerplate/blob/master/lib/create-multiple.js) and [here](https://github.com/sveneisenschmidt/codeceptjs-boilerplate/blob/master/codecept.conf.js).

**Technical explaination**
I'd like to propose to remove the `fsPath.resolve` call for passing the pattern to glob. This enables glob to perform a bracket expansion. To make sure that glob is still being executed in the correct directory I'd pass the `cwd` option to glob and set it to `global.codecept_dir` to basically mimic the same `fsPath.resolve` call. To have a absolute path pushed into `this.files` the second `fsPath.resolve` inside the loop is kept.

**Example**
```js
{
  "tests": "{./file1_test.js,./file2_test.js}"
}
```

**Todos**
- [x] Add `test-runner` service in `docker-compose.yml`
- [x] Write test case and verify nothing is breaking
- [x] Adapt `loadTests` function
- [x] Create a Pull Request in the fork https://github.com/sveneisenschmidt/CodeceptJS/pull/1
- [x] Feedback by maintainers
- [x] Create a Pull Request in this repository

What do you think? 

PS: If you want to see why glob support is needed please look at my demo repository:

- https://github.com/sveneisenschmidt/codeceptjs-boilerplate#todo
- https://github.com/sveneisenschmidt/codeceptjs-boilerplate/blob/master/lib/create-multiple.js
- https://github.com/sveneisenschmidt/codeceptjs-boilerplate/blob/master/codecept.conf.js

sveneisenschmidt:
[...]

PR #982, opened 2018-03-03
[feature] Enable glob patterns in `config.test` / `Codecept.loadTests`

sveneisenschmidt:
See #976

### DIFF
--- a/lib/codecept.js
+++ b/lib/codecept.js
@@ -85,7 +85,12 @@ class Codecept {
    */
   loadTests(pattern) {
     pattern = pattern || this.config.tests;
-    glob.sync(fsPath.resolve(global.codecept_dir, pattern)).forEach((file) => {
+
+    const options = {
+      cwd: global.codecept_dir,
+    };
+
+    glob.sync(pattern, options).forEach((file) => {
       this.testFiles.push(fsPath.resolve(file));
     });
   }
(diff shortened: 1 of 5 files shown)


reubenmiller:
Great work :) Though can you please fix the linting issues because it is causing the ci tests to fail.

sveneisenschmidt:
@DavertMik @reubenmiller  `Hound` and `Codacy` seems to have a different opinion here. Any idea how to fix this? I use the default .editorconfig and the file just looks fine.

reubenmiller:
@sveneisenschmidt what do you get when you run the linting locally? `npm run lint`. If it fails there then you might be able to use `npm run lint-fix`.

Though it's strange because in the travis tests, it runs `npm run lint` and I've already seen that the tests are passing (so that would suggest the above commands are working fine).

sveneisenschmidt:
@reubenmiller Yep, seems all good:

```console
> docker run -i -t --rm -v $PWD:/data:cached -w /data node:9 /bin/bash
root@c981a7f8eb1e:/data# npm install
added 136 packages and updated 1 package in 92.467s

root@c981a7f8eb1e:/data# npm run lint
> codeceptjs@1.1.6 lint /data
> eslint bin/ examples/ lib/ test/ translations/

root@c981a7f8eb1e:/data# npm run lint-fix
> codeceptjs@1.1.6 lint-fix /data
> eslint bin/ examples/ lib/ test/ translations/ --fix

root@c981a7f8eb1e:/data
```

How to proceed now? :)

DavertMik:
I'm still pretty not sure what to use .hound or codacy so I'm keeping both to look who serves better :)
Anyway, this is not that important, everything can be fixed with fixing lint.
So I'm nerging this PR too

PR #991, opened 2018-03-08
Fixing I.click so it accepts an explicit xpath selector. Fixes #989

### LINKED ISSUE #989: CodeceptJS's Puppeteer library not working with "complex" Xpath selectors
Use CodeceptJS's Puppeteer library with Xpath selectors

#### What do you get instead?
xpath check is  not covering all xpath "templates": 
```js
function isXPath(locator) {
  return locator.substr(0, 2) === '//' || locator.substr(0, 3) === './/';
}
```
is not covering the case where the fist child of a xpath must be selected, like in the example below:

```js
'(//*[@data-protractor-id="recipient-select"] //*[contains(concat(" ", normalize-space(@class), " "), "ui-select-choices-row")])[1]'
```

### Details

* CodeceptJS version: 1.1.6
* NodeJS Version: 8.9.1
* Operating System: OS High SIerra
* Puppeteer 1.1.1
* Configuration file:

reubenmiller:
### Highlights

* Fixed bug with `I.click` when using an object selector with the xpath property. It only affected the **Puppeteer** and **Protractor** helpers. Fixes #989
* Added tests for all browser helpers to check for this condition

#### Example which now works
```js
Scenario('I click using custom xpath selector', (I) => {
  I.amOnPage('/');
  I.click({
    xpath: '(//*[@class])[1]',
  });
});
```

### DIFF
--- a/lib/helper/Protractor.js
+++ b/lib/helper/Protractor.js
@@ -1130,7 +1130,7 @@ async function proceedIsChecked(assertType, option) {
 async function findClickable(matcher, locator) {
   locator = new Locator(locator);
   if (!locator.isFuzzy()) {
-    const els = await this._locate(locator.value, true);
+    const els = await this._locate(locator, true);
     assertElementExists(els, locator.value);
     return els[0];
   }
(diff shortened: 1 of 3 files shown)


DavertMik:
Thanks!

PR #993, opened 2018-03-08
[WIP][feature] Introduce chunking to run-multiple

sveneisenschmidt:
Reworked `run-multiple` command to support chunking by dynamical expanding the configuration. Also simplifies internal expansion by browsers.

- [x] Tests pass
- [x] No breaking changes
- [x] Documentation

**Changelog:**
- added documentation
- renamed: Suite -> Run
- renamed: runSuite -> executeRun
- renamed: preferredSuites -> selectedRuns
- added new class in run-multiple/collection
- added new class in run-multiple/run
- added new set of functions in run-multiple/chunk
- added: collection.prepareRuns - creates unified suites coniguration based on preferredSuites (preferred suites are either suites passed via --all or explicitly)
- added: collection.prepareChunks - expands configuration by (n) chunks if chunks are configured
- added: collection.prepareBrowsers - expands configuration by (n) browsers if browsers are configured
- added: collection.filterSelectedBrowsers - filters configuration by requested browsers set via selectedRuns
- added: chunking also supports a simple version of grep
- added: chunking also supports a callback instead of a number
- updated: runSuite -> executeRun
  - browser argument is removed, it is now retrieved through a unified browser config
  - forking over multiple browsers is removed, it became obsolete as soon as we prepared
    and expanded the whole configuration beforehand, including chunks and browsers


lib/suite.js
```diff
@@ -0,0 +1,126 @@
...
+          if (preferredSuiteName === suiteConfig.parentSuiteName && preferredSuiteBrowserName !== suiteBrowserName) {
+            delete suites[suiteName];
+          }
+        }
+      });
+    });
+  });
+
+  return suites;
+};
```
DavertMik: It looks to me this should be refactored as a class. That adds more readability to the code. 
`preferredSuites, suites, config` params are the same for all functions so I'd better add them to constructor of a class and make them functions inside it.

DavertMik:
Btw, currently `suite` represents a test file. 
We even have events fired:

* event.suite.before(suite) - before a suite
* event.suite.after(suite) - after a suite

which handled with BeforeAll, AfterAll hooks. The suite object comes from Mocha.
If this file does not correspond to this type of a suite it should be renamed

lib/suite.js
sveneisenschmidt: Yes, I’d like to introduce a collection object that holds the items including “suite” config, also manages add and remove. 

Second I’d like to have the prepare* and filter* functions as callbacks on that collection.

sveneisenschmidt:
@DavertMik I’d like to rename it to just ‘run’ and ‘collection‘.

lib/suite.js
DavertMik: Ok, let's call that `Collection`!

lib/chunk.js
```diff
@@ -0,0 +1,57 @@
...
+    return config.grep ? grepFiles(file, config.grep) : true;
+  });
+
+  const size = Math.ceil(files.length / config.chunks);
+  const chunks = splitFiles(files, size);
+  const chunkConfig = { ...config };
+
+  delete chunkConfig.chunks;
+
+  return chunks.map((chunkFiles) => {
```
DavertMik: I'd prefer to make `chunk` to be a class as well.
We need to allow users to replace `splitFiles` algorithm with a custom implementation.
So `splitFiles` could have a default implementation and an option to pass in a function which will replace it.

DavertMik:
@sveneisenschmidt any updates?

sveneisenschmidt:
@DavertMik Still on it 🙂

sveneisenschmidt:
@DavertMik @reubenmiller Please see my latest code changes and the updated Pull Request. :)

lib/command/run-multiple/collection.js
```diff
@@ -0,0 +1,158 @@
...
+ * Bootstraps a collection of runs, it combines user defined selection of runs
+ * to be executed and retrieves the run-specific configuration
+ */
+const createRuns = (selectedRuns, config) => {
+  return createCollection(config)
+    .prepareRuns(selectedRuns, config)
+    .prepareChunks(config)
+    .prepareBrowsers(config)
+    .filterSelectedBrowsers(selectedRuns, config)
+    .getRuns();
```
DavertMik: Very nice API!

DavertMik:
@APshenkin could you look into it as well?

Looks nice to me. Could you update documentation https://github.com/Codeception/CodeceptJS/blob/master/docs/advanced.md#multiple-execution with examples of splitting tests by chunks (by number or by function)

APshenkin:
@DavertMik Looks good for me

@sveneisenschmidt  good work!

DavertMik:
@sveneisenschmidt  So to finalize this we need some documentation added :)

DavertMik:
>  Have you seen my thread message in Slack about the effectiveness of chunking?

Not sure. We had issues with `glob` usage and that's last I see in dev channel.

sveneisenschmidt:
@DavertMik I‘ll give my best. Let’s do this!

DavertMik:
Looks awesome, thanks. I will merge it when I push my own changes prepared for 1.2 version

PR #998, opened 2018-03-14
[Protractor] Fixing bug with I.seeAttributesOnElements and I.seeCssPropertiesOnElement

reubenmiller:
#### Highlights

* `I.seeAttributesOnElements` and `I.seeCssPropertiesOnElements` were incorrectly passing when the attributes/properties did not match.


DavertMik:
:+1:

PR #1008, opened 2018-03-29
Proper hash handling for Puppeteer URL

gavoja:
Puppeteer's `page.url()` ignores hash. This is an attempt to fix it in order to maintain consistency with other helpers.

### DIFF
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -259,8 +259,7 @@ class Puppeteer extends Helper {
    * Gets page URL including hash.
    */
   async _getPageUrl() {
-    const hash = await this.executeScript(() => window.location.hash);
-    return this.page.url() + hash; // page.url() ignores hash.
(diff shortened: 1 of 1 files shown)


lib/helper/Puppeteer.js
```diff
@@ -255,6 +255,14 @@ class Puppeteer extends Helper {
     });
   }
 
+  /**
+   * Gets page URL including hash.
+   */
+  async _getPageUrl() {
+    const hash = await this.executeScript(() => window.location.hash);
```
DavertMik: Maybe we just should simply use `window.location.href` then without calling to `page.url()`
I fear that mixing both approaches can have side effects

gavoja:
Yes, we should, I realized this as well and I was planning to update the PR after the Easter (vacationing at the moment). Cheers!

DavertMik:
Great! Have a good holiday!

gavoja:
Fixed.

PR #1020, opened 2018-04-11
Add the method translation of Chinese

TechQuery:
One question: [ISO 639.1](http://www.loc.gov/standards/iso639-2/php/code_list.php) or [RFC 4646](http://www.ietf.org/rfc/rfc4646.txt) should I follow on the language name ?


DavertMik:
> One question: ISO 639.1 or RFC 4646 should I follow on the language name ?

Sorry, I have no idea 😟
So choose whatever you like and thanks. That's really awesome!

TechQuery:
I chose our customary usage, it's OK to merge it.

DavertMik:
Thanks!

TechQuery:
But the document on official website has not been updated.

PR #1044, opened 2018-04-30
[DOCS] Fix typo

mikementor:


DavertMik:
This doc is generated from the source code. Could you please update the source at https://github.com/Codeception/CodeceptJS/blob/master/lib/helper/Appium.js

Otherwise the patch wil be overwritten on next release

DavertMik:
Not sure why but it will be regenerated on release )

PR #1045, opened 2018-05-03
Use proper "def" command description

pablopaul:


DavertMik:
Thanks! Copy-paste can go wrong (sometimes) :smiley:

PR #1056, opened 2018-05-09
Remove misleading info

### LINKED ISSUE #1055: Step object docs are inconsistent
https://codecept.io/pageobjects/ states "Technically they are the same as PageObjects but with no locators inside them. StepObjects can inject PageObjects and use multiple POs to make a complex scenarios"

`codeceptjs go --type step`

creates a boilerplate file with the contents 

```js
// insert your locators and methods here
```

should "locators" be removed from boilerplate code or is the above docu wrong?

pablopaul:
Close #1055


DavertMik:
Thanks!

PR #1058, opened 2018-05-09
Updated typescipt definitions.

LukoyanovE:
Added Data, xData, xScenario definitions.
Added options for actor and Feature functions.

Rewrote declarations to "declare function ..." style to allow use of overloading for Scenario and xScenario. Now it allow use both with or without opts parameter.
Don't found variant to allow it with code style which has already been used. Therefore, rewrote all such declarations in single style.

Added declaration for codecept_helper, to remove errors in custom helpers. Maybe there are other ways to solve this, please, tell me if it's true. I'll remove it.

### DIFF
--- a/lib/command/definitions.js
+++ b/lib/command/definitions.js
@@ -21,14 +21,20 @@ interface ILocator {
   ios?: string;
 }
 
-declare const actor: () => CodeceptJS.{{I}};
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks, I  was working on improving type definitions today as well. Fortunately our changes do not contradict so I can merge it

PR #1096, opened 2018-06-05
Enable test runner for nested test directories

LukoyanovE:
"npm test" did not use tests in nested directories, for example "test/unit/assert"
see https://mochajs.org/#the-test-directory

Fixed test according to linter rules.


DavertMik:
Thanks!

PR #1099, opened 2018-06-06
[WIP] Cucumber / Gherkin support

DavertMik:
Run Gherkin scenarios inside CodeceptJS

selection_308

Implementation:

```js
const I = actor();

Given(/I have product with \$(\d+) price/, (price) => {
  I.addItem(parseInt(price, 10));
});
When('I go to checkout process', () => {
});

Then('I should see that total number of products is {int}', (num) => {
  I.seeNum(num);
});
Then('my order amount is ${int}', (sum) => { // eslint-disable-line
  I.seeSum(sum);
});

```


bitcoder:
a clarification: this PR doesn't include support for the Cucumber JSON report, does it?

PR #1101, opened 2018-06-06
Feature: add filter function to DataTable

LukoyanovE:
**What**
Add function "filter" to DataTable, the same as Array.Filter.

**Why**
To not use all dataset. Specially, when data table is big and part of it should not be used for several tests.


LukoyanovE:
It's look like selenium server for appium test was offline.
Could anybody restart semaphoreci build?

DavertMik:
Very nice! thank you

PR #1105, opened 2018-06-10
Added a draft of document about CI/CD integration.

ontytoom:
@DavertMik and everyone looking at this, please review the doc and fix anything that is not good enough ;-) let me know any feedback please.

### DIFF
--- a/docs/ci-cd.md
+++ b/docs/ci-cd.md
@@ -1,0 +1,70 @@
+# Integrating CodeceptJS into Your CI/CD Pipeline
+
+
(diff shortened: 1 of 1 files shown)


DavertMik:
Good as a start but lacks examples for different CIs. Depending on CI the instructions can be very different

DavertMik:
Sorry, this guide is outdated and misleading. For instance, in Puppeteer we don't need XWindow environment... But we should start a new CI/CD guide with better description of usageideas for new

PR #1110, opened 2018-06-14
properly propagate errors in asyncWrapper around async functions

### LINKED ISSUE #1109: Errors in async function of PageObjects are catched in Before
I tried to setup async stuff with a page object in the Before function of a Testsuite. If the asynchronous code fails, the error is catched and the test proceeds. I broke down the issue to a minimal example (which is still quite big): 

```js
// codecept.conf.js
module.exports = {
  ...
  'include': {
        'test': './test'
   },
  ...
};
```

```js
// scenario.js
Feature('Test');

Before(async function (test) {
    await test.foo();
});

Scenario('Test', function () {
});
```

```js
// ./test.js
module.exports = {
  foo: async function () {
    await bar();
  },
};
```

### Details

* CodeceptJS version: 1.2.1
* NodeJS Version: v9.11.1
* Operating System: OSX High Sierra
* WebDriverIO: 4.11.0

As far as I can see, the error is catched by the asyncWrapper (in codeceptjs/lib/container.js) and saved inside the recorder as 'asyncError' but is not thrown after the end of Before. I would expect the test to fail in Before because of that async error.

johnyb:
I don't really see, why this wrapper would be needed. It silently catches any
errors thrown by async functions. Even if a rejected promise is being
returned. This is very different from just returning a promise. In case
it's rejected, the error will bubble up and fail any test.

This fixes #1109 for me.

### DIFF
--- a/lib/container.js
+++ b/lib/container.js
@@ -168,6 +168,7 @@ function createSupportObjects(config) {
     return function () {
       return f.apply(this, arguments).catch((e) => {
         recorder.saveFirstAsyncError(e);
+        throw e;
       });
     };
   };


DavertMik:
To be honest I don't remember what problem does it solve. @APshenkin do you remember why we have it?

johnyb:
I also tried to imagine, what this might do. My first version of this change looked a little different. I just threw the exception again, after adding it to the recorder object. It had the same effect and fixed the bug for me. Then I tried to just remove the complete block and it didn't make a difference. So that's what I wanted to have discussed here :)

May be something is failing in the acceptence tests, now. This might give a hint on why this would be needed.

APshenkin:
@DavertMik @johnyb 
I created this wrapper because of saving proper error that leads test failure.

When you use await in Page Objects and then use them in test code, then use it in code there are some cases, when the error appears in pageObject, but test go farther and fail on next step (because we write all things in one promise before)

I don't have a good example currently, but I think that we should create several tests for these cases

johnyb:
I see and this use case sounds reasonable to me. Let me update my branch in this case …

johnyb:
I adjusted the code, so errors are now properly propagated, so they can be handled elsewhere.

> foo function doesn't has return value. await bar() produce error but it's not return to test. Because it just silently fails

async functions implicitly return a promise which rejects in case an error is thrown. Otherwise this promise will resolve with "undefined" as an argument, if nothing is explicitly returned. So that's expected behaviour. The reason this can not be handled is, that the wrapper just catched the error and did not handle it properly. It just adds it to the recorder and everything higher in the stack trace will never know about the error if not rethrown. That's what this patch now does.

This still fixes #1109 and still also propagates the error to the recorder object.

DavertMik:
Ok, let's try it!

Issue #1069, opened 2018-05-17
Using restart=false with webdriverio backend leaves browser instances running


hubidu:
#### What are you trying to achieve?

I am setting restart to false in order to speed up overall test execution by not restarting browsers between tests.

#### What do you get instead?

Browsers are not restarted between tests but after the test run the used browser instance is not terminated. There seems to be even a new one with a blank page to be opened.

### Details

node_modules\\.bin\\codeceptjs run --debug --verbose
CodeceptJS v1.2.0
Using test root "C:\Users\stefan.huber\projects\e2e-dashboard-client\examples\codeceptjs"

Failing test scenarios --
   [1] Starting recording promises
   Emitted | suite.before ([object Object])
 > [Session] Starting singleton browser session
 The test will fail when an element on the page can not be found
   Emitted | test.before ([object Object])
   Emitted | test.start ([object Object])
   Emitted | step.before (I am on page "http://www.check24.de")
   Emitted | step.after (I am on page "http://www.check24.de")
   Emitted | step.before (I click "#this-selector-does-not-exist")
   Emitted | step.after (I click "#this-selector-does-not-exist")
   Emitted | step.start (I am on page "http://www.check24.de")
 * I am on page "http://www.check24.de"
   Emitted | step.passed (I am on page "http://www.check24.de")
   Step finished in 1.347 sec
   Emitted | step.finish (I am on page "http://www.check24.de")
   Emitted | step.start (I click "#this-selector-does-not-exist")
 * I click "#this-selector-does-not-exist"
 > [SmartWait (2000ms)] Locating #this-selector-does-not-exist in 2000
   [1] Error | Error: Clickable element #this-selector-does-not-exist was not found by text|CSS|XPath
   Emitted | step.failed (I click "#this-selector-does-not-exist")
   Step finished in 2.069 sec
   Emitted | step.finish (I click "#this-selector-does-not-exist")
   [1] Error | Error: Clickable element #this-selector-does-not-exist was not found by text|CSS|XPath
   [1] Starting <teardown> session
   Emitted | test.failed ([object Object])
   Emitted | test.finish ([object Object])
   [1] <teardown> Stopping recording promises
 > Screenshot has been saved to C:\Users\stefan.huber\projects\e2e-dashboard-client\examples\codeceptjs\__out\The_test_will_fail_when_an_element_on_the_page_can_not_be_found.failed.png
 × FAILED in 3855ms

   [2] Starting recording promises
[...]

PR #1113, opened 2018-06-15
fix-1069: fix WebDriverIO browser opening after tests with option restart=false

LukoyanovE:
Fix typo (I think), made in [1049](https://github.com/Codeception/CodeceptJS/pull/1049/files#diff-c873b3f20f4ccf0b26a50fbebbf074d6R359) pull request.

Now WebDriverIO does not start driver/browser session after test completion.
#1069


DavertMik:
Thanks!

PR #1115, opened 2018-06-16
Run waitUntil script in browser context for WebDriverIO

GREENpoint:
`I.waitUntil(fn)` runs `fn` in context of node, not browser.

### DIFF
--- a/docs/helpers/Nightmare.md
+++ b/docs/helpers/Nightmare.md
@@ -530,7 +530,7 @@ First parameter can be set to `maximize`
 ## saveScreenshot
 
 Saves a screenshot to ouput folder (set in codecept.json).
-Filename is relative to output folder. 
+Filename is relative to output folder.
 Optionally resize the window to the full available page `scrollHeight` and `scrollWidth` to capture the entire page by passing `true` in as the second argument.
 
 ```js
@@ -788,6 +788,16 @@ button: "left"
 **Parameters**
 
 -   `event`  
+
+## waitForFunction
+
+Waits for a function to return true (waits for 1sec by default).
+Running in browser context.
+
+```js
+I.waitForFunction(() => window.requests == 0);
+I.waitForFunction(() => window.requests == 0, 5);
+```
 
 ## wait
 
--- a/docs/helpers/Protractor.md
+++ b/docs/helpers/Protractor.md
@@ -1280,7 +1280,6 @@ I.waitForFunction(() => window.requests == 0, 5);
 
 ## waitUntil
 
-Deprecated. Use `waitFor` instead.
 Waits for a function to return true (waits for 1sec by default).
 Running in Node context.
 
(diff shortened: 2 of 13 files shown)


LukoyanovE:
Not sure, but waitUntil should be run in context of Node.js.
Test can wait for any condition, not only browser condition.

For example I can wait for database result.
If you wanna wait browser script result, you should use `browser.executeScript()` function in `waitUntil` arguments.

I think you can add it as separate function, like `waitUntilBrowserScriptResult` or something similar.

@DavertMik , what do you think about this request?

LukoyanovE:
Look at original WebDriverIO definition of function
[docs](http://webdriver.io/api/utility/waitUntil.html)

GREENpoint:
@LukoyanovE But `waitUntil` of `Puppeteer` helper runs in browser context. And documentation of helper hints me this behavior too https://codecept.io/helpers/WebDriverIO/#waituntil

truesrc:
Its PR very correct. waitUntil must run in browser context. If you want to execute a function in the context of a nodeJS, then there are common js.

truesrc:
You need too in this line
`return client.waitUntil(async () => (await client.execute(fn)).value, aSec, timeoutMsg);`
do `return client.waitUntil(async () => (await client.execute(fn)).value, aSec * 1000, timeoutMsg);`

LukoyanovE:
This PR is important initiative, but look at this:
> This wait command is your universal weapon if you want to wait on something. It expects a condition and waits until that condition is fulfilled with a truthy value.

http://webdriver.io/api/utility/waitUntil.html

waitUntil should be universal, it should have possibility to wait anything, like this
```I.waitUntil((await I.grabText("#some_important_field")) === "some_important_text_condition")```

If you wanna wait function result from browser context, you should use `I.waitUntil(await I.executeAsyncScript( ...... ))`.
We can make alias for this, like `waitForFunction`, to be the same as Puppeteer's `waitForFunction`.

In Puppeteer project, this feature was not called a different name without reason

truesrc:
> If you wanna wait function result from browser context, you should use I.waitUntil(await I.executeAsyncScript( ...... )).

I cannot because `waitUntil` from http://webdriver.io/api/utility/waitUntil.html may work only with method from http://webdriver.io/api.html not with common js DOM-BOM method.

truesrc:
https://github.com/Codeception/CodeceptJS/blob/master/lib/helper/Puppeteer.js#L1593
This is the same that proposed @GREENpoint

DavertMik:
Historically current `waitUntil` was introduced earlier than Puppeteer's so I'm not sure if we break compatibility if we change that behavior. So this is my only concern.

So I really like `waitForFunction` method as proposed by @LukoyanovE 

However, for `waitUntil` in Puppeteer we can deprecate it's current behavior and rename it to `waitForFunction` as well. 

Also, we need a test for them...

So next steps:

1. add new function `waitForFunction`
2. write a test for it (to stabilize it's api)
3. add `waitForFunction` to puppeteer, protractor, and nightmare.
4. derecate `waitUntil` in Puppeteer

DavertMik:
Thanks! Good job!

Issue #1111, opened 2018-06-14
Parallel execution error on version 1.2.1 - "Maximum call stack size exceeded"


jsparks19:
#### What are you trying to achieve?
Run multiple tests in parallel

#### What do you get instead?

RangeError: Maximum call stack size exceeded
    at normalizeStringWin32 (path.js:33:30)
    at Object.resolve (path.js:338:20)
    at sync (C:\...\npm\node_modules\codeceptjs\node_m
odules\mkdirp\index.js:68:14)
    at sync (C:\...\npm\node_modules\codeceptjs\node_m
odules\mkdirp\index.js:77:24)
    at sync (C:\...\npm\node_modules\codeceptjs\node_m
odules\mkdirp\index.js:78:17)
    at sync (C:\...\npm\node_modules\codeceptjs\node_m
odules\mkdirp\index.js:78:17)
    at sync (C:\...\npm\node_modules\codeceptjs\node_m
odules\mkdirp\index.js:78:17)
    at sync (C:\...\npm\node_modules\codeceptjs\node_m
odules\mkdirp\index.js:78:17)
    at sync (C:\...\npm\node_modules\codeceptjs\node_m
odules\mkdirp\index.js:78:17)
    at sync (C:\...\npm\node_modules\codeceptjs\node_m
odules\mkdirp\index.js:78:17)
path.js:33
function normalizeStringWin32(path, allowAboveRoot) {

Seems related to #1068 

### Details

* CodeceptJS version: 1.2.1
* NodeJS Version: 8.9.4
* Operating System:  Windows 7
* WebDriverIO Version: 4.12.0
* Configuration file:

```js
{
  "output": "./output",
  "helpers": {
    "WebDriverIO": {
      "url": "url",
      "browser": "chrome",
      "host": "localhost",
      "port": "4444",
      "coloredLogs": true,
      "restart": false,
      "windowSize": "maximize",
      "keepCookies": true,
      "smartWait": 5000
    }
  },
  "multiple": {
    "blah": {
      "chunks": 2,
      "grep": "@blah"
    }
  },
  "include": {
    ...
  },
  "mocha": {},
  "bootstrap": false,
  "teardown": null,
  "hooks": [],
  "tests": "./tests/**/*/*_test.js",
  "timeout": 10000,
  "name": "codecept"
}
```

sveneisenschmidt:
Have you tried adding ’"browsers": ["chrome"]’ to your blah config in multiple?

[...]

Issue #1084, opened 2018-05-29
Output directory already exists error when running run-multiple parallel


mikecbrant:
#### What are you trying to achieve?

Run all tests in parallel.  Tests run successfully in serial.

#### What do you get instead?

Error indicating that top-level test output directory already exists, which is odd considering the step right before it actually creates the sub-directory for chunk output.  It is not clear why there would be an attempt to create this top level output directory after the sub-directory is already created.

Output below:

```bash
PS C:\Users\a0225521\Documents\git\ticom-mobile> .\node_modules\.bin\codeceptjs run-multiple parallel --debug
Initializing application for UI tests in forked process.
creating output directory: C:\Users\a0225521\Documents\git\ticom-mobile\test-output\parallel:chunk1__browser_Nightmare__1
C:\Users\a0225521\Documents\git\ticom-mobile\node_modules\mkdirp\index.js:90
                    throw err0;
                    ^

Error: EEXIST: file already exists, mkdir 'C:\Users\a0225521\Documents\git\ticom-mobile\test-output'
    at Object.fs.mkdirSync (fs.js:909:18)
    at sync (C:\Users\a0225521\Documents\git\ticom-mobile\node_modules\mkdirp\index.js:71:13)
    at sync (C:\Users\a0225521\Documents\git\ticom-mobile\node_modules\mkdirp\index.js:77:24)
    at sync (C:\Users\a0225521\Documents\git\ticom-mobile\node_modules\mkdirp\index.js:78:17)
    at sync (C:\Users\a0225521\Documents\git\ticom-mobile\node_modules\mkdirp\index.js:78:17)
    at sync (C:\Users\a0225521\Documents\git\ticom-mobile\node_modules\mkdirp\index.js:78:17)
    at sync (C:\Users\a0225521\Documents\git\ticom-mobile\node_modules\mkdirp\index.js:78:17)
    at sync (C:\Users\a0225521\Documents\git\ticom-mobile\node_modules\mkdirp\index.js:78:17)
    at sync (C:\Users\a0225521\Documents\git\ticom-mobile\node_modules\mkdirp\index.js:78:17)
    at sync (C:\Users\a0225521\Documents\git\ticom-mobile\node_modules\mkdirp\index.js:78:17)
```

> Provide test source code if related

Did not include test files as these would not seem to be problem.  They run correctly in serial.

### Details

[...]

PR #1117, opened 2018-06-19
Fixing invalid output directory path for run-multiple

mikecbrant:
This should address related issues #1084 and #1111 for which root cause was invalid output directory path on Windows due to colon character being present (i.e. `parallel:...`).  This was causing `mkdirp` module to infinitely recurse, leading to max call stack size error.

Proposed solution uses pre-existing `clearString()` function in `lib/utils.js` to replace invalid characters in string.


DavertMik:
Thanks!

mikecbrant:
Glad to help. Anything else needed from my end prior to npm package publishing?

PR #1118, opened 2018-06-20
Send non-nested array of files to custom parallel execution chunking function.

mikecbrant:
Currently, an (unexpected?) array wrapper is applied to the array of files being passed to the parallel execution custom file chunking function.  I would think expected behavior here is just getting the array of files as noted in online documentation.

This might need to be considered a breaking change depending on how users are currently using this functionality.

### DIFF
--- a/docs/helpers/Nightmare.md
+++ b/docs/helpers/Nightmare.md
@@ -530,7 +530,7 @@ First parameter can be set to `maximize`
 ## saveScreenshot
 
 Saves a screenshot to ouput folder (set in codecept.json).
-Filename is relative to output folder. 
+Filename is relative to output folder.
 Optionally resize the window to the full available page `scrollHeight` and `scrollWidth` to capture the entire page by passing `true` in as the second argument.
 
 ```js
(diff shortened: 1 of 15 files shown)


DavertMik:
@sveneisenschmidt any ideas?

sveneisenschmidt:
@DavertMik That's the fix we need. All correct.

DavertMik:
Thanks!

mikecbrant:
The custom function is shown below. All I am trying to achieve is to execute all test files in parallel, which might seem to be something that ought to be an available option.

```
    multiple: {
        parallel: {
            chunks: (files) => {
                return files[0].reduce(
                    (arr, file) => {
                        arr.push([ file ]);
                        return arr;
                    },
                    []
                );
            },
            browsers: ['Nightmare']
        }
    }
```

A tangential question as working with parallel execution.  Are there any reporting set-ups you might recommend here?  Finding the interleaving of stdout to terminal making it hard to see consolidated result report at the end.

PR #1131, opened 2018-06-27
Add info on using 'direct' Protractor driver

xt1:
Protractor needs 'directConnect': true in the config in order to use the 'direct' driver.
Otherwise it keeps trying to talk to the Selenium server.


DavertMik:
Thanks!

PR #1138, opened 2018-06-29
Reinstate typescript generation for custom steps

xt1:
generates TS declarations for custom steps (like login(user,pass) ) and say(msg).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ testpullfilecache*
 .DS_Store
 package-lock.json
 yarn.lock
+/.vs
--- a/lib/command/definitions.js
+++ b/lib/command/definitions.js
@@ -83,6 +83,7 @@ declare function session(selector: string, callback: Function): Promise<any>;
 declare function session(selector: ILocator, callback: Function): Promise<any>;
 declare function session(selector: string, config: any, callback: Function): Promise<any>;
 declare function session(selector: ILocator, config: any, callback: Function): Promise<any>;
+declare function pause(): void;
 
 declare const codecept_helper: any;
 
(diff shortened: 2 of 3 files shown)


DavertMik:
Nice! Do you think it would be possible to add TS definitions to page objects as well? 
That would be really helpful

xt1:
I will see what I can do about page objects - need to become a bit more familiar with the guts of the system. Back on monday.

I guess each page object needs an interface in the steps.d.ts file.

DavertMik:
> I guess each page object needs an interface in the steps.d.ts file.

Yeah, I think it as well. I didn't get any further as I didn't have enough time for that. But that's what really needs to be done.

Here is what you can use:

```js
container.support() // returns all support objects: PageObjects and actor
```

you can go through them and get all methods out of them

xt1:
Now generates page object interfaces and adds placeholders to the callback signature.

```js
type ICodeceptCallback = (i: CodeceptJS.I, aPage:any, anotherPage:any) => void;


declare namespace CodeceptJS {
  export interface I {
    amOutsideAngularApp() : void,
    amInsideAngularApp() : void,
    amOnPage(url: string) : void,
// ... etc
   say(msg: string) : void,
    retryStep(opts: string) : void,
  }

  export interface aPage {
    send_a_Form(email: string, password: string) : void,
  }

  export interface anotherPage {
    sendForm(email: string, password: string) : void,
    bing(words: string) : void,
    bong() : void,
  }
}
```

xt1:
Ready to merge.

DavertMik:
Thanks!

PR #1148, opened 2018-07-12
Fix typo in `include:` configuration, add actor include.

wwerner:
The configuration example said `includes:` with an additional `s` instead of `include:`.
I also added the actor import.


DavertMik:
Thanks!

PR #1150, opened 2018-07-13
Add usage description for Page Fragments in Scenarios and Page Objects

wwerner:


docs/pageobjects.md
```diff
@@ -145,6 +145,34 @@ module.exports = {
...
+  within(modal.root, function () {
+    I.see('Login failed');
+  })
+});
+```
+
+To use a Page Fragment within a Page Object, you need to `require` it on top of the Page Object file:
+
+```js
+const I = actor()
```
DavertMik: I thnk it's better to add `;`

docs/pageobjects.md
```diff
@@ -145,6 +145,34 @@ module.exports = {
...
+const modal = require('../fragments/modal')
```
DavertMik: I thnk it's better to add `;`

docs/pageobjects.md
```diff
@@ -145,6 +145,34 @@ module.exports = {
...
+
+module.exports = { 
+  doStuff() {
+    ...
+    modal.accept()
```
DavertMik: I thnk it's better to add `;`

DavertMik:
Thank you!

PR #1151, opened 2018-07-13
fix: xpath detection fails when it starts with `./`

BenoitZugmeyer:
Context: when using `Locator.clickable.self`, the xpath generated starts with `./self::...`, but the locator was not considered as `xpath` (fallbacking to `css`) so it failed to select the element.


DavertMik:
Thanks! Nice catch.
This was not so important before this locator builder was introduced.

PR #1205, opened 2018-09-12
Enable config override when running parallel tests

svarlet:
Just found that the `--override` CLI flag was ignored when running tests in parallel. This fixes it.

Also, using the `--colors` with run-multiple exit with `error: unknown option `--colors'`


DavertMik:
Thanks!

PR #1212, opened 2018-09-18
Allure reports

DavertMik:
* **Allure reports introduced**
* Introduced **CodeceptJS plugins**
* Added `allure` plugin
* Improved detecting metasteps for BDD and nested steps
* Setting `within` as metastep
* Moved saving a screenshot on failure into `screenshotOnFail` plugin
* Added `stepByStepReporter` plugin which displays simple HTML report with screenshots for each taken step
* Include screenshots of failed tests into allure reports
* Per step screenshots for when `stepByStepReports` plugin is enabled

Documentation is coming


DavertMik:
selection_352

Added screenshot

package.json
```diff
@@ -36,11 +36,13 @@
     "test": "mocha test/unit --recursive && mocha test/runner --recursive"
   },
   "dependencies": {
+    "allure-js-commons": "^1.3.2",
```
APshenkin: Looks like this dep should be in dev dependencies. If I don't need allure repots, I don't want to install it with codeceptjs

DavertMik: It's a very basic package. It doesn't install Allure by itself, only some JS APIs.
I'd rather keep it in package.json to simplify Allure installation. 

In future versions, this will probably be changed.

PR #1214, opened 2018-09-19
Correctly extract default extension

Akxe:
By using this regexp it will correctly match extensions from `config.test`, that fallow format like `./**/*_test.js`, or `./**/*.test.js`. 
Example on [regexer.com](regexr.com/3vo4k)

Also, forcing `_helper.js` extension in JS world is weird, therefor option to make it `.helper.js` would be nice.

### DIFF
--- a/lib/command/generate.js
+++ b/lib/command/generate.js
@@ -26,7 +26,7 @@ module.exports.test = function (genPath) {
   output.print('Creating a new test...');
   output.print('----------------------');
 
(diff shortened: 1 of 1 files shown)


DavertMik:
> Also, forcing _helper.js extension in JS world is weird, therefor option to make it .helper.js would be nice.

Thanks, please create that as an issue for 2.0 version

Akxe:
@DavertMik How do I make issue for 2.0?

PR #778, opened 2017-11-15
add option noreverse

APshenkin:
Add option --noreverse for cli reporter
This option check should we reverse failed steps output or not.

PR #1223, opened 2018-09-24
Improve cli configuration and output

APshenkin:
1. Add new option `notruncate`. If set `true`, cli logs will not be truncated
2. Add new option `noreverse`. If set `true`, CLI failure steps will not be in reversed order (see https://github.com/Codeception/CodeceptJS/pull/778)
3. Add new option `outputStyle`. Supports types: `actor` and `pageObject`.
So now there are several configuration options:
 a. Steps: true, outputStyle: actor. Then in console you will see only I methods calls without pageObject history (as before 1.3.0 release)
 b. Steps: true, Debug: false, outputStyle: pageObject. Then in console you will see pageObjects methods with arguments, but only top level (that are used in Scenario)
 c. Steps: true, Debug: true, outputStyle: pageObject. Then in console you will see pageObjects methods with arguments, with internal calls of pageObjects and I method

so. lets take a look on example:
Test:
```
Scenario.only('Should get network info', async (I, blockchain) => {
  const resp = await blockchain.getBlockchainInfo(200, coin);

  const fee = await blockchain.getRippleFees();

  I.assertEqual(resp.fees, fee);
  I.assertOk(resp.height);
  I.assertOk(resp.height > 12801200);
});
```

a version output:
```
@ripple --
  Should get network info
    I send rabbit request "blockchain.get-blockchain-info", {"network":"XRP"}, {}
    I assert equal "200", 200, "Request 'blockchain.get-blockchain-info' return 200 status, but expected 200"
    I send request "https://s.altnet.rippletest.net:51234/", "POST", {"headers":{"Content-Type":"application/json"},"json":{"method":"server_state"}}
    I assert status code 200, 200
    I assert equal "10", "10"
    I assert ok "12886351"
    I assert ok true
  ✔ OK in 505ms
```

b version output:
```
@ripple --
  Should get network info
    blockchain get blockchain info 200, "XRP"
    blockchain get ripple fees 
    I assert equal "10", "10"
    I assert ok "12886360"
    I assert ok true
  ✔ OK in 355ms
```

c version output:
```
@ripple --
  Should get network info
    blockchain get blockchain info 200, "XRP"
      sbus request body 200, "blockchain.get-blockchain-info", {"network":"XRP"}
          I send rabbit request "blockchain.get-blockchain-info", {"network":"XRP"}, {}
          I assert equal "200", 200, "Request 'blockchain.get-blockchain-info' return 200 status, but expected 200"
    blockchain get ripple fees 
        I send request "https://s.altnet.rippletest.net:51234/", "POST", {"headers":{"Content-Type":"application/json"},"json":{"method":"server_state"}}
        I assert status code 200, 200
    I assert equal "10", "10"
        I assert equal "10", "10"
    I assert ok "12886368"
        I assert ok "12886368"
    I assert ok true
        I assert ok true
  ✔ OK in 1351ms
```


DavertMik:
Sorry I need some more time to get through this )

DavertMik:
Ok, my opinion is in this setup cli reporter is doing too much. 
The code looks over complicated and I don't think I will be able to support it.

I think instead of providing 3 output modes it's better to provide different reporters and the ability to switch between them.

PR #1224, opened 2018-09-24
Make actor type definition useful

Akxe:
By adding type annotation to `customSteps`, we get hint, that `this` is actually `CodeceptJS.I` and therefore we can do much more, while having some type checks.


DavertMik:
Thanks! That should be useful

PR #1231, opened 2018-09-26
Fix to make the event.test.before hook send test arg

jamesbrucepower:
I cannot see any reason why the test wouldn't be passed as an argument to the _before event.

In my case I have written a helper which makes use of the test.title and test.parent.title to make a get request to my backend system to present a pretty HTML page displaying the test feature (I call it context) and test title before each test.

I could not make it work without this, despite trying all sorts of ways to inject another event.test.before event in the _beforeSuite() hook.  It was becoming insanely complicated for what really is a simple fix in the main code base

Regards
James

### DIFF
--- a/lib/listener/helpers.js
+++ b/lib/listener/helpers.js
@@ -39,7 +39,7 @@ module.exports = function () {
 
   event.dispatcher.on(event.test.before, (test) => {
     // schedule config to revert changes
-    runAsyncHelpersHook('_before');
+    runAsyncHelpersHook('_before', test, true);
   });
 
   event.dispatcher.on(event.test.passed, (test) => {


DavertMik:
Thanks!

Yeah, that makes sense. And I hope this will help others. 
Btw, plugins can be useful in this case. They have much richer API. https://codecept.io/hooks/#plugins
Please take a look.

Also, I'd ask you for input on documentation. Maybe you have ideas on how to restructure current documentation. I know the topic of extending CodeceptJS via helpers/plugins is not very clear. So if you have some thoughts how to improve that, please tell.

jamesbrucepower:
@DavertMik 

Regarding the documentation feedback, here are my first thoughts

I appreciate that Hooks and Plugins are similar, but perhaps having them in their own section in the documentation would help.  Its not clear to me without studying the code how at first glance hooks and plugins differ.  Perhaps it would be a good idea to show a comparison on the features of each.
When I looked at plugins, the most helpful thing I found was looking at the existing plugins in CodeceptJS/lib/plugin. 

I am still slightly confused about what can be imported/required from codeceptjs and how those things get used within the main codecept library.  For example, I see from the plugin 
https://github.com/DavertMikexamples that it

`const Container = require('../container');`

This is the first time I've seen this being used, I couldn't find any reference to it in the documentation and it turns out its the piece of the puzzle I've been missing, i.e. its the thing that actually holds all the helpers and support objects, it wasn't clear what was doing that, I did look through the code but unfortunately hadn't looked at that particular file, i.e. CodeceptJS/lib/container.js

Where can I find a list of objects/functions etc that can be used in a plugin and what those things do.

On the plus side, I like the fact that with plugins, its a lot easier to control and setup what events are used within the plugin, unlike with hooks where you had to implement a method with a particular name, perhaps that could be used in the comparison I referred to above.

PR #1238, opened 2018-10-02
fix multiple.parallel.chunks

ngadiyak:
if multiple.parallel.chunks param in codecept.conf.js set as environment variable, codeceptjs throw "chunks is neither a finite number or a valid function".

example:

```
    multiple: {
        parallel: {
            chunks: process.env.CHUNKS || 20,
            ...
        }
```


DavertMik:
Thanks!

PR #1248, opened 2018-10-08
Move Allure screenshot attachment

DevinWatson:
Moving to Allure screenshot attaching inside the screenshot try/catch block

If there is an exception saving a screenshot it should not try and add a file that does not exist to the Allure report.


DavertMik:
Makes sense. Thank you!

PR #1253, opened 2018-10-10
Remove some extra back-ticks in documentation

samwarnick:
I noticed what I assume are some extra back-ticks in the documentation.


DavertMik:
Yep, but this should be placed into docblocks inside source code

PR #1255, opened 2018-10-11
using Axios for REST API call

PeterNgTr:


PeterNgTr:
Better to add them as Community Helpers. https://github.com/Codeception/CodeceptJS/wiki/Community-Helpers

PR #1259, opened 2018-10-16
Fix retryFailedStep name in example

pablopaul:


DavertMik:
This docs were taken from the plugin source code
SO I fixed it there and rebuilt the docs.
Thanks!

PR #1270, opened 2018-10-19
Fix/bootstrap starting for def

LukoyanovE:
Removed config's bootstrap and teardown calls for `codeceptjs def` and `codeceptjs list`

Before, bootstrap has started (and set up connection to browserstack, for example) before `def` command, which is unnecessary.

Also, teardown has started after these commands. `TypeError: done is not a function` error occurred.


lib/codecept.js
```diff
@@ -66,12 +66,11 @@ class Codecept {
...
   /**
-   * Executes hooks and bootstrap.
-   * If bootstrap is async, second parameter is required.
+   * Executes hooks.
    *
-   * @param {*} done
+   * @param {() => any} [done]
    */
-  bootstrap(done) {
+  runHooks(done) {
```
DavertMik: looks like `runHooks` are synchronous now, so `done` is not needed and callback functions.
Also I'm going to deprecate custom hooks in config so this code should be 100% synchronous

lib/command/gherkin/steps.js
```diff
@@ -13,16 +13,22 @@ module.exports = function (genPath, options) {
...
-    for (const step of Object.keys(steps)) {
-      output.print(`  ${output.colors.bold(step)} \n    => ${output.colors.green(steps[step].line || '')}`);
-    }
-    output.print();
-    if (!Object.keys(steps).length) {
-      output.error('No Gherkin steps defined');
-    }
+    if (err) throw new Error(`Error while running init :${err}`);
+
+    codecept.runBootstrap((err) => {
```
DavertMik: bootstrap should not be run for listing steps definition (as it's not executed for def)

lib/command/gherkin/snippets.js
```diff
@@ -34,77 +36,81 @@ module.exports = function (genPath, options) {
...
 
-    const files = [];
-    glob.sync(config.gherkin.features, { cwd: global.codecept_dir }).forEach((file) => {
-      if (!fsPath.isAbsolute(file)) {
-        file = fsPath.join(global.codecept_dir, file);
-      }
-      files.push(fsPath.resolve(file));
-    });
-    output.print(`Loaded ${files.length} files`);
+    codecept.runBootstrap((err) => {
```
DavertMik: gherkin:snippets should not execute bootstrap (as it's omitted for def/list commands as well)

LukoyanovE:
Done 1st and 3rd.

About 2nd:
`bootstrap` runs for every browser config in `multiple` section.
It should not be called for `run-multiple`.
We have `bootstrapAll`/`teardownAll` for multiple hook

DavertMik:
Awesome, thanks!

PR #1274, opened 2018-10-22
Adds clickLink method to Puppeteer helper

luismanuel001:
Fixes the `click()` method so by default it doesn't trigger a `waitForNavigation()` and instead adds a new `clickLink()` method to trigger a `waitForNavigation()` when clicking an element that should trigger a navigation.

### DIFF
--- a/docs/helpers/Puppeteer.md
+++ b/docs/helpers/Puppeteer.md
@@ -314,6 +314,19 @@ I.click({css: 'nav a.login'});
 -   `locator`  clickable link or button located by text, or any element located by CSS|XPath|strict locator
 -   `context`  (optional) element to search in CSS|XPath|Strict locator
 
+## clickLink
+
+Performs a click and waits for navigation before moving on.
+
+```js
+I.click('Logout', '#nav');
+```
+
+**Parameters**
+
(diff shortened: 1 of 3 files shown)


DavertMik:
Can you explain usecase for both methods?

A new method can be confusing. Maybe we can add a 3rd parameter where you can specify 3rd parameter: `{ waitForNavigation: false }` ?

luismanuel001:
Yes, a 3rd parameter or extra options was my other approach, but then I saw how you treat other click related methods that use `proceedClick()`, ex. `doubleClick()` and `rightClick()`, and they both have their own method without extra parameters, so I thought this was more in the same line of thought.

But if you think it will be better with the 3rd parameter, so it becomes something like `click(locator, context = null, waitForNavigation = false)`, just let me know and I'll update the PR with those small changes.

DavertMik:
What do you think for default behavior, should waitForNavigation be enabled for clicks or disabled?

luismanuel001:
For my use case, on SPA's I think it it makes more sense that the default is false, since you do a lot more clicks without `waitForNavigation()` than with it.

DavertMik:
Ok, then I'm more convinced by your initial idea - `clickLink`....
Having 2 methods more usable in this case.

luismanuel001:
Thanks man

PR #1278, opened 2018-10-23
Extend browsers description in docs

LukoyanovE:
Add more expressive examples for `run-multiple` command.
Move "parallel" section after "multiple", cause the first uses examples of browsers from the second.


DavertMik:
Thanks. Makes sense!

PR #1281, opened 2018-10-23
Created autoDelay plugin

DavertMik:
While developing a test suite for one of my clients I found useful to put a small delay before and after click and fillField to ensure that page could respond to this actions. Here is a plugin which an easily add delays to your actions.

---

# autoDelay
 Sometimes it takes some time for a page to respond to user's actions.Depending on app's perfromance this can be either slow or fast.

 For instance, if you click a button and nothing happens - probably JS event is not attached to this button yet. Also, if you fill field and input validation doesn't accept your input - maybe because you typed value too fast.

 This plugin allows to slow down tests execution when a test running too fast.
It puts a tiny delay for before and after action commands.

```js
"plugins": {
   "autoDelay": {
     "enabled": true
   }
}
```

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -37,6 +37,46 @@ Launch Allure server and see the report like on a screenshot above:
 
 -   `config` **Any** 
 
+# autoDelay
+
+Sometimes it takes some time for a page to respond to user's actions.
+Depending on app's perfromance this can be either slow or fast.
+
+For instance, if you click a button and nothing happens - probably JS event is not attached to this button yet
+Also, if you fill field and input validation doesn't accept your input - maybe because you typed value too fast.
+
+This plugin allows to slow down tests execution when a test running too fast.
+It puts a tiny delay for before and after action commands.
+
+Commands affected (by default):
+
+-   `click`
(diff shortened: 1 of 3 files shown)


tfiwm:
Should the test not be related to some condition to continue? In my opinion this is a dangerous feature which slows down every test -> Maybe it is better to give him a selector or function which needs to return true instead of a static delay?

DavertMik:
@tfiwm In some cases, this is just not possible to do so. 
If action is taken faster than JavaScript can respond to it you can't rely on such test.

For instance, if a button is rendered to page but a JavaScript handler is not attached yet, there is nothing you can specify to wait for. Visually a button exists. It is clickable. But if you click too fast it won't react. Basuically this plugin slows down test execution to more human-like speed. It is still much faster than human, it affects only some commands. But it stabilizes a test as it allows a browser to react to our actions.

tfiwm:
But even than it could be flaky because it relates at the condition the the js attached the handler

-> i would prefer to at a label (css class, ...) as initial and remove this one if i attach the handler -> than i have my condition which i can wait and the test can't be flaky.

Issue #1264, opened 2018-10-17
[Enhancement] Introduce run-multiple into CODECEPT_ARGS in Docker-Compose


kinlu:
#### What are you trying to achieve?

Currently, the arguments placed in ```CODECEPT_ARGS``` is only attached to the command ```codeceptjs run```.  
```
environment:
      - CODECEPT_ARGS=--debug
```
It would be great to have some way to run ```codeceptjs run-multiple``` to do parallel testing with docker with Docker-Compose.

kinlu:
I would suggest make "-multiple" as a separate argument, like "run -multiple".

DavertMik:
I would not run multiple inside one container. 
I think it's better to have multiple containers for that.

DavertMik:
But If you implement it I will accept such PR

PeterNgTr:
@kinlu @DavertMik Implemented it here https://github.com/Codeception/CodeceptJS/pull/1285

PR #1285, opened 2018-10-25
Introduce run-multiple into CODECEPT_ARGS in Docker-Compose

PeterNgTr:
This to fulfill this https://github.com/Codeception/CodeceptJS/issues/1264


DavertMik:
Thanks!

Issue #1241, opened 2018-10-04
Test fails on Safari start with selenium-standalone


alexTimosh:
such error appeared Immediately after startup and session was stopped. At the same time test works perfectly on a earlier codeceptjs versions (i.e. ˜0.6).
```Suite 1 --
    [1] Starting recording promises
    Emitted | suite.before ([object Object])
  Check Sign In form
    Emitted | test.before ([object Object])
    [1] Error | Error: Unknown timeout type in 'timeouts': 'type'
Build info: version: '3.12.0', revision: '7c6e0b3', time: '2018-05-08T15:15:08.936Z'
System info: host: 'Alexeys-MacBook-Pro.local', ip: '', os.name: 'Mac OS X', os.arch: 'x86_64', os.version: '10.13.6', java.version: '10.0.2'
Driver info: driver.version: unknown
    Emitted | test.failed ([object Object])
  ✖ "before each" hook: codeceptjs.before for "Check Sign In form" in 20853ms
    [1] Error | Error: Unknown timeout type in 'timeouts': 'type'
Build info: version: '3.12.0', revision: '7c6e0b3', time: '2018-05-08T15:15:08.936Z'
System info: host: 'Alexeys-MacBook-Pro.local', ip: '', os.name: 'Mac OS X', os.arch: 'x86_64', os.version: '10.13.6', java.version: '10.0.2'
Driver info: driver.version: unknown
Error: Unknown timeout type in 'timeouts': 'type'
Build info: version: '3.12.0', revision: '7c6e0b3', time: '2018-05-08T15:15:08.936Z'
System info: host: 'Alexeys-MacBook-Pro.local', ip: '', os.name: 'Mac OS X', os.arch: 'x86_64', os.version: '10.13.6', java.version: '10.0.2'
Driver info: driver.version: unknown
    Emitted | test.after ([object Object])
    Emitted | suite.after ([object Object])

-- FAILURES:

  1) Suite 1
       "before each" hook: codeceptjs.before for "Check Sign In form":
     Unknown timeout type in 'timeouts': 'type'
Build info: version: '3.12.0', revision: '7c6e0b3', time: '2018-05-08T15:15:08.936Z'
System info: host: 'Alexeys-MacBook-Pro.local', ip: '', os.name: 'Mac OS X', os.arch: 'x86_64', os.version: '10.13.6', java.version: '10.0.2'
Driver info: driver.version: unknown
  rror: Unknown timeout type in 'timeouts': 'type'
[...]

PR #1288, opened 2018-10-29
fix timeouts definition for W3C compatible drivers for WebDriverIO

LukoyanovE:
Modern browsers starts to drop out old webdriver syntax.
For example Safari 12 and FF 63 does not support old syntax for timeout setting.

We should use new w3c compatible syntax in such cases
http://webdriver.io/api/protocol/timeouts.html


DavertMik:
Thanks!

LukoyanovE:
Fix for #1241

Issue #1272, opened 2018-10-21
Test w/multiple async scenarios & callback function throws error 


adams0619:
#### What are you trying to achieve?

I've created a test file with 2 `Scenarios` with one test using `Yahoo.com` & the other test using `Google.com`. I am using an internally hosted package (`aat`) to validate a11y & am using `async / await` syntax to wait for callback function from my a11y test to return before passing / failing a test. However, I have noticed that only one test runs, the other test never runs and fails immediately. The error I am getting points to an internal Codeceptjs method where a test's arguments are extracted. Please help me determine where I might be going wrong in my implementation?

#### What do you get instead?
```bash
CodeceptJS v1.4.3
Using test root "/Users/${home_dir}/test-ccjs"
Helpers: Nightmare
Plugins: screenshotOnFail
test --
   [1] Starting recording promises
   Emitted | suite.before ([object Object])
› [Session] Starting singleton browser session
  yahoo first @callback
   Emitted | test.before ([object Object])
   Emitted | test.start ([object Object])
   Emitted | step.before (I am on page "http://yahoo.com")
   Emitted | step.after (I am on page "http://yahoo.com")
   Emitted | step.before (I fill field "p", "github nightmare")
   Emitted | step.after (I fill field "p", "github nightmare")
   Emitted | step.before (I click "#uh-search-button")
   Emitted | step.after (I click "#uh-search-button")
   Emitted | step.before (I wait for element "#main")
   Emitted | step.after (I wait for element "#main")
   Emitted | step.before (I see element "#main .searchCenterMiddle li a")
   Emitted | step.after (I see element "#main .searchCenterMiddle li a")
   Emitted | step.before (I see element "//a[contains(@href,'github.com/segmentio/nightmare')]")
   Emitted | step.after (I see element "//a[contains(@href,'github.com/segmentio/nightmare')]")
   Emitted | step.before (I see "segmentio/nightmare", "#main li a")
   Emitted | step.after (I see "segmentio/nightmare", "#main li a")
   Emitted | step.before (I execute script () => {
        return document.body.innerHTML;
    })
   Emitted | step.after (I execute script () => {
        return document.body.innerHTML;
    })
   Emitted | step.start (I am on page "http://yahoo.com")
    I am on page "http://yahoo.com"
    › log: DARLA notice: 405
    › log: DARLA notice: 405
    › warn: yui: NOT loaded: highlander-vertical-video
[...]

PR #1290, opened 2018-10-29
Add check to filter out inherited array properties inside getInjectedArguments

adams0619:
When `codeceptjs` breaks down the user written `Scenarios` it checks for Injected arguments inside each `Scenario`. When `Scenarios` are written to be `async` and contain a custom helper which returns promise this functionality breaks down and throws and error when testing 2 or more `Scenarios`. The `for...in` loop from `getInjectedArguments` adds additional properties which are part of the array prototype causing the error to be thrown.

This PR add a simple check within the `for...in` to ensure that current properties inside the loop is a valid property within the `params` array and not an inherited property

See Issue #1272 for more detailed information regarding the error.  

[`for...in` MDN article](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#Array_iteration_and_for...in) for reference

### DIFF
--- a/lib/scenario.js
+++ b/lib/scenario.js
@@ -233,15 +233,17 @@ const getInjectedArguments = (fn, test) => {
   const params = getParamNames(fn) || [];
   const objects = container.support();
   for (const key in params) {
-    const obj = params[key];
-    if (test && test.inject && test.inject[obj]) {
-      testArguments.push(test.inject[obj]);
-      continue;
+    if (Object.prototype.hasOwnProperty.call(params, key)) {
+      const obj = params[key];
+      if (test && test.inject && test.inject[obj]) {
+        testArguments.push(test.inject[obj]);
+        continue;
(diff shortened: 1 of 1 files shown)


lib/scenario.js
```diff
@@ -233,15 +233,17 @@ const getInjectedArguments = (fn, test) => {
...
   const objects = container.support();
   for (const key in params) {
-    const obj = params[key];
-    if (test && test.inject && test.inject[obj]) {
-      testArguments.push(test.inject[obj]);
-      continue;
-    }
-    if (!objects[obj]) {
-      throw new Error(`Object of type ${obj} is not defined in container`);
+    if (Object.prototype.hasOwnProperty.call(params, key)) {
```
DavertMik: Can this be simplified with? 

```js
if (!params[key]) continue;
```

PR #1294, opened 2018-10-30
Feature. Add require param to config

LukoyanovE:
**Feature added:**
Added `require` param to require modules before codecept running:

For example, for Typescript and Should.js using in tests:
```
{
  "tests": './*_test.js',
  "timeout": 10000,
  ...
  // require modules
  "require": ["ts-node/register", "should"]
}
```

**Please, give some ideas for improvements.**
Or give other solution, which made this PR useless.

### DIFF
--- a/bin/codecept.js
+++ b/bin/codecept.js
@@ -4,8 +4,6 @@ const path = require('path');
 const Config = require('../lib/config');
 const Codecept = require('../lib/codecept');
 const { print, error } = require('../lib/output');
-const fileExists = require('../lib/utils').fileExists;
-const fs = require('fs');
 
 if (process.versions.node && process.versions.node.split('.') && process.versions.node.split('.')[0] < 8) {
   error('NodeJS >= 8 is required to run.');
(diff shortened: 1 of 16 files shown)


LukoyanovE:
Also, I copied part of mocha [docs](https://mochajs.org/#-r---require-module-name). Maybe, it should be rewrote to avoid copy-pasting

DavertMik:
> Requires described module before run. This option is useful for assertion libraries, so you may `--require should` instead of manually invoking `require('should')` within each test file.

Ok, this makes a little sense for an optional parameter.
If you don't specify require - all your tests fail. Also you need to keep in mind all required modules before executing tests. If there will be more than 1 it would be hard to type them in.

I think this `require` section should be moved to config file, so it would be always set for all runs

LukoyanovE:
> If you don't specify require - all your tests fail. Also you need to keep in mind all required modules before executing tests. If there will be more than 1 it would be hard to type them in.

Test does not fail without `--require` param.
I wrote tests for this situation. And there are no error/failing without it.
Will check again.

DavertMik:
That's very cool
Yey!

PR #1303, opened 2018-11-01
Fixed improper attached screenshot with unique name

PeterNgTr:
This to fix the case when uniqueScreenshotNames is true, on the html mocha report shows broken image using this helper
`"helpers": {
    "Mochawesome": {
        "uniqueScreenshotNames": "true"
    }
  },`


DavertMik:
Thanks

PR #1308, opened 2018-11-02
Doc: fix mustache files to improve docs

LukoyanovE:
Edited mustache files, and helpers files to uniform docs and arguments.
Improved params descriptions.

For some reasons after mustache files include "Appium: ..." on similar phrases were concatenated to last string from these files.
One possible solution - remove line breaks before such "appium" strings.
Does not know why, but works.

Maybe, there are other solution for this problem fix.


DavertMik:
:+1: :+1: :+1:

Issue #1315, opened 2018-11-07
cant run run-multiple parallel with ---profile


vitaly87:
#### What are you trying to achieve?
trying to call run-multiple parallel with ---profile
#### What do you get instead?
looks like when running it parallel the codeceptjs ignore this param

* CodeceptJS version:
codeceptjs@1.4.5

LukoyanovE:
Could you share your config?

vitaly87:
the issue that I want to be able to pass parameters on start with flag profile it is working on codeceotjs run and should also work on parallel. what I am passing is user and pass.

pablopaul:
My current workaround here is to avoid usage of `--profile` and instead use a separate config file with `-c configs/…`

pablopaul:
Should be fixed with newest release 2.6 - use process.env.profile instead of process.profile.

jltrippett:
I am using codeceptjs 2.6.6 and still seeing an issue with this. It looks like one codeceptjs instance is using the profile, but the second is undefined. 

Test run output:
` npx codeceptjs run-multiple parallel --profile dev
{
  // printing our config urls
}
profile dev // console.log(process.env.profile)
{
 // printing our config urls
}
profile undefined // console.log(process.env.profile)`

joshuacoveauth0:
I am using codeptjs 2.6.7 and process.env.profile is not parsed even when using `chunks: 1`

paulcredmond:
Looks like `--profile` is not passed to child processes using `run-multiple`. 

For example, I want to pass a test URL as an argument: `--profile env="https://test.com"`. I have a function to retrieve this by looking at the `process.argv` and returning when it matches. @pablopaul any suggestions on how to get this working?

paulcredmond:
I made a workaround function to get the config passed to the child processes, parse it and get the URL from there.

idxn:
@paulcredmond can you please share us that function? It would benefit others.

paulcredmond:
Sure @idxn 

This very much tailored for my needs which is to pass a string (URL) for Webdriver to use as the URL to be tested.

So it would look like: `npm run e2e --profile env=https://URL.com`

```
// Profile is passed to the first process and the config is passed to child processes, but --profile is not
// This function will search for the config in the process.argv, parse the JSON and find the URL

getEnvObject: (argsArray) => {
    let url = null;
    const env = argsArray.filter((el) => el.includes('WebDriver'));
    const obj = env[0] || null;
[...]

Issue #1311, opened 2018-11-05
Running codeceptjs with multiple browsers


omeriko:
#### What are you trying to achieve?

I'm running this command:
codeceptjs run-multiple parallel --profile "oracle,Change123" --grep "@Hosted|@PrivateSqlServer" --invert 

#### What do you get instead?

I'm getting the following error message

> Provide console output if related. Use `--verbose` mode for more details.

error: unknown option `--invert' 

```bash
# paste output here
```

> Provide test source code if related

```js
// paste test
```

### Details

* CodeceptJS version:^1.4.3
* NodeJS Version: 8.12.0
* Operating System: windows 10 enterprise
* Protractor || WebDriverIO || Nightmare version (if related):  WebDriverIO
* Configuration file: codecept.conf.js

```js
# paste suite config here
```

exports.config = {
  multiple:{
    parallel: {
      chunks: 2
    }
  },
  output: "./output",
  helpers: {
    WebDriverIO: {
      url: "https://ids.ics.dev/identity-service/home",
      browser: "chrome",
      smartWait: 5000,
      windowSize: "maximize"

    },
    cihHelper: {
      require: "./helpers/cih_helper.js"
    },    
    RestHelper: {
      require: "./helpers/restHelper.js"
    }
  },
  multiple: {
    parallel: {
      // Splits tests into 2 chunks
      chunks: 2
    }
  },
  plugins: {
    stepByStepReport: {
      enabled: true,
      deleteSuccessful:true
    },
    allure: {
      enabled: true
    }
  },
  include: {
    I: "./logicalOperations.js"
  },
  mocha: {},
  bootstrap: false,
  teardown: null,
  hooks: [],
  gherkin: {},

  tests: "./tests/OnlineSubscription/*.js",

  timeout: 10000,
  name: "build-assets",
  auth: {}
};
// we use 1 parameter to update config
if (process.profile) {
  let params = process.profile.split(",")
  // update config for cloud service provider
  exports.config.auth.user = params[0];
  exports.config.auth.pass = params[1];
} else {

  exports.config.auth.user = 'hosted1@infa.com';
  exports.config.auth.pass = '******';
}

EgorBodnar:
Fixed in #1322

Issue #1310, opened 2018-11-05
cant run run-multiple parallel with --invert


vitaly87:
#### What are you trying to achieve?
Hi,
I am tring to run codeceptjs with params, its looks like  run-multiple parallel with --invert is getting an error on a console 
#### What do you get instead?
error: unknown option `--invert'
> Provide console output if related. Use `--verbose` mode for more details.

```bash
# paste output here
```

> Provide test source code if related

```js
// paste test
```

### Details

* CodeceptJS version:
* NodeJS Version:
* Operating System:
* Protractor || WebDriverIO || Nightmare version (if related)
* Configuration file:

```js
# paste suite config here
```

PR #1322, opened 2018-11-09
Fix - add invert option to run-multiple

LukoyanovE:
Fix for #1310 #1311

`--invert` option was not passed through `run-multiple` command.
Tests were added

Try to research #1315 with `--profile` option. Did not found relation with it.


DavertMik:
Super!

PR #1325, opened 2018-11-12
Add option `output` to stepByStepReport plugin

fpsthirty:
add option `output` to stepByStepReport plugin for customize output-directory.
Example:

```
  "plugins": {
    "stepByStepReport": {
      "enabled": true,
      "output": "./output/report"
    }
 }
```


DavertMik:
Thanks!

PR #1307, opened 2018-11-02
Fix/allure report no parent steps

PeterNgTr:
There is a case when you have Before/BeforeSuite/AfterSuite/After with allure report enabled

```
Feature('Login functionality');

BeforeSuite( (I) => {
    I.amOnPage('/login-old');
})

AfterSuite( (I) => {
    I.amOnPage('/login-old');
})

Scenario('See login page', (I) => {
    I.dontSee('Welcome', 'span');
});
```

you will see something like this

> allure-js-commons: Unexpected startStep() of Context: BeforeSuite . There is no parent step

```
Login functionality --
    [1] Starting recording promises
    Emitted | suite.before ([object Object])
 › [Session] Starting singleton browser session
    Emitted | hook.start ([object Object])
    Emitted | step.before (I am on page "/login-old")
    Emitted | step.after (I am on page "/login-old")
    Emitted | step.start (I am on page "/login-old")
allure-js-commons: Unexpected startStep() of Context: BeforeSuite . There is no parent step
allure-js-commons: Unexpected startStep() of I am on page "/login-old". There is no parent step
    Context: BeforeSuite
      I am on page "/login-old"
      › [Url] https://app-develop.mcmakler.com/login-old
    Emitted | step.passed (I am on page "/login-old")
allure-js-commons: Unexpected endStep(). There are no any steps running
    Emitted | step.finish (I am on page "/login-old")
    Emitted | hook.passed ([object Object])
  See login page
    Emitted | test.before ([object Object])
    Emitted | test.start ([object Object])
    Emitted | step.before (I dont see "Welcome", "span")
    Emitted | step.after (I dont see "Welcome", "span")
    Emitted | step.start (I dont see "Welcome", "span")
allure-js-commons: Unexpected endStep(). There are no any steps running
    I dont see "Welcome", "span"
    [1] Error | Error
```

PR #1326, opened 2018-11-13
Fix/no steps in allure report

PeterNgTr:
After this changes https://github.com/Codeception/CodeceptJS/pull/1307 there is no step in allure report anymore in CodeceptJS 1.4.5 so technically this patches that issue.


DavertMik:
You broke it, you fixed it. Fair enough!
Thank you )
:+1:

PR #1327, opened 2018-11-13
Feature/rest helper with axios

PeterNgTr:
Moving to Axios to avoid the receiving a security alert at
https://github.com/uc-cdis/gen3-qa/network/alerts

The reason is codecept's rest uses unirest which are not updated for more than one year and the latest version uses outdated libraries including `mime` and `request` (that uses `cryptiles` and `hoek`
https://github.com/Kong/unirest-nodejs/blob/master/package.json


DavertMik:
Thanks! 
I think we need to release this into 1.5.0 
Also, could you list breaking changes in this release?
Maybe documentation should be updated too?

PeterNgTr:
WIP - Updating doc and unit tests

PeterNgTr:
Updated REST helper with axios:
- updated docs
- updated REST unit tests
- removed `resetHeaders` and `followRedirect` cause I'm not really sure if we still need this.

DavertMik:
Let's go with this!

PR #1334, opened 2018-11-17
Changed type definition of PageObjects to get auto completion

rhicu:
In order to get auto completion in test files for page objects the type of the page object must not be any in type definitions.


DavertMik:
Thanks

Issue #1266, opened 2018-10-18
Screenshot on fail does not include additional sessions


GerroDen:
#### I only get a screenshot of the main session and none of the additional.

> Feature with failing test inside a session:
```js
Feature("Screenshots on fail in session");

Scenario("Failing in session", (I) => {
	session("new session", () => {
		I.amOnPage('http://google.com');
		I.dontSeeInCurrentUrl('http://google.com');
	});
});
```

### Details

* CodeceptJS version: 1.4.3
* NodeJS Version: v10.12.0
* Operating System: Ubuntu 16.04
* Puppeteer: 1.9.0

koushikmohan1996:
Fix is merged

PR #1335, opened 2018-11-20
Fixed/screenshot on additional sessions

PeterNgTr:
This fixes issue that screenshot on failed cases only with first session not the current one. Now, it will take screenshot of the current one. https://github.com/Codeception/CodeceptJS/issues/1266


DavertMik:
Thanks

PR #1350, opened 2018-11-22
Add support to use WSEndpoint on Puppeteer

gabrielcaires:
Hello,
With this PR will be possible to use remote chrome instances like https://www.browserless.io/ or docker containers like https://github.com/joelgriffith/browserless.

I saw that currently the project is not mocking Puppeteer and we don't have tests to `_startBrowser`, so I don't know how test it. Should I mock Puppeteer?

### DIFF
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -59,7 +59,8 @@ const consoleLogStore = new Console();
  *
  * ```js
  * "chrome": {
- *   "executablePath" : "/path/to/Chrome"
+ *   "executablePath" : "/path/to/Chrome",
+ *   "browserWSEndpoint": "ws://localhost:3000"
  * }
  * ```
  *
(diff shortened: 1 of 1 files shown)


lib/helper/Puppeteer.js
```diff
@@ -59,7 +59,8 @@ const consoleLogStore = new Console();
  *
  * ```js
  * "chrome": {
- *   "executablePath" : "/path/to/Chrome"
+ *   "executablePath" : "/path/to/Chrome",
+ *   "browserWSEndpoint": "ws://localhost:3000"
```
DavertMik: thanks. Can you add here an example here for using CodeceptJS with browserless.io?

DavertMik:
I think for such case we are fine without a test

DavertMik:
Ok, I will update docs on my own

Issue #1342, opened 2018-11-21
CSS object not expanded in the mochawesome report


xMutaGenx:
#### What are you trying to achieve?

See the expanded (recommended way to use) css element when a test fails

#### What do you get instead?

`Error: element ([object Object]) still visible after 10 sec`

```
Scenario('Test', (I) => {
    I.amOnPage('https://www.google.com');
    I.waitForInvisible({css: 'input[name="q"]'});
});
```

### Details

* CodeceptJS version: 1.4.4
* NodeJS Version: 10.13.0
* Operating System: OSX
* WebDriverIO: 4.14.0

PeterNgTr:
![screenshot from 2018-11-21 15-09-50](https://user-images.githubusercontent.com/7845001/48846480-8878c800-ed9f-11e8-8917-70c735f1c59c.png)
I didn't get the same

xMutaGenx:
@PeterNgTr ran this with puppeteer, it works there, seems to be related to wdio

PeterNgTr:
The fix is on the way https://github.com/Codeception/CodeceptJS/pull/1351

PR #1351, opened 2018-11-23
Enh/parse object error wio

PeterNgTr:
https://github.com/Codeception/CodeceptJS/issues/1342

This fixes the object locator is not parsed when showing in report/cli.


DavertMik:
Cool!

Issue #1317, opened 2018-11-07
v1.4.5 screenshots are no longer taken when we run-multiple


xMutaGenx:
#### What are you trying to achieve?

Framework takes a screenshot after failure on v1.4.5, works fine on v1.4.4

#### What do you get instead?

no screenshot is taken

### Details

* CodeceptJS version: 1.4.5
* NodeJS Version: >=8.9.1
* Operating System: OSX
* WebDriverIO: 4.14.0
* Configuration file:

```js
helpers: {
        WebDriverIO: {
          ...
          fullPageScreenshots: true,
          uniqueScreenshotNames: true,
          ...
       },
        Mochawesome: { uniqueScreenshotNames: true }
    }
```

PeterNgTr:
<img width="905" alt="screen shot 2018-11-13 at 11 59 54 am" src="https://user-images.githubusercontent.com/7845001/48409304-c17bc180-e73b-11e8-9737-a90bef74fd94.png">
Cannot reproduce this. It worked for me

xMutaGenx:
Hi @PeterNgTr the screenshot gets generated, it does not get attached to the report, so if you open up the mochawesome.html it won't be there, also, we use a specific output directory, with `run-multiple`, rather than just `run`

PeterNgTr:
![screenshot from 2018-11-19 15-11-23](https://user-images.githubusercontent.com/7845001/48712178-694a3100-ec0d-11e8-84d6-4ecba773d36f.png)
I can see the attached image for failed case.

PeterNgTr:
![screenshot from 2018-11-19 15-14-05](https://user-images.githubusercontent.com/7845001/48712332-c9d96e00-ec0d-11e8-8b9e-73cf201d74dc.png)
`run-multiple` also worked for me

xMutaGenx:
@PeterNgTr just to confirm before I go digging through our stack, which version of grid/hub are you using?

PeterNgTr:
I'm running against `puppeteer`

xMutaGenx:
Ah, we've got a grid/hub setup of v3.141.5, will take a look if that is causing it as well, thanks for the info though!

yogainalift:
I can also confirm that for 1.4.6 the screenshots are not even generated and not attached to the form. Reverted back to 1.4.4 for now.

mirao:
[...]

PR #1354, opened 2018-11-28
Fix/no mocha screenshot with wio

PeterNgTr:
There is a report that screenshot is not generated with wio. https://github.com/Codeception/CodeceptJS/issues/1317

This fixed that issue.


DavertMik:
Thanks!

PR #1357, opened 2018-11-29
Feat/test rail integration

PeterNgTr:
I think people are currently interested in how we could send test results to testrail once the tests are executed by codeceptjs (https://codecept.discourse.group/t/testrail-plugin/36). This one will basically fulfill this. 

Test run will be created and push test results to test cases. Currently, there are just following statuses
- passed
- failed

Other statuses are not currently supporting by codeceptjs events.


DavertMik:
Yeah, I have such concern as well. 
TestRail is cool but it adds a dependency and I don't know if it's a good idea to take it into the core. I think it should be a standalone package and it can be listed somewhere on a site or in our discourse group. Anyway, we need a better home for 3rd party plugins.

DavertMik:
@PeterNgTr 
Please create a standalone package. Add it to the list of community helpers https://github.com/Codeception/CodeceptJS/wiki/Community-Helpers

On the new website there is a link to that page so everyone can search for community helpers. 
So if you add it there it won't be lost :smiley:

PeterNgTr:
Thanks, I’ll do it

PeterNgTr:
Added to suggested page :)

PR #1360, opened 2018-11-30
Update Node version to latest lts

dustinleblanc:


LukoyanovE:
I'm glad to see, that you track Node and Codecept, but I have doubt about this PR.

Don't think it's good idea.
I use (and, I think, a lot of programmers) 8th node in projects.
If this PR is accepted, I could not use Codecept in my project.

It's breaking change, without any improvements.
Do you really need to update minimal node version?

dustinleblanc:
Also helps if I add the right version :laughing: 

@LukoyanovE after I made this PR I also discovered you can do some crazy stuff with rebuilding the image yourself and passing build args to change the node version. For me, all of our projects track the latest LTS of Node which swapped over to 10.x series a little bit ago. As far as improvements, you'll want to review the NodeJS change log to keep up with them, not to mention staying on the current LTS means better stability long term.

DavertMik:
I think CodeceptJS 1.5 is the right version to update Docker image to new LTS

Issue #1371, opened 2018-12-07
mochawesome screenshots not saving if report directory is changed.


null-saad:
#### What are you trying to achieve?
I changed the report directory when running codeceptjs with mochawesome report.

#### What do you get instead?
The failed screenshots get saved in the "output" default report directory and not the in the new directory and hence the report is unable to access is when served.

> Provide console output if related. Use `--verbose` mode for more details.

```bash
# paste output here
```
There is no error on console.

> Provide test source code if related
codeceptjs run --reporter mochawesome --reporter-options reportDir='./myreport' sample_test.js

```js
// paste test
```

### Details

* CodeceptJS version: 1.4.6
* NodeJS Version: 10.9.0
* Operating System: Windows 10
* Protractor || WebDriverIO || Nightmare version (if related) Pupeeteer
* Configuration file:
{
  "output": "./output",
  "helpers": {
    "Mochawesome": {
      "uniqueScreenshotNames": "true"
  },
    "Puppeteer": {
      "url": "http://localhost",
      "show": "show",
      "restart": "false",
      "waitForNavigation": "networkidle0",
      "waitForTimeout": "60000",
      "windowSize" : "640x480"
    },
    "Rhodium": {
      "require": "./rhodium_helper.js"
    }
  },
  "include": {
    "I": "./steps_file.js"
  },
  "mocha": {
    "reporterOptions": {
      "reportDir": "output"
    }
  },
  "bootstrap": false,
  "teardown": null,
  "hooks": [],
  "gherkin": {},
  "tests": "./*_test.js",
  "timeout": 10000,
  "name": "CC"
}

```js
# paste suite config here
```

PeterNgTr:
The fix is coming here https://github.com/Codeception/CodeceptJS/pull/1372

PR #1372, opened 2018-12-07
Enh/custom mocha report dir for screenshot

PeterNgTr:
I changed the report directory when running codeceptjs with mochawesome report.

**What do you get instead?**
The failed screenshots get saved in the "output" default report directory and not the in the new directory and hence the report is unable to access is when served.

This fixed https://github.com/Codeception/CodeceptJS/issues/1371


lib/helper/Puppeteer.js
```diff
@@ -1335,7 +1335,14 @@ class Puppeteer extends Helper {
    */
   async saveScreenshot(fileName, fullPage) {
     const fullPageOption = fullPage || this.options.fullPageScreenshots;
-    const outputFile = path.join(global.output_dir, fileName);
+    let outputFile;
+
+    if (!fileName.includes('/')) {
```
DavertMik: copy-paste detected :)
let's move it somewhere to `utils`

PeterNgTr: moved it to `utils`

DavertMik:
:+1:

lib/utils.js
```diff
@@ -268,3 +268,10 @@ module.exports.deleteDir = function (dir_path) {
     fs.rmdirSync(dir_path);
   }
 };
+
+module.exports.screenshotOutputFolder = function (fileName) {
+  if (!fileName.includes('/')) {
```
trinhpham: This causes problem on Windows.
Why didn't you use `path.sep`?

trinhpham:
Found that screenshot won't be saved anymore on Windows because of this commit.
I've added comment to the bad code line.
@DavertMik + @PeterNgTr please help review again.

PeterNgTr:
sure @trinhpham which line is it?

lib/utils.js
PeterNgTr: I missed it. will fix

trinhpham: Great, @PeterNgTr !
When could we have a new build for this fix?
I've just upgraded CodeceptJS version today and going to downgrade it now 😢

PeterNgTr: https://github.com/Codeception/CodeceptJS/pull/1395
I fixed it here, can you try it on windows. I don't have windows machine :)

new build push should depend on @DavertMik

trinhpham: Glad to know version 2.0.1 has your fix.
Thanks 😄

Issue #1228, opened 2018-09-25
Allure report doesn't include whole feature if at least one of its scenarios contain session


alfiakh:
#### Allure report doesn't include feature with all containing scenarios if one of them 

Here is an repository containing example of in which cases allure report doesn't work properly https://github.com/alfiakh/allure-session-sample

> Feature with session items

```js
Feature('Feature with session');

Scenario('Scenario without session', (I) => {
  I.amOnPage('http://google.com');
})

Scenario('Scenario with session', (I) => {
  I.amOnPage('http://google.com');
  session('second', () => {
    I.amOnPage('http://google.com');
  })
})

```

> Feature without session items

```js
Feature('Feature without session');

Scenario('Scenario without session', (I) => {
  I.amOnPage('http://google.com');
})

```

> Console execution

```bash
➜  allure-session-sample git:(master) npm test

> allure-session-sample@1.0.0 test /Users/alf/repos/allure-session-sample
> codeceptjs run --plugins allure

CodeceptJS v1.4.2
Using test root "/Users/alf/repos/allure-session-sample"

Feature with session --
  ✔ Scenario without session in 1177ms
  ✔ Scenario with session in 2361ms
Feature without session --
  ✔ Scenario without session in 1115ms

  OK  | 3 passed   // 6s
```

### Details

* CodeceptJS version: 1.4.2
* NodeJS Version: v8.12.0
* Operating System: mac OS 10.13.6
* Protractor || WebDriverIO || Nightmare version (if related)
* Configuration file: listed in repository

### Actual result
Allure report lists only feature without scenarios containing session
![joxi_screenshot_1537880871331](https://user-images.githubusercontent.com/8930318/46016371-44d64a80-c0dd-11e8-87ea-53f0ce66123a.png)

### Steps to reproduce:
1. Clone repo
2. ```npm i```
3. ```npm test```
4. ```npm report```

PeterNgTr:
Fixed here https://github.com/Codeception/CodeceptJS/pull/1375

![screenshot from 2018-12-10 12-40-26](https://user-images.githubusercontent.com/7845001/49731034-5958c680-fc7a-11e8-9f46-07cac14d8d2f.png)

PR #1375, opened 2018-12-10
Fix/no allure report when using session

PeterNgTr:
There is a case when using session in scenarios, allure report is not generated. -> https://github.com/Codeception/CodeceptJS/issues/1228


lib/plugin/allure.js
```diff
@@ -104,6 +104,9 @@ module.exports = (config) => {
     if (isHookSteps === false) {
       startMetaStep(step.metaStep);
       if (currentStep !== step) {
+        if (step.actor.includes('\u001b[36m')) {
```
DavertMik: Could you write a line in comment about this tricky hook
We don't have allure tests so at least we should have a comment on this block
Why do you do this and why is that so

PeterNgTr: Updated comment to reflect why we should do this

DavertMik:
:+1:

Issue #1103, opened 2018-06-07
I.say() - user can't change a color text 


ko4ergins:
#### What are you trying to achieve?
The user can edit text color in a console when doing I.say();
#### What do you get instead?
The user can write tests in BDD style with different colors - steps.
Example:
I.say("Given: admin on the 'Home page'", red):
Step1();
Step2();
I.say("When: I do bla-bla", green);
Step(3);
I.say("Then: I did it", purple);

> Provide console output if related. Use `--verbose` mode for more details.

```bash 
N/A
```

> Provide test source code if related

```js
N/A
```

### Details

* CodeceptJS : v1.2.1
* NodeJS Version : v8.11.2
* Operating System: Linux Mint x64
* Protractor || WebDriverIO || Nightmare version (if related)
* Configuration file:

```js
N/A
```

PR #1377, opened 2018-12-12
Enh/support color for say method

PeterNgTr:
We now can give the second param to define which color you want to print out with `I.say()` method https://github.com/Codeception/CodeceptJS/issues/1103

For instance:
```
I.say('This is red', 'red'); //red is used
I.say('This is blue', 'blue'); //blue is used
I.say('This is by default'); //cyan is used
```


DavertMik:
Nice. Could you update documentation for this method?

PeterNgTr:
Updated doc!

DavertMik:
Thanks!

Issue #1208, opened 2018-09-14
Switching to codecept.conf.js by default


DavertMik:
We support `.js` and `.json` config formats. Originally JSON was chosen because of its simplicity: it's easy to write and easy to update. For instance, it was easy for generators (like page objects, or page fragments) to include the newly created objects into a config automatically.

However, `.json` does not provide the flexibility needed to build a complex system. That's why at some point we started to [support `.conf.js` config](https://codecept.io/configuration/#dynamic-configuration).

The next idea is to switch from `codecept.json` in favor of `codecept.conf.js` for all new setups and provide a migration script (which should be very simple) to do the switch.

Nazanin1369:
I agree. Using .js gives more flexibility building dynamic configurations.

PR #1378, opened 2018-12-12
Enh/support js conf by default

PeterNgTr:
We support .js and .json config formats. Originally JSON was chosen because of its simplicity: it's easy to write and easy to update. For instance, it was easy for generators (like page objects, or page fragments) to include the newly created objects into a config automatically.

However, .json does not provide the flexibility needed to build a complex system. That's why at some point we started to support .conf.js config.

The next idea is to switch from codecept.json in favor of codecept.conf.js for all new setups and provide a migration script (which should be very simple) to do the switch.

As @DavertMik described here https://github.com/Codeception/CodeceptJS/issues/1208


DavertMik:
Switching from codecept.json and removing its support should come to 2.0 version

PR #1379, opened 2018-12-14
Enh/add hook to perform pre checks

PeterNgTr:
So, I think it would be a good practice to do some pre checks before we commit or push something. Contributors are growing, I guess and we want to make it kinda easy to keep calm and contribute. Then pre checks enter the story.

The following pre checks will be performed:
- when you commit: `npm run docs && npm run lint-fix` will be performed to update docs and lint fix if any
- when you push: `npm test` will be performed to make sure we won't break anything ;)

Hope this helps. **Keep calm and contribute** :100:


DavertMik:
Makes sense. Thank you!
Keep calm and contribute :100:  - is awesome motto

PR #1381, opened 2018-12-18
Fix syntax of the generate commands

textbook:
A colon is required between `generate` and the type of object to create


DavertMik:
Right! Thanks

Issue #244, opened 2016-10-06
A good way to get server response code?


SuneRadich:
I have just started using CodeceptJS and so far I am quite impressed. One of the things I want to test against, is a bunch of core URLs, and see if their response code matches what I expect.

I am using Nightmare and can see that Nightmares .goto returns the response code I want. However, I cant figure out how to get that data when using the regular I.amOnPage (that in the background uses .goto). So my current solution is to use a custom helper function, to perform the .goto() and then store result on the browser object.

I then have another custom function that tests against that result on the browser object.

But somehow this seems to be the wrong of doing it, by adding data to the browser object. Is there a built in way to get the server response code, or is tests like this not within the "area" of CodeceptJS ?

My helper functions can be seen below

```
/**
    * Go to a given page, and store the response on the browser object for later reference
    */
    customAmOnPage(url, headers) {

        // let's use chai assertion library
        let expect = require('chai').expect;

        // get nightmare instance
        let browser = this.helpers['Nightmare'].browser;

        if (url.indexOf('http') !== 0) {
            url = browser.options.url + url;
        }

        return browser.goto(url, headers).then((res) => {
            //Store results on the current browser object, giving us access to response code etc.
            browser.currentPageResponse = res;
        });
    }

    getServerResponseCode(expectedValue) {

        // let's use chai assertion library
        let expect = require('chai').expect;

        // get nightmare instance
        let browser = this.helpers['Nightmare'].browser;

        expect(browser.currentPageResponse.code).to.equal(expectedValue);
    }
```

DavertMik:
Sorry, I'd love to answer this question but unfortunately I don't have time to try this in action. It seems to me you are close to get the answer by yourself, I think you just need to get a bit more into Nightmare helper and corresponding helper. 

I think CodeceptJS should have method like `seeResponseCodeIs()`. If you will get an idea how to implement that it would be nice to see the Pull Request.

SuneRadich:
My solution works for me, but it seems to be the wrong approach to make a custom amOnPage test. I'll see if I have time to look into the code and make a PR.

PeterNgTr:
make a PR here https://github.com/Codeception/CodeceptJS/pull/1383

PeterNgTr:
no activities on this. Closed it for now.

PR #1383, opened 2018-12-21
Feat/see response code

PeterNgTr:
Fulfilled https://github.com/Codeception/CodeceptJS/issues/244


PeterNgTr:
Updated code as grabbing the statusCode of current URL and make assertion against it.

PeterNgTr:
Fixed failed helper tests

DavertMik:
Sorry, still no. 

Here is the thing. When you send requests via 3rd party library (which is got) you are not inside the session of current browser so the results will be different from what you get in browser. 

How I think it should be implemented

* Nightmare has API for status codes: https://github.com/segmentio/nightmare/issues/656
* Puppeteer has its API too.
* WebDriver/Protractor doesn't have such API at all.

In Nightmare/Puppeteer you can just create a property for a status code and update it after each request. When you call `seeResponseCodeIs` it will compare expected value with the value of the last request.

PeterNgTr:
I‘m trying to implement this with your suggestions

PeterNgTr:
Updated function `seeResponseCodeIs`, right now, after doing a `amOnPage`, the `statusCode` will be updated and `seeResponseCodeIs` will make the assertion against this `statusCode`

lib/helper/WebDriverIO.js
```diff
@@ -773,6 +774,14 @@ class WebDriverIO extends Helper {
     return stringIncludes('web page title').assert(text, title);
   }
 
+  /**
+   * {{> ../webapi/seeResponseCodeIs }}
+   */
+  async seeResponseCodeIs(expectedStatusCode) {
```
DavertMik: WebDriver can't return status codes

lib/helper/WebDriver.js
```diff
@@ -818,6 +819,14 @@ class WebDriver extends Helper {
     return stringIncludes('web page title').assert(text, title);
   }
 
+  /**
+   * {{> ../webapi/seeResponseCodeIs }}
+   */
+  async seeResponseCodeIs(expectedStatusCode) {
+    const statusCode = await getStatusCode(await this.grabCurrentUrl());
```
DavertMik: WebDriver can't return status codes.
Please remove this method from WebDriver

lib/helper/Puppeteer.js
```diff
@@ -466,10 +467,18 @@ class Puppeteer extends Helper {
...
+    const response = await this.page.goto(url, { waitUntil: this.options.waitForNavigation });
+    statusCode = response.status();
     return this._waitForAction();
   }
 
+  /**
+   * {{> ../webapi/seeResponseCodeIs }}
+   */
+  async seeResponseCodeIs(expectedStatusCode) {
+    return statusCode === expectedStatusCode;
```
DavertMik: this should be an assertion

lib/helper/Nightmare.js
```diff
@@ -412,9 +413,17 @@ class Nightmare extends Helper {
...
       this.debugSection('Headers', JSON.stringify(res.headers));
+      statusCode = res.code;
     });
   }
 
+  /**
+   * {{> ../webapi/seeResponseCodeIs }}
+   */
+  async seeResponseCodeIs(expectedStatusCode) {
+    return statusCode === expectedStatusCode;
```
DavertMik: should be assertion!

lib/helper/WebDriverIO.js
```diff
@@ -13,6 +13,7 @@ const {
   chunkArray,
   convertCssPropertiesToCamelCase,
   screenshotOutputFolder,
+  getStatusCode,
```
DavertMik: what does this function do?
I don't see it in this PR

PeterNgTr:
Found another issue in test, the `statusCode` will not be set if we are not `await I.amOnPage()`

PeterNgTr:
I'm not sure if we really need this anymore. So closed this

Issue #1384, opened 2018-12-21
ENAMETOOLONG error with data driven tests


vitalykorolev:
#### What are you trying to achieve?
Trying to run data driven tests as per https://codecept.io/advanced/#data-driven-tests
The data consists of multiple columns of data including cookie value.

#### What do you get instead?
When one of the scenarios fails, the entire script fails with the following error:
Error: ENAMETOOLONG: name too long, open '/Users/.../output/Verify_presence_of_disable_option_for_a_user_-_Administrator___{ACC_uname_qatest10,ACC_fname_Org2,ACC_lname_Admin,ACC_email_qatest10@org2.com,ACC_country_Canada,ACC_password_SomePass#123,ACC_role1_SEC-ADMN,ACC_role2_SERV-ADMN,ACC_role3_ACCT-ADMN}.failed.png'

### Details

* CodeceptJS version: 1.4.6
* NodeJS Version: 6.4.1
* Operating System: MacOS
* Protractor

Looks like there were issues opened for this, but both were closed referencing each other.
https://github.com/Codeception/CodeceptJS/issues/1106
https://github.com/hubidu/bifrost-io/issues/2

PeterNgTr:
Fix is here https://github.com/Codeception/CodeceptJS/pull/1385

Issue #1106, opened 2018-06-12
Data-driven tests don't save screenshots on failure


elijah-schow:
#### What are you trying to achieve?

I want a screenshot to be saved when a data driven tests fails.

#### What do you get instead?

It looks like the screenshot cannot be saved because the file name is too long.

> Screenshot has been saved to /Users/elijahschow/Dev/prod/e2e/output/Validate_business_application_FOCUS___{'Test_Case_Step'_'If_Phone_Number_is_not_entered','Amount'_21500,'Business_Type'_'LLC','Equipment_Type'_'LED_Lighting_and_Signage','Equipment_Condition'_'New','First_Name'_'Any','Last_Name'_'Any','Email'_'123@test.com','Phone'_'','Fico_SSN'_'999,999,700','Address'_'Customer_Address','City'_'Minnesota','State'_'MN','Zip'_111,'Legal_Company_Name'_'LeaseQ','Address_(Company)'_'Address_of_Company','City(Company)'_'Minnesota','Company_State'_'MN','Zip(Company)'_1111,'Phone(Company)'_6036307427,'Time_in_Business'_'1_Year','Vehicle_Type'_'NA','Truck_Age_Restriction___for_Long_Haul_Vehicle_Type(Vehicle_Year_)'_'NA','Mileage_Restriction_for_Long_Haul_Vehicle_Type'_'NA','Minimum_Number_of_Trucks'_'NA','Minimum_Number_of_Years_With_CDL'_'NA','Annual_Revenue'_90000,'Ownership_Percentage'_100,'Pop_Up_message_Displayed'_'Please_complete_the_FOLLOWING_REQUIRED_FIELDS__Customer_Phone.','Tier'_'','12_Months'_'','24_Months'_'','36_Month'_'','48_Month'_'','60_Month'_''}.failed.png

```js
Data(TestCases).Scenario('Validate business application FOCUS', async (I) => {
   ...
});
```

### Details

* CodeceptJS version: 1.2.1
* NodeJS Version: v10.3.0
* Operating System: MacOS High Sierra 10.13.4
* WebDriverIO version: 4.12.0
* Configuration file:

```js
exports.config = {
  "tests": "./tests/*.[t|j]s",
  "timeout": 10000,
  "output": "./output",
  "helpers": {
    "WebDriverIO": {
      "url": process.env.TARGET || 'http://localhost:8080/',
      "browser": "chrome",
      "windowSize": "maximize",
      "coloredLogs": true,
[...]

PR #1385, opened 2018-12-22
Fix/enametoolong error with data driven tests

PeterNgTr:
There is such a case when we inject the much data driven to scenarios, once the test is failed, screenshot could not be created due to the file name is too long. 

This fixed that issue -> https://github.com/Codeception/CodeceptJS/issues/1384, https://github.com/Codeception/CodeceptJS/issues/1106, https://github.com/hubidu/bifrost-io/issues/2


PeterNgTr:
Updated code to cut data string when saving screenshot + added unit test

DavertMik:
Sorry I was not clear enough. 
I think screenshots should be generated for data tests and we should not remove the data parameter completely. In this case we could have a screenshot per failed data tests, which is expected behavior. 
So removing the data string is not a solution. 
Maybe it's better to take that data string, cut to some normal length, like 20-30 chars and use it in filename.

DavertMik:
Cool I will merge this to fix the bug

PR #1386, opened 2018-12-23
Webdriverio v5

DavertMik:
Upgrade to webdriverio v5 https://github.com/webdriverio/webdriverio/blob/master/CHANGELOG.md


.travis.yml
```diff
@@ -13,7 +13,7 @@ env:
   - HELPER=Nightmare
   - HELPER=Puppeteer
   - HELPER=ProtractorWeb
-  - HELPER=WebDriverIO
```
LukoyanovE: Why WebDriver, when it's called WebDriverIO?

DavertMik: Good point. 
The upgrade is huge so we will need to keep both helpers in repo.
I decided to name new one this way so we could distinguish them by name...

LukoyanovE: Good decision, 
Maybe, better than `WebDriverIO5`

DavertMik: I thought about this too. 
But it can be ugly for onboarding of new users. Why IO why 5 if they only know about WebDriver.

There is a point here. Since v5 webdriverio focuses on webdriver spec implementation, so it's pure webdriver client with less opinionated features. So I think it's ok to call it just Webdriver. At least users may not really care by which library and which version webdriver is called.

PR #1389, opened 2018-12-26
Fix/wdio5/no screenshot with failed test

PeterNgTr:
Fixed the issue screenshot is not generated when there is failed case.


DavertMik:
Thanks!

Issue #1392, opened 2018-12-27
Unable to initialize CodeceptJS in version 2.0.0


cuki:
#### What are you trying to achieve?
Initialize CodeceptJS in current project directory in order to create new configuration.

#### What do you get instead?
Nothing.

Upon after answering the questions there is no **codecept.conf.js** and/or other defined files.
```bash
$ codeceptjs init

  Welcome to CodeceptJS initialization tool
  It will prepare and configure a test environment for you

Installing to /project/tests
? Where are your tests located? ./*_test.js
? What helpers do you want to use? Puppeteer
? Where should logs, screenshots, and reports to be stored? ./output
? Would you like to extend I object with custom steps? Yes
? Do you want to choose localization for tests? English (no localization)
? Where would you like to place custom steps? ./steps_file.js

$ ls
<nothing>
```

### Details
* CodeceptJS version: 2.0.0
* NodeJS Version: 10.14.2
* Operating System: macOS High Sierra 10.13.6

### Additional context
It seems that in version 2.0.0 CodeceptJS is using a newer version of inquirer, when upgrading inquirer from ^0.11.0 to ^2.0.0 certain occurrences were not updated.

For example this snippet:
https://github.com/Codeception/CodeceptJS/blob/d958a75e14023e1ea954f3c9dfac09a4b8613cb2/lib/command/init.js#L115

Should be the following:
```js
]).then((result) => {
```
As `inquirer.prompt` returns a `Promise`.

Which means all occurrences of `inquirer.prompt` need to be updated accordingly, next to initialization (of helper configuration) this includes migration and helper/pageObject/test generation.

cuki:
Closing as this was fixed with #1394.

PR #1394, opened 2018-12-28
Fix/init command

PeterNgTr:
As `inquirer.prompt` returns a `Promise`.

Which means all occurrences of inquirer.prompt need to be updated accordingly, next to initialization (of helper configuration) this includes migration and helper/pageObject/test generation.

https://github.com/Codeception/CodeceptJS/issues/1392


DavertMik:
Thank you very much!

PR #1395, opened 2018-12-28
Fix/no screenshot on windows

PeterNgTr:
Due to different sep on windows and *unix env, screenshot is not saved.


DavertMik:
Thanks

PR #1396, opened 2018-12-28
Fix/wdio 5 is not initialized

PeterNgTr:
Due to this 
```
Let's get serious
(If you haven't already, navigate back to the project root directory)

This was just a warm up. Let's move forward and run WebdriverIO with the test runner. If you want to use WebdriverIO in your project for integration testing we recommend to use the test runner because it comes with a lot of useful features that makes your life easier. With WebdriverIO v5 and up the testrunner has moved into the @wdio/cli NPM package. To get started, we need to install this first:

$ npm i --save-dev @wdio/cli
```


DavertMik:
I don't think we need wdio cli

PR #1403, opened 2018-12-31
Properly render images linked in support documentation

martomo:
Updated occurrences of invalid image urls which were forgotten when changing support documentation.


DavertMik:
:1st_place_medal: 
So nice! Thank you!

PR #1411, opened 2019-01-04
Update webdriverio package to version 5.2.2.

martomo:
* Use `isElementEnabled` instead of previously named `getElementEnabled`.


DavertMik:
Thank you!

PR #1417, opened 2019-01-07
Fix Maximum call stack size exceeded via Object reference in retryFailedStep plugin

xMutaGenx:
Currently when using the plugin you will see `Maximum call stack size exceeded` error due to infinite object reference, rather than a function. Instead we take a copy of passed in `when` and use the static copy, additionally if no `when` is specified in, we should be retrying all of the failed steps.


DavertMik:
Thanks! I also fixed this but I didn't commit my change.

Issue #1413, opened 2019-01-06
Step-by-step Report failed to save screenshot


tsuemura:
#### What are you trying to achieve?

I want to take step by step report. but now it fails because of the problem of resolving path.

```
    › Screenshot is saving to /home/suemura/Work/codeceptjs-sample/record_stepbystep_1546741916/0000.png
    › <Can't save step screenshot: Error: ENOENT: no such file or directory, open '/home/suemura/Work/codeceptjs-sample/record_stepbystep_1546741916/0000.png'> undefined
```

#### What do you get instead?

> Provide console output if related. Use `--verbose` mode for more details.

```bash
$ npm run test -- --grep stepbystep --verbose

> codeceptjs-sample@1.0.0 test /home/suemura/Work/codeceptjs-sample
> codeceptjs run --debug "--grep" "stepbystep" "--verbose"

creating output directory: /home/suemura/Work/codeceptjs-sample/output
CodeceptJS v2.0.2
Using test root "/home/suemura/Work/codeceptjs-sample"
Helpers: REST, Puppeteer, FetchMail, ForceClick, Ssh
Plugins: screenshotOnFail, stepByStepReport

Regression Test --
    [1] Starting recording promises
    Emitted | suite.before ([object Object])
  stepbystep
    Emitted | test.before ([object Object])
    Emitted | test.start ([object Object])
    Emitted | step.before (I am on page "https://google.com")
    Emitted | step.after (I am on page "https://google.com")
    Emitted | step.start (I am on page "https://google.com")
    I am on page "https://google.com"
    › [Url] https://www.google.com/
    › Screenshot is saving to /home/suemura/Work/codeceptjs-sample/record_stepbystep_1546741916/0000.png
    › <Can't save step screenshot: Error: ENOENT: no such file or directory, open '/home/suemura/Work/codeceptjs-sample/record_stepbystep_1546741916/0000.png'> undefined
    Emitted | step.passed (I am on page "https://google.com")
    Emitted | step.finish (I am on page "https://google.com")
    Emitted | test.passed ([object Object])
    Emitted | test.finish ([object Object])
  ✔ OK in 1371ms

    Emitted | test.after ([object Object])
    Emitted | suite.after ([object Object])

  OK  | 1 passed   // 2s
    Emitted | global.result ([object Object])
◉ Step-by-step preview: file:///home/suemura/Work/codeceptjs-sample/output/records.html
    Emitted | global.after ([object Object])```

> Provide test source code if related

[...]

PR #1418, opened 2019-01-07
Fix/no step by step reports

PeterNgTr:
There is no screenshot is saved for step by step reports, so this basically to fix that issue. https://github.com/Codeception/CodeceptJS/issues/1413


DavertMik:
So far so good. But `fileName.includes(fileSep) ` should be replaced with something more meaningful.
I don't think that we should change logic depending on separators.

Issue #1423, opened 2019-01-08
Could not connect to WebDriver  if trying to run tests on browser stack


KaterynaGubar:
#### What are you trying to achieve?
I want to test site on mobile device, using browserstack

#### What do you get instead?
Can't connect to WebDriver

> Provide console output if related. Use `--verbose` mode for more details.

```bash
# paste output here

Can't connect to WebDriver.
Error: getaddrinfo ENOTFOUND http http:80

Please make sure Selenium Server (ChromeDriver or PhantomJS) is running and accessible

**output with verbose:**

  [1] Starting recording promises
    Emitted | suite.before ([object Object])
  "Previous" CTA on Select Shop page to become Back Button @mobile @BKM-223
    Emitted | test.before ([object Object])
    [1] Error | Error: Can't connect to WebDriver.
Error: getaddrinfo ENOTFOUND http http:80

Please make sure Selenium Server (ChromeDriver or PhantomJS) is running and accessible
    Emitted | test.failed ([object Object])
  × "before each" hook: codeceptjs.before for ""Previous" CTA on Select Shop page to become Back Button @mobile @BKM-223" in 9049ms
    [1] Error | Error: Can't connect to WebDriver.
Error: getaddrinfo ENOTFOUND http http:80

Please make sure Selenium Server (ChromeDriver or PhantomJS) is running and accessible
Error: Can't connect to WebDriver.
Error: getaddrinfo ENOTFOUND http http:80

Please make sure Selenium Server (ChromeDriver or PhantomJS) is running and accessible
    Emitted | test.after ([object Object])
    Emitted | suite.after ([object Object])
```

### Details

* CodeceptJS version: 1.4.4
* NodeJS Version: 8.11.3
* Operating System: Windows 10
* Protractor: 5.4.1
* Configuration file:

**appium.conf.js:**

exports.config = {
    output: "./output",
    helpers: {
        REST: {
            endpoint: "****************************",
            defaultHeaders: {
                appId: "***********************************"
            }
        },
        Appium: {
            browser: "iPhone",
            device: "iPhone 8 Plus",
            realMobile: true,
            os_version: "11.0",
            user: "*******",
            key: "********",
            host: "http://hub-cloud.browserstack.com/wd/hub",
            baseUrl: '****************************',
            waitForTimeout: 60000
        },
        MyHelper: {
            require: "./page_objects/helpers/MyHelper.js"
        }
    },

    include: {
        I: "./steps/custom_steps.js",
[...]

PR #1424, opened 2019-01-08
Fix/wdio cannot start with cloud

PeterNgTr:
Some defaults in config prevent cloud services to be initialized. https://github.com/Codeception/CodeceptJS/issues/1423


lib/helper/WebDriver.js
```diff
@@ -244,11 +244,12 @@ class WebDriver extends Helper {
   _validateConfig(config) {
     const defaults = {
       // webdriverio defaults
-      protocol: 'http',
-      host: '0.0.0.0',
-      port: 4444,
-      path: '/wd/hub',
-      logLevel: 'error',
```
DavertMik: Log level `error` must be set to disable verbose output.
All other defaults can be removed... I think

PeterNgTr: updated code as your suggestion

DavertMik:
Ok, let's take it as it is for now, but I will update these settings too.

PR #1438, opened 2019-01-11
Fixes users from merging on autoLogin plugin

nealfennimore:
Prevents from using last user's methods in all cases. E.g. `admin` actually ends up using `inactiveUser`'s login.

```js
	plugins: {
		autoLogin: {
			enabled: true,
			saveToFile: false,
			inject: 'loginAs', // use `loginAs` instead of login
			users: {
				admin: {
					// methods
				},
				activeUser: {
					// methods
				},
				inactiveUser: {
					// methods
				},
			},
		},
	},
```

### DIFF
--- a/lib/plugin/autoLogin.js
+++ b/lib/plugin/autoLogin.js
@@ -143,7 +143,7 @@ const defaultConfig = {
 */
 module.exports = function (config) {
   config = Object.assign(defaultConfig, config);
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks! My bad. 

I was also thinking maybe I should change defaults so restore/fetch methods to be empty by default.
What do you think?

Issue #1437, opened 2019-01-11
I.waitForElement() not Working after last Update in Android


eduardofinotti:
#### What are you trying to achieve?
About this bug:
https://github.com/Codeception/CodeceptJS/pull/1434

#### What do you get instead?

> Provide console output if related. Use `--verbose` mode for more details.

```bash
# paste output here
```

> Provide test source code if related

```js
// paste test
```

### Details

* CodeceptJS version:
* NodeJS Version:
* Operating System:
* Protractor || WebDriverIO || Nightmare version (if related)
* Configuration file:

```js
# paste suite config here
```

narlei:
Same problem :(

When use `I.waitForElement('#com.android...')` return the error: 

`Promise was rejected with the following reason: Error: selector needs to be typeof string or function`

narlei:
Please aprove the PR: https://github.com/Codeception/CodeceptJS/pull/1442

Working with Android now 🙏

PR #1442, opened 2019-01-14
Fix Issue 1437 - waitFor...

eduardofinotti:
Fixed the issue #1437.

I tested on iOS and Android platforms, and it's working.

- Added "this" as paramter in parseLocator.call(), in Appium.js file.
- Added “sec” as paramter in waitFor…, in Appium.js file.


eduardofinotti:
@DavertMik Thanks man! 
I'm happy to help the community.

Issue #1445, opened 2019-01-14
Test with data and "only" run only latest data


Vorobeyko:
#### What are you trying to achieve?
Run test with different data and "only"

#### What do you get instead?
The test was performed only with the latest data

Defect in this part in regexp
```js
    //codeceptjs/ui.js
    context.Scenario.only = function (title, opts, fn, data) {
      const reString = `^${escapeRe(`${suites[0].title}: ${title}`.replace(/( \| {.+})?$/g, ''))}`;
      mocha.grep(new RegExp(reString));
      return addScenario(title, opts, fn, data);
    };
```

> Provide console output if related. Use `--verbose` mode for more details.

```bash
Pages/API/Private/Pages/Create --
  ✔ Test name | Название страницы на русском cd5211ad-8f61-4131-94a2-71762d96ded1 in 271ms

  OK  | 1 passed   // 907ms
```

> Provide test source code if related

```js
Data([
  `Automation name ${uuid4()}`, 
  `Название страницы на русском ${uuid4()}`
]).only.Scenario('Test name', async (I, pagesFactory, current) => {
  let spec = await I.havePageSpec({name: current})

  let pages = await pagesFactory.createPages([spec])
  I.seeValue('name', pages[0].page, { name: current })
})
```

### Details

* CodeceptJS version: 2.0.3
* NodeJS Version: 10.15.0
* Operating System: Ubuntu 18.04
* Protractor || WebDriverIO || Nightmare version (if related)
* Configuration file:

Issue #1443, opened 2019-01-14
Logged Test name | [object Object] when used Data


Vorobeyko:
#### What are you trying to achieve?
I used Data for parameterized tests.
Used data as array (or generator) like 
```js
Data(
  yield {name: `Automation name ${uuid4()}`};
  yield {name: `Название страницы на русском ${uuid4()}`};
).only.Scenario('Test name', async (I, pagesFactory, current) => {})
```

#### What do you get instead?
Test title logged like Test name | [object Object]

> Provide console output if related. Use `--verbose` mode for more details.

```bash
Pages/API/Private/Pages/Create --
  Test name | [object Object]
    Runner: runTest
    I have page spec {"name":"Automation name 5633285d-e405-49c0-9526-334aa451900b"}
    pages: createPage
    I call private method "createPage"
```

### Details

* CodeceptJS version: 2.0.3
* NodeJS Version: 10.15.0
* Operating System: Ubuntu 18.04
* Protractor || WebDriverIO || Nightmare version (if related) REST

PR #1446, opened 2019-01-14
Fix issue 1445 and 1443

Vorobeyko:
#1445 
#1443

### DIFF
--- a/lib/data/context.js
+++ b/lib/data/context.js
@@ -49,15 +49,11 @@ module.exports = function (context) {
 };
 
 function replaceTitle(title, dataRow) {
-  if (typeof dataRow.data === 'object') {
-    return `${title} | ${JSON.stringify(dataRow.data)}`;
-  }
-
-  if (Array.isArray(dataRow.data)) {
+  if (typeof dataRow.data !== 'object') {
     return `${title} | {${JSON.stringify(dataRow.data)}}`;
   }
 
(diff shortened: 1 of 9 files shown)


lib/data/context.js
```diff
@@ -46,6 +48,18 @@ module.exports = function (context) {
   };
 };
 
+function replaceTitle(title, dataRow) {
+  if (typeof dataRow.data === 'object') {
+    return `${title} | ${JSON.stringify(dataRow.data)}`;
```
DavertMik: Looks like JSON.stringify should work fine for all cases :)

lib/data/context.js
```diff
@@ -46,6 +48,18 @@ module.exports = function (context) {
...
+  }
+
+  if (Array.isArray(dataRow.data)) {
+    return `${title} | {${JSON.stringify(dataRow.data)}}`;
```
DavertMik: `{${JSON.stringify(dataRow.data)}}` - additional `{}` is not needed here

Vorobeyko: additional `{}` needed, because in file `ui.js` on 139 line you used regexp `/( \| {.+})?$/g` where used `{}` and when data is array of string codeceptjs run only last data

lib/data/context.js
Vorobeyko: fix this.
And use for all cases

DavertMik:
Thank you!

PR #1448, opened 2019-01-15
Skip test in allure-report

Vorobeyko:
Good decision will be then when will be added event `test.skipped`.
But this decision faster and lighter.


DavertMik:
Thank you!
I should add such event

PR #1449, opened 2019-01-15
add tags to allure report + fix collect tag

Vorobeyko:
+ Add tags to allure report
image

+ when set tag like `@skip-wdio` codeceptjs collected only `@skip`


DavertMik:
Thanks!

PR #1450, opened 2019-01-16
Add payload parameter to DELETE method on REST helper

thxv3n0lvl:
I was trying to execute a DELETE method with a payload then I realized it just accepts headers.
Didn't investigate much about it and hopefully I'm not violating anything.

Found this one:
https://stackoverflow.com/questions/299628/is-an-entity-body-allowed-for-an-http-delete-request

TL;DR
```If a DELETE request includes an entity body, the body is ignored [...]```

### DIFF
--- a/lib/helper/REST.js
+++ b/lib/helper/REST.js
@@ -210,10 +210,11 @@ class REST extends Helper {
    * @param {*} url
    * @param {object} headers
    */
-  async sendDeleteRequest(url, headers = {}) {
+  async sendDeleteRequest(url, payload = {}, headers = {}) {
     const request = {
       baseURL: this._url(url),
       method: 'DELETE',
+      data: payload,
       headers,
     };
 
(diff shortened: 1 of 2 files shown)


DavertMik:
Ok, this looks like a very rare case. If you could switch `headers` and `payload` to not break compatibility, I will merge this.

atrevino-1819:
hey @DavertMik wouldn't it be a bit weird to use 
POST this way: `sendPostRequest(url, payload, header)`
and DELETE this other way: `sendDeleteRequest(url, header, payload)` ?

Will be checking why unit tests are passing if they are not using headers, perhaps they're not properly designed.

textbook:
Having the ordering of the parameters be different isn't ideal, but otherwise your PR will break existing tests for anyone passing headers to a DELETE. Although a body is _allowed_ (see e.g. https://stackoverflow.com/q/299628/3001761), it has no defined semantics - maybe it would be better to use `_executeRequest` for unconventional requests.

DavertMik:
So far I recommend using it in custom helper.
As this is a very rare case I'd like not to change the API

PR #1453, opened 2019-01-18
fixed the case when there are more than one matched element

PeterNgTr:
Fixed this case:

> I have noticed one thing about grabAttributeFrom method: there are 6 elements on a page and I expected grabAttributeFrom to return me an array. Here is my code
I.amOnPage('https://www.mcmakler.de/lp/de/immobilienbewertung'); const src = await I.grabAttributeFrom('.footer-branded div img', 'src'); console.log(src);

but it actually returns the value from the first element and that’s it. The page I am testing is public, so you can see that there are actually 6 elements found using this selector.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -36,6 +36,7 @@
     "no-cond-assign": 0,
     "import/no-unresolved": 0,
     "no-await-in-loop": 0,
-    "arrow-body-style": 0
+    "arrow-body-style": 0,
+    "no-loop-func": 0
   }
 }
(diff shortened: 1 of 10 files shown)


ewojjowe:
@PeterNgTr, is the change applicable to other helpers? if so, can you please also update them? Thanka!

PeterNgTr:
@ewojjowe so far, the wdio works. I haven't checked the nighmare and protractor helpers. Would definitely update them if they need

ewojjowe:
@PeterNgTr, yep for wdio is already added. For nightmare and protractor only.

DavertMik:
Yeah, please check Nightmare and Protractor...

PeterNgTr:
Pushed the update for Nightmare and Protractor helpers.

DavertMik:
Thanks!

PR #1455, opened 2019-01-18
Added WebDriver to autoDelay plugin supported helpers list

mattin4d:
WebDriverIO is deprecated and should be removed from the list when the deprecation notices are removed. For now, both should be supported.


DavertMik:
Thanks. I almost forgot about this plugin.
Cool that you are using it!

PR #1461, opened 2019-01-23
fix(locator): fix xpath expressions

dshvets90:
Fix xpath expressions for locate


DavertMik:
Tests failed and never fixed

PR #1466, opened 2019-01-25
Make override option in `run-multiple` to respect the generated overridden config

kinyat:
`run-multiple` generates some overridden configs based on config file.

The current problem we have is that when we try to override one of the `run-multiple` generated config, it will simply pass down to the child runner and not respecting the `run-multiple` generated configs.

One of the symptoms is that when we override the `tests` config, it will make all runner running the same tests because the chunked config is being overridden.

This PR is to make `run-multiple`generate config based on the overridden version so that when we overrides the setting, it is still respecting the `run-multiple` generated configs.


DavertMik:
Thanks!

PR #1470, opened 2019-01-28
Fix/1422 save screenshot on windows

PeterNgTr:
@hubidu can you try to test this fix on windows?


DavertMik:
Sorry, I'm not taking library from 2015 on board.
Maybe it is something regular `path` can do?

PeterNgTr:
Will find another solution ^^

LukoyanovE:
@peterngtr, looks pretty simple)

DavertMik:
Thanks!

PR #1471, opened 2019-01-29
docs: fix bootstrap description for custom runner documentation

LukoyanovE:
Fixed doc to avoid https://codecept.discourse.group/t/boostrap-is-not-a-function-error/64
After #1270 

Also fix a few lines to linter requirements.

@DavertMik, maybe we should add linter run on commit before test run?
And add ~"run linter" line in CONTRIBUTING doc ?

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -467,7 +467,10 @@ codecept.initGlobals(__dirname);
 Container.create(config, opts);
 
 // initialize listeners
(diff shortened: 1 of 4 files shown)


DavertMik:
> maybe we should add linter run on commit before test run?
And add ~"run linter" line in CONTRIBUTING doc ?

Yes, good idea!

Issue #1346, opened 2018-11-22
 bootstrapAll/teardownAll pathes not working with run-multiple parallel


vitaly87:
#### What are you trying to achieve?
I am trying to run bootstrapAll/teardownAll with run-multiple parallel but getting error 
#### What do you get instead?

> throw new TypeError('Path must be a string. Received ' + inspect(path));
   ^

LukoyanovE:
Looks like global variables, defined for `run` and several others commands happens in Codecept.init()
```
  init(dir) {
    this.initGlobals(dir);
    // initializing listeners
    Container.create(this.config, this.opts);
    this.runHooks();
  }
...
  initGlobals(dir) {
    global.codecept_dir = dir;
    global.output_dir = fsPath.resolve(dir, this.config.output);
...
```

`run-multiple` does not call `Codecept.init` , 
I think `codecept_dir` and `output_dir` definiton should be moved to separate method and called for each command before `Codecept` instance creation

DavertMik:
Should be fixed with workers

PR #1484, opened 2019-02-02
Fix bootstrapAll and teardownAll launch with path as argument

LukoyanovE:
Fix for #1346;

global variable with path was not defined for `run-multiple`, so `bootstrapAll` and `teardownAll` were run with error

```
TypeError: Path must be a string. Received undefined
    at assertPath (path.js:28:11)
    at Object.join (path.js:1236:7)
    at module.exports (CodeceptJS/lib/hooks.js:8:31)
    at Command.module.exports (CodeceptJS/lib/command/run-multiple.js:67:3)

```


DavertMik:
Thanks

PR #1486, opened 2019-02-03
Fix url protocol detection for non-http urls

LukoyanovE:
1. Fixed url protocol detection for non-http urls.
Now `file://path/to/file`, `ftp://...` and similar is correct fo Puppeteer, Nightmare and Protractor helpers.
Before `https` and `https` have worked only.

2. Fixed test for page slider in `Puppeteer_test.js`
It could not load w3c site and test fails last 2 days.
Now this test use local sandbox html.

### DIFF
--- a/lib/helper/Nightmare.js
+++ b/lib/helper/Nightmare.js
@@ -401,7 +401,7 @@ class Nightmare extends Helper {
    * ```
    */
   async amOnPage(url, headers = null) {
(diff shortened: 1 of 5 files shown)


DavertMik:
Thanks!
Not sure who will need that but ok :)

LukoyanovE:
For serverless testing of libraries on local files, for example

PR #1489, opened 2019-02-05
add husky rules for pre-commit and pre-push

LukoyanovE:
add linter run on pre-commit
and linter with test runs on pre-push

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,5 +95,10 @@
     "npm": ">=5.6.0"
   },
   "es6": true,
-  "husky": {}
+  "husky": {
+    "hooks": {
+      "pre-commit": "npm run lint",
+      "pre-push": "npm run lint && npm run test"
+    }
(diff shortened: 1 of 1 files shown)


DavertMik:
How about lint --fix
I always run --fix 
Also it should not affect gh-pages branch.

LukoyanovE:
I just did, as I saw.
As I know from big projects' repositories, it's used as warning for programmer, not as auto-formatter. (! as I know !).
Programmers rule the code. Instruments notice them.
For auto-format 'prettier' and similar with IDEs are used.


But there are no problem to implement `--fix`.
Maybe it's better.
One way or another, code will be modified for linter expectations .

If You, as repository owner, wanna try auto fix, why not?
Just choose red or blue pile )

DavertMik:
Ok, let's have it like this

PR #1490, opened 2019-02-05
Webdriver. Add possibility to define conditional checks interval for waitUntil

LukoyanovE:
Add 4th argument to `waitUntil` in `WebDriver` and `WebDriverIO`
It defines interval time between conditional checks.

https://webdriver.io/docs/api/browser/waitUntil.html
http://v4.webdriver.io/api/utility/waitUntil.html

### DIFF
--- a/lib/helper/WebDriver.js
+++ b/lib/helper/WebDriver.js
@@ -1742,12 +1742,13 @@ class WebDriver extends Helper {
   /**
    * {{> ../webapi/waitUntil }}
    *
-   *
+   * @param interval (optional) time in seconds between condition checks.
    * * *Appium*: supported
    */
-  async waitUntil(fn, sec = null, timeoutMsg = null) {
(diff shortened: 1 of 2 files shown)


DavertMik:
Maybe interval check should be moved into config?
4 params are bit to much

LukoyanovE:
> Maybe interval check should be moved into config?
Config parameter good idea for all case setting. I can implement it similar to `smartWait`.
But if User needs `waitUntil` interval change in one-two cases from bunch of them only, how it should be set?

> 4 params are bit to much
It is optional parameter, like other two.

There are not so many cases, when User sets interval different from default. That's why WebDriverIO use it as last.
I have a few queries to server/base, where time limit could longer.
For example, I don't need to send queries to server every 0.5 second, if it updates one time per 0.5-2 minutes. 10-15 seconds are enough.

PR #1492, opened 2019-02-06
Fix bootstrapAll and teardownAll calls from exported object

LukoyanovE:
`bootstrapAll` and `teardownAll` were not called, if created as `bootstrap` and `teardown` in object, like:
```
module.exports = {
  "bootstrapAll": () => {...},
  "teardownAll": () => {...},
};
```

Before this fix, to call `bootstrapAll` and `teardownAll` from object, object should be defined this way:
```
module.exports = {
  "multiple.bootstrap": () => {...},
  "multiple.teardown": () => {...},
};
```
It's not obvious and it is not described anywhere in docs.

Fixed and wrote some tests for `bootstrapAll`/`teardownAll`.


LukoyanovE:
@davert, ok, I'll do it

DavertMik:
Thanks. Let's do it!

Issue #1494, opened 2019-02-07
[Allure report] For deleteSuccessful option set to false in stepByStepReport - passed test cases in Allure still have screenshots attached


AdamOakman:
#### What are you trying to achieve?

Only successful test cases should have screenshots attached in Allure report.

#### What do you get instead?

Every single test case and step (including passed) has a screenshot attached in Allure report.

> Provide test source code if related

### Details

* CodeceptJS version: 2.0.4
* NodeJS Version: docker
* Operating System: docker
* Protractor
* Configuration file:

```js
    plugins: {
        screenshotOnFail: {
            enabled: true,
            fullPageScreenshots: false
        },
        allure: {
            enabled: true
        },
        autoDelay: {
            enabled: true,
            delayBefore: 100,
            delayAfter: 150
        },
        stepByStepReport: {
            enabled: true,
        }
    },
```

**deleteSuccessful** option set to false by default according to documentation

PeterNgTr:
```
 * ##### Allure Reports
 *
 * If Allure plugin is enabled this plugin attaches each saved screenshot to allure report.
 *
 *
```

AdamOakman:
@PeterNgTr Sorry. My bad. I corrected issue summary and description. I had "**deleteSuccessful** option set to **true** by default" on my mind because that's how this works I think. Passed test case in Allure still have screenshots attached in this case.

PeterNgTr:
So I think current, either test passed or failed when you have allure with stepByStep enabled, screenshots are always attached to allure. :)

AdamOakman:
That's how it is! Shouldn't we make them attached for passed tests only? if below configuration happens
`        
stepByStepReport: {
            deleteSuccessful: true,
            enabled: true
        }
`

PeterNgTr:
That combination, I guess yes, a PR is always welcome. :)

PeterNgTr:
so, `screenshotOnFail` is enabled by default, you don't need to add this in your config.

AdamOakman:
Thank you @PeterNgTr

PR #1496, opened 2019-02-09
fixed no screenshots on passed cases allure

PeterNgTr:
By this reported https://github.com/Codeception/CodeceptJS/issues/1494 I found out that:

- The screenshot to attach to allure for failed test cases is already handled in `screenshotOnFail.js`
```
        const allureReporter = Container.plugins('allure');
        if (allureReporter) {
          allureReporter.addAttachment('Last Seen Screenshot', fs.readFileSync(path.join(global.output_dir, fileName)), 'image/png');
        }
```

so, we don't need to handle it again in `stepByStepReport.js`
```
    const allureReporter = Container.plugins('allure');
    if (allureReporter && (step.status !== 'success')) {
      output.plugin('stepByStepReport', 'Adding screenshot to Allure');
      allureReporter.addAttachment('Screenshot', fs.readFileSync(path.join(dir, fileName)), 'image/png');
    }
```

this will lead to 
- duplicate screenshots on allure reports.
- screenshot will be added to passed cases on allure reports.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -270,7 +270,7 @@ Possible config options:
 
 -   `config`  
 
-## stepByStepReport
+## screenshotsForAllureReport
 
 ![step-by-step-report][2]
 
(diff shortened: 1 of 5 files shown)


DavertMik:
Ok, but If I want step-by-step reports for allure?

PeterNgTr:
you mean, you explicitly want to add screenshots for passed cases on allure? I think that's possible if I change this abit

PeterNgTr:
But still there are a lot of screenshots are attached to allure
screenshot 2019-02-11 at 13 57 40

DavertMik:
Can we bring this as a config option?

PeterNgTr:
I‘ll check it out.

PeterNgTr:
Now when the `screenshotsForAllureReport` is set to true, you will see something like this
screenshot 2019-02-25 at 14 03 51

DavertMik:
Thanks!

PR #1503, opened 2019-02-14
Enable ScenarioContext functions for data driven tests

jjm409:
Currently the `ScenarioContext` functions `throws()`,` fails()`, `retry()`, `timeout()`, `config()` and `tag()` are not supported when running data driven scenarios with `Data()`. The pull request would make them available for data driven scenarios and apply them to all scenarios within the data table.


jjm409:
@PeterNgTr Added tests and moved the DataScenarioConfig into its own class.

DavertMik:
Thanks. This is a very needed PR

PR #1504, opened 2019-02-15
New puppeteer plugin to dump code coverage

dvillarama:
After creating your beautiful tests, you'll eventually get asked what is your code coverage.

This plugin will dump the code coverage for you.

You'll need to convert the coverage to istanbul readable coverage and run istanbul to create the report.
(I'll provide my script that does exactly that when I get time to clean it up for public use).

### DIFF
--- a/lib/plugin/puppeteerCoverage.js
+++ b/lib/plugin/puppeteerCoverage.js
@@ -66,7 +66,7 @@ function buildFileName(test, uniqueFileName) {
  *  To work, you need the client javascript code to be NOT uglified.  They need to be built in "development" mode.
  *  And the end of your tests, you'll get a directory full of coverage per test run.  Now what?
  *  You'll need to convert the coverage code to something istanbul can read.  Good news is someone wrote the code
- *  for you (see puppeteer-to-istanbul link below).  Then using istanbul you need to combine the converted 
+ *  for you (see puppeteer-to-istanbul link below).  Then using istanbul you need to combine the converted
  *  coverage and create a report.  Good luck!
  *
  *  Links:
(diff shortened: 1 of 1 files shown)


lib/plugin/puppeteerCoverage.js
```diff
@@ -0,0 +1,154 @@
...
+              process.cwd(),
+              options.coverageDir,
+            );
+            fs.mkdirSync(coverageDir, { recursive: true });
+
+            const coveragePath = path.resolve(
+              coverageDir,
+              buildFileName(test, options.uniqueFileName),
+            );
+            debug(`writing ${coveragePath}`);
```
DavertMik: I think there should be some notification to user that code coverage was saved.
Use require('../output').print

lib/plugin/puppeteerCoverage.js
```diff
@@ -0,0 +1,154 @@
...
+
+  if (!helper) {
+    console.error('Coverage is only supported in Puppeteer');
+    return; // no helpers for screenshot
+  }
+
+  const options = Object.assign(defaultConfig, helper.options, config);
+
+  event.dispatcher.on(event.all.before, async (suite) => {
+    debug('*** Collecting coverage for tests ****');
```
DavertMik: This can be printed to end-user.
Use `require('../output').debug`

lib/plugin/puppeteerCoverage.js
villarama37: done.

lib/plugin/puppeteerCoverage.js
villarama37: done.

DavertMik:
Thank you!

PR #1506, opened 2019-02-18
WebDriver: Fix cloud provider hostname rewrite

LukoyanovE:
`undefined` was assigned to `hostname` config options,
So cloud provider host detection was broken: 
object 
```
{"protocol":"https","hostname":"hub-cloud.browserstack.com","port":443}
```
was rewrited by incorrect object field:
```
{ ...,
  hostname: undefined,
  ...
}
```


DavertMik:
Thanks!

PR #1507, opened 2019-02-18
WebDriver: Fix a failed _after Hook if not specified desiredCapabilities

Vorobeyko:
If not specified desiredCapabilities in config - tests failed in _after hook


DavertMik:
Could you be more specific here? What is error message? 
Why does it needs a browserName? And I think that browserName is set every time you define a browser in config. 

Thanks

Vorobeyko:
Sorry, this was my mistake.

PR #1513, opened 2019-02-21
Fix/locator method withChild

Vorobeyko:
When I searched for the second input with name= 'name1', I had two elements! First input with name= ' name0' and input with name='name1' but I tried to get one item

```js 
locate('.form-field').withChild(locate('//input[@name="name1"]'));
```
HTML:
```html
<div class="form-field">
  <input name="name0" label="Выберите услугу" type="text" value=""/>
</div>
<div class="form-field">
  <input name="name1" label="Выберите услугу" type="text" value=""/>
</div>
```


lib/locator.js
```diff
@@ -116,7 +116,7 @@ class Locator {
   }
 
   withChild(locator) {
-    const xpath = sprintf('%s[//%s]', this.toXPath(), removePrefix((new Locator(locator, 'css')).toXPath()));
+    const xpath = sprintf('%s//descendant::%s', this.toXPath(), removePrefix((new Locator(locator, 'css')).toXPath()));
```
LukoyanovE: How about change to `'%s//child::%s'` and create separate method `withDescendant` ?
Looks more "xpathier", than "child" as any descendant?

DavertMik:
Thanks!

Issue #1502, opened 2019-02-14
Is there a way to add screenshot to JUnit report as attachment?


iljao:
Hi

I'd like to generate a JUnit report for my tests, and include a screenshot, which is automatically saved by codeceptjs on fail.
Searching didn't give any results. But maybe there is a simple way to achieve this though?

This is a part of my config describing reporter options:
```
  helpers: {
    Puppeteer: {
      uniqueScreenshotNames: true,
      [...]
    }
  },
  mocha: {
    reporterOptions: {
      codeceptjs-cli-reporter: {
        stdout: "-",
        options: {
          debug: true
        }
      },
      mocha-junit-reporter: {
        stdout: "./output/console.log",
        options: {
          mochaFile: "./output/result.xml"
        }
      }
    }
  },
```
here's how I run tests:
`codeceptjs run --reporter=mocha-multi`

PeterNgTr:
@iljao well, I took some times to look at this, there are several things to make this works:

1. The current npm of this repo`https://github.com/michaelleeallen/mocha-junit-reporter` is not updated, which means attachments will not work -> https://github.com/michaelleeallen/mocha-junit-reporter/commit/36ea8b8ee95f8a3c3950d2dd7a19f2ea58f453c1#diff-168726dbe96b3ce427e7fedce31bb0bcR23 -> to solve this, you need either wait for the author to publish the latest version or clone it ;)

2. We need also to turn it on from `codeceptjs` side, the config should be like this

```javascript
mocha: {
        reporterOptions: {
          'codeceptjs-cli-reporter': {
            stdout: "-",
            options: {
              debug: true
            }
          },
          'mocha-junit-reporter': {
            stdout: "./test/codeceptjs/output/console.log",
            options: {
              mochaFile: "./test/codeceptjs/output/result.xml"
            },
            attachments: true // <<< this line
          }
        }
      },
```

and also there is a change in the core of the this file `screenshotOnFail.js`

```javascript
if (Container.mocha().options.reporterOptions['mocha-junit-reporter'] && Container.mocha().options.reporterOptions['mocha-junit-reporter'].attachments) {
          test.attachments = [path.join(global.output_dir, fileName)];
        };
```

PeterNgTr:
in the end, it will work when everything is changed.
```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites name="Mocha Tests" time="0.989" tests="1" failures="1">
[...]

PR #1514, opened 2019-02-21
Add junit example in Multi Reports

serieznyi:
I spend few hours before understand how use XUnit with Mocha Multi Reports.
Thanks God I found example of configuration in https://github.com/Codeception/CodeceptJS/issues/1502


DavertMik:
Thanks! That will be useful for others

PR #1516, opened 2019-02-22
Add doc for IE 11 and selenoid

Vorobeyko:


DavertMik:
Thanks! I just stumbled upon connecting Selenoid a few days ago.

Issue #1524, opened 2019-02-25
grabValueFrom returns empty array on firefox and new codeceptjs version


yogainalift:
#### Getting empty array for firefox when trying to grabValueFrom element
Additionally chrome returns array of 1 element even though there was only 1 result.

> Console output with --verbose flag enabled (relevant part for firefox)
```bash
Emitted | step.passed (I wait 2)
Emitted | step.finish (I wait 2)
Emitted | step.start (I grab value from "#exampleFormControlSelect1")
I grab value from "#exampleFormControlSelect1"
Emitted | step.passed (I grab value from "#exampleFormControlSelect1")
Emitted | step.finish (I grab value from "#exampleFormControlSelect1")
[]
Emitted | test.passed ([object Object])
Emitted | test.finish ([object Object])
```

> Reproduction test

```js
Feature('CodeceptJS demo');

Scenario('check Welcome page on site', async (I) => {
  I.amOnPage('https://getbootstrap.com/docs/4.0/components/forms/');
  I.waitForElement('#exampleFormControlSelect1');
  I.selectOption('#exampleFormControlSelect1', '5');
  I.wait(2);
  const value = await I.grabValueFrom('#exampleFormControlSelect1');
  console.log(value);
});

```

### Version Details

* CodeceptJS version: 2.0.6
* NodeJS Version: 10.12.0
* Operating System: linux
* WebDriver version: 5.6.2
* selenium standalone: 3.141.59
* geckodriver: v0.24.0
* firefox: 60.3.0esr
* chromedriver: 2.46
* chrome: 72.0.3626.96

### Configuration file:

```js
const outputFolder = './output';

exports.config = {
  tests: './tests.js',
  output: './output',
  helpers: {
    WebDriver: {
      url: 'http://localhost:4200',
      browser: 'firefox',
      windowSize: '2560x1440',
      restart: false,
      desiredCapabilities: {
        'moz:firefoxOptions': {
          args: [
            '-headless',
          ],
        },
      },
      // Might cause issues, remove if it does.
      timeouts: {
        script: 60000,
        pageLoad: 10000,
      },
      waitForTimeout: 10000,
      coloredLogs: true,
    },
    Mochawesome: {},
  },

  include: {
    I: './steps_file.js',
  },
  bootstrap: null,
  name: 'testcjs',

  multiple: {
    headless: {
      browsers: [
        {
          browser: 'firefox',
        },
        {
          browser: 'chrome',
[...]

PR #1525, opened 2019-02-25
WDIO v5 - Updated some deprecated methods

PeterNgTr:
This PR updated some deprecated method names when running against wdio v5 with firefox for instance and will fix the issue in https://github.com/Codeception/CodeceptJS/issues/1524

- https://webdriver.io/docs/api/element/getTagName.html
- https://webdriver.io/docs/api/element/getAttribute.html
- https://webdriver.io/docs/api/element/getValue.html


DavertMik:
Ups: `TypeError: el.getValue is not a function`...
I think you should use `if (browser.isW3C)` to detect which API to use...

PeterNgTr:
Tests are now passed!!!

DavertMik:
Thank you!

PR #1526, opened 2019-02-25
typescript may not mark async functions correctly so just await everything

RohanHart:
no need to check for async


DavertMik:
Sorry, it looks like your change actually breaks our tests.
Please look into travis CI log

RohanHart:
the scenario test wrapper always creates callback-style asynchronous tests so the assertions needed to wait for the callback

DavertMik:
Thanks!

PR #1541, opened 2019-03-05
fix locator's simplify call, fix toXPath function

LukoyanovE:
Fix error `TypeError: locator.stringify is not a function` for iOS app testing

https://stackoverflow.com/questions/54992632/codeceptjs-locator-stringify-is-not-a-function-error-while-running-codecept-with/

I think it will not fully fix error.
fuzzy selectors are not processed for iOS app testing


DavertMik:
Thanks. 
Fuzzy selectors are not technically possible in iOS. Maybe they are but they are slow as a turtle.
Probably we should recommend to ignore them on iOS

LukoyanovE:
Fix for #1547

PR #1542, opened 2019-03-05
adding downloadFile action for puppeteer

PeterNgTr:
- Introduce download action for puppeteer.
- test is added.

This is basically to click a button, link in AUT to download file. This file is downloaded to output folder. And this function makes sure download function is working on the application.

For instance: you have a web application that let users can upload/ download their images, documents. By using `I.downloadFile('locator')`, if the file is downloaded to output folder, then downloading function is working. Moreover, after the file is downloaded, we can do further check with downloaded file if we want, like parsing the pdf and checking contents, etc.

### DIFF
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -891,6 +891,7 @@ class Puppeteer extends Helper {
     try {
       return new Promise((resolve, reject) => {
         const wstream = fs.createWriteStream(outputFile);
+        console.log(response.data);
         wstream.write(response.data);
         wstream.end();
         this.debug(`File is downloaded in ${outputFile}`);
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -891,7 +891,6 @@ class Puppeteer extends Helper {
     try {
       return new Promise((resolve, reject) => {
         const wstream = fs.createWriteStream(outputFile);
-        console.log(response.data);
         wstream.write(response.data);
         wstream.end();
         this.debug(`File is downloaded in ${outputFile}`);
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -889,11 +889,14 @@ class Puppeteer extends Helper {
     const outputFile = path.join(`${global.output_dir}/${fileName}`);
 
     try {
-      const wstream = fs.createWriteStream(outputFile);
-      wstream.write(response.data);
-      wstream.end();
-      this.debug(`File is downloaded in ${outputFile}`);
-      return fileName;
+      return new Promise((resolve, reject) => {
+        const wstream = fs.createWriteStream(outputFile);
+        wstream.write(response.data);
+        wstream.end();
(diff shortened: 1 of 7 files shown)


PeterNgTr:
@DavertMik This is following up the old PR -> https://github.com/Codeception/CodeceptJS/pull/1343

test/helper/Puppeteer_test.js
```diff
@@ -558,6 +577,15 @@ describe('Puppeteer', function () {
       assert.notEqual(before, after);
     });
   });
+
+  describe.only('#downloadFile', () => {
```
DavertMik: ....ups! 
Someone forgot `only` :)

lib/helper/Puppeteer.js
```diff
@@ -851,6 +853,50 @@ class Puppeteer extends Helper {
...
+   * {{> ../webapi/downloadFile }}
+   */
+  async downloadFile(locator) {
+    let fileName;
+    await this.page.setRequestInterception(true);
+    this.click(locator);
+
+    const xRequest = await new Promise((resolve) => {
+      this.page.on('request', (request) => {
+        fileName = request.url().split('/')[request.url().split('/').length - 1];
```
DavertMik: I have some concerns about the filename.
Let's look into 2 scenarios:

1. File name is generated and there is no way to obtain it. How can we access that file in next steps of a test? 
2. File name is constant. Each next test overwrites the previous test. What if at one run file was not downloaded but test passes because, the file was left from previous run? 

I propose to generate a file name based on a template passed as second parameter.

```js
I.downloadFile('.download', 'avatar_{id}.jpg');
```

A method should return a complete file name so it could be used in a test with FileSystem helper:

```js
let filename = await I.downloadFile('.export', 'users_{id}.txt');
I.seeFile(filename);
I.seeInThisFile('Davert');
I.seeInThisFile('PeterNgTr');
```

test/helper/Puppeteer_test.js
PeterNgTr: removed 😆

lib/helper/Puppeteer.js
PeterNgTr: amazing proposal. I made the changes.
- now it returns the `fileName`
- moreover if the custom name is provided, it's used for downloaded file, otherwise default one is used.

PeterNgTr:
screenshot 2019-03-06 at 13 29 38
Not sure why it's failed on CI.

DavertMik:
But not for Travis :(

PeterNgTr:
Should we skip that test on CI?

DavertMik:
Can we somehow figure out what's going wrong on Travis? 
Puppeteer is not executed in a container, so should not be an issue of filesystem.
I think you could try to add console logs and see what happens there.
Just to make us confident this feature works for everyone

PeterNgTr:
@DavertMik oops, well, it's passed. -> https://travis-ci.org/Codeception/CodeceptJS/jobs/504583768

DavertMik:
Thanks!

Issue #1545, opened 2019-03-07
[DOCS] autoLogin example with local storage has error


RikdeVos:
If I use the example in the docs the restore function of autoLogin always fails.

The docs state:
`I.executeScript(() => localStorage.setItem('session_id', session), session);`

Instead it should be
`I.executeScript((session) => localStorage.setItem('session_id', session), session);`

PR #1546, opened 2019-03-07
fixed autoLogin local storage example

RikdeVos:
Issue: https://github.com/Codeception/CodeceptJS/issues/1545

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -247,7 +247,7 @@ plugins: {
       },
       restore: (I, session) => {
         I.amOnPage('/');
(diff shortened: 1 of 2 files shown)


DavertMik:
Thanks, I reproduced this example by memory.
But could you puth this fix into https://github.com/Codeception/CodeceptJS/blob/master/lib/plugin/autoLogin.js

plugins.md is generated from it.

RikdeVos:
Good one, fixed it

DavertMik:
Thank you!

PR #1550, opened 2019-03-09
Allure/ Added gitlab executor info

Vorobeyko:


DavertMik:
How about moving this to a different plugin?
GitLab executors goes beyond standard Allure reports.

PR #1551, opened 2019-03-09
Allure/ Add ability enable screenshotDiff plugin

Vorobeyko:
See [Allure Screenshot Plugin](https://github.com/allure-framework/allure2/blob/master/plugins/screen-diff-plugin/README.md)


DavertMik:
@puneet0191 what do you think about it? I think Resemble Helper should support this functionality

puneet0191:
@DavertMik I think it is a great help, I will add support for Allure in coming days. thanks @Vorobeyko

PR #1556, opened 2019-03-11
Fix typo

pablopaul:


DavertMik:
Thanks, could you send this update to https://github.com/Codeception/CodeceptJS/blob/master/lib/helper/ApiDataFactory.js ?

pablopaul:
@DavertMik sure, done. sorry for always forgetting about that.

PR #1564, opened 2019-03-14
fixed issue 1561

PeterNgTr:
I want to test if the exact text "Authentication failed." appears in the p subelement of

```
<div class="ui error message">
  <div class="content">
    <p></p>
  </div>
</div>
```

**Let's say now there is no value in p.**
`So I.seeTextEquals("Authentication failed.", ".error.message"); should fail.`

### DIFF
--- a/test/helper/WebDriver_test.js
+++ b/test/helper/WebDriver_test.js
@@ -199,7 +199,7 @@ describe('WebDriver', function () {
         // e.inspect().should.include("expected element h1 'Welcome to test app' to equal 'Welcome to test app!'");
       }));
 
-    it.only('should check text is not equal to empty string of element text', () => wd.amOnPage('https://codecept.discourse.group/')
+    it('should check text is not equal to empty string of element text', () => wd.amOnPage('https://codecept.discourse.group/')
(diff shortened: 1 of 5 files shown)


test/helper/WebDriver_test.js
```diff
@@ -198,6 +198,13 @@ describe('WebDriver', function () {
         // e.should.be.instanceOf(AssertionFailedError);
         // e.inspect().should.include("expected element h1 'Welcome to test app' to equal 'Welcome to test app!'");
       }));
+
+    it.only('should check text is not equal to empty string of element text', () => wd.amOnPage('https://codecept.discourse.group/')
```
DavertMik: ONLY!

DavertMik: Your linter did a lot of fixes but it bypasses this pretty important one. 
I'm using "eslint-plugin-mocha" for this kind of checks. You can try it too )

PeterNgTr: I'm using the `npm run lint-fix` in the repo :D

DavertMik:
Thank you!

Issue #1569, opened 2019-03-15
[Puppeteer Helper] checkOption wrong(?) behaviour and missing uncheckOption


aml2610:
#### What are you trying to achieve?

I was just testing out the `checkOption` command under Puppeteer helper. Ideally:
1) we would also have the `uncheckOption` command which is currently missing from this helper 
2) if I call `checkOption` two times on the same element it should only check it the first time i.e not attempt to check it again and thus mistakenly uncheck it

#### What do you get instead?

1) `uncheckOption` is not implemented for Puppeteer helper
2) `checkOption` behaviour is broken I think

if I do 
```
Scenario('my scenario', I => {
    I.checkOption('#elm');
    I.checkOption('#elm');
});
```
it will check it both times at the moment

I looked at the code and the function looks like this:

```
  async checkOption(field, context = null) {
    const els = await findCheckable.call(this, field, context);
    assertElementExists(els[0], field, 'Checkbox or radio');
    if (await els[0].getProperty('checked') === true) return;
    await els[0].click();
    return this._waitForAction();
  }
```

so there is a check in place but I think it's wrong, if I replace line `if (await els[0].getProperty('checked') === true) return;` with something like `if (await els[0].getProperty('checked').then(el => el.jsonValue()) === true) return;` it then works as expected.

So basically you have to extract the json value from that call before checking the boolean.

Then implementing `uncheckOption` is simple enough I think, you copy most of the `checkOption` functionality but instead of checking for `=== true` you check for `=== false` so you don't uncheck something that is already unchecked.

I tested both my fix and the addition of `uncheckOption` and they work locally for me. I am happy to make a PR for both if you decide that'd be useful.

Thanks

### Details

* CodeceptJS version: I am testing with up-to-date master branch code
* NodeJS Version: v8.11.0
* Operating System: macOS

PeterNgTr:
Just push your PR, I think it's always more than welcome!

PR #1572, opened 2019-03-18
[Puppeteer Helper] Implement uncheckOption and fix behaviour of checkOption if called twice

aml2610:
Proposed fixes for issues described in https://github.com/Codeception/CodeceptJS/issues/1569


lib/helper/Puppeteer.js
```diff
@@ -707,6 +714,22 @@ class Puppeteer extends Helper {
...
+   *
+   * ```js
+   * this.helpers['Puppeteer']._clickElm('Click me').then // ...
+   * ```
+   *
+   * @param elm the DOM element to click on
+   * @returns {Promise<*>} delegating to waitForAction()
+   * @private
+   */
+  async _clickElm(elm) {
```
DavertMik: and yeah, I don't think this method is needed if it just clicks&waits...

aml2610:
@DavertMik Sure, good points, thanks. I updated the PR.

DavertMik:
Thanks!

PR #1576, opened 2019-03-23
fixed 1561 properly

PeterNgTr:
well the previous patch for `seeTextEquals` is not proper. This aims to make a proper fix.

I found the root cause, the problem is when `codeceptjs` try to get the text from the locator, it returns an `array`

```
const selected = await forEachAsync(res, async el => this.browser.getElementText(getElementId(el)));
```
In this case, the locator has no text it returns an `empty array`. However the the text you want to compare is a `string`. And the `equals` method is 

```
const comparator = function (a, b) {
      return a === b;
    };
```

and by this, it returns `false`, even though both are empty but one is `string`, one is `array`.

### DIFF
--- a/lib/assert/equal.js
+++ b/lib/assert/equal.js
@@ -6,6 +6,9 @@ const output = require('../output');
 class EqualityAssertion extends Assertion {
   constructor(params) {
     const comparator = function (a, b) {
+      if (b.length === 0) {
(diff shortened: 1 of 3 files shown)


DavertMik:
Wow, that was huge issue. Thank you!

PR #1586, opened 2019-03-27
Use local selenium-standalone

pablopaul:


DavertMik:
Any reasons why selenium standalone should be local?

pablopaul:
Yes, to be in line with latest CodeceptJS setup docs which suggests to
install codeceptjs as local node module

DavertMik:
selenium-server-standalone is absolutely unrelated to codeceptjs package. It installs a global cli which can be used to install/update/start selenium server.
Currently, CodeceptJS do not provide any connection with this package, so you always need to start selenium manually.

pablopaul:
OK, I close this issue.

PR #1592, opened 2019-03-31
fix absolute path support

batalov:
I found an issue with running a single test by absolute path, which doesn't allow to run or debug tests in WebStorm IDE


DavertMik:
Thanks!

PR #1596, opened 2019-04-03
Fixed methods first() and toXpath()

AnotherAnkor:
I fixed two methods to make it works.


DavertMik:
Could you add a test to locator_test.js to show the case you are fixing

DavertMik:
Tests are failing... No replies. Closing

PR #1597, opened 2019-04-05
fix async login, check + fetch for autologin plugin

nitschSB:
for more complex login functions it's sometimes necessary to grab
information from the page
this isn't possible without this fix


DavertMik:
Thanks, that makes sense!

PR #1603, opened 2019-04-12
Fix typo. Improve order of prompts.

pablopaul:
I think grouping the questions by topic makes sense - or why do I get asked about custom steps, then about translations and then about custom steps again? Any specific reason?


DavertMik:
Thanks!

PR #1605, opened 2019-04-13
fix actor path replacement in page object template

batalov:
Had a problem with extending actor object with my methods due to incorrect actor require in page object files:
Was: const I = actor();
Now: const I = require('../actor_path.js')();


DavertMik:
Closed in favor of #1608

PR #1607, opened 2019-04-13
Appium test fixes

DavertMik:
@PeterNgTr I'd like to open up your PR #1564 once again, as it appeared that all Appium tests were disabled and they actually didn't run :sob: 

So let's try to enable them once again, and see if they pass. 

Also, I didn't understand what should we do with current swipe* actions? I mean you implemented swipe via `performTouchAction`, however, there is no transition plan for current actions and API is not compatible.

Should we deprecate current swipe*? 
How should they be replaced with newer API?


PeterNgTr:
@DavertMik fixed appium tests here . -> https://github.com/Codeception/CodeceptJS/pull/1609

PR #1610, opened 2019-04-16
Fix Puppeteer launch with browserWSEndpoint

ngadiyak:
Puppeteer crashes with `WebSocket is not open: readyState 3 (CLOSED)` because we call `closeOtherTabs` before set a current page.

Example config
```
Puppeteer: {
            url: "http://localhost",
            chrome: {
                browserWSEndpoint: "ws://localhost:9990/devtools/browser/1bcc1a6c-faa7-42c2-a229-ac20ebac7b36",
                ignoreHTTPSErrors: true
            }
        },
```

### DIFF
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -404,10 +404,6 @@ class Puppeteer extends Helper {
         }
         throw err;
       }
-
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks!

PR #1615, opened 2019-04-18
Add example configuration for TestingBot

testingbot:
Add example configuration for TestingBot + mention alternative
helper https://github.com/testingbot/codeceptjs-tbhelper


DavertMik:
Thank you, will merge that soon!

DavertMik:
Thanks, I need to replace this PR with my own as we need to update lib/helper/WebDriver.js file as well.
But it is taken fully and available on the website

PR #1618, opened 2019-04-23
fixed the scrollTo does not work when running in mobile testing

PeterNgTr:
Fixed this issue
```
Lokyy   [Today at 2:43 PM]
Hello All , For Mobile Testing,  when I use,  I.scrollTo('some-locator'); is giving me the following error -  `Error: Malformed type for "xoffset" parameter of command touchScroll
 Expected: number
 Actual: string`
```


DavertMik:
Thanks!

PR #1631, opened 2019-04-30
Add isElementPresent method

eduardofinotti:
I was a problem. I needed to check if an element was present on my page, to get a decision in my test. So this method returns just true or false in my test. It is can send time out to this method. 

**For example:**

````javascript
var element = await I.isElementPresent('~Allow', 10)

if(element){
   //do it
}else{
   //do it
}
````

Thanks.

### CODE REVIEW

**Comment by PeterNgTr** [comment]
Tips: Please try `npm run lint-fix` before you push your commit to avoid such those lint issues.

**Comment by eduardofinotti** [comment]
@PeterNgTr problems solved :)

**Comment by tsuemura** [comment]
`I is element present`, the grammer of the sentence looks weird to me :confused: 

By the way, is this example works?
I think the `element` has been undefined yet when the `recorder` collects all promises in the promise chain.
Could you write some acceptance tests?

```javascript

var element = await I.isElementPresent('~Allow', 10)

if(element){
   //do it
}else{
   //do it
}
```

PeterNgTr:
how about some thing like `I.getElementPresence`?

eduardofinotti:
> how about some thing like `I.getElementPresence`?

Hummm, sounds good for me.
But my method just return (true or false) if the element is present in the screen or not. Just it. 

I think like: `I.haveElement`. It's good?

DavertMik:
I see the point. But I can't accept this implementation. I will try to come up with alternative implementation soon.

PR #1642, opened 2019-05-07
fixed WebDriver findFields() method: added label priority

haveac1gar:
This PR fixes bug when you're try to fill field on page has field already filled with value included text from label of field you want to fill. Bug is appeared when second field placed under first field.

Example:
I.fillField('coolField', 'value includes label name for awesomeField')
I.fillField('awesomeField', 'new value') <-- this method will fill `coolField` instead of `awesomeField` cause Locator.field.byText() find first matches text, not label

### DIFF
--- a/lib/helper/WebDriver.js
+++ b/lib/helper/WebDriver.js
@@ -2216,7 +2216,7 @@ async function findFields(locator) {
   if (!locator.isFuzzy()) return this._locate(locator.simplify(), true);
 
   const literal = xpathLocator.literal(locator.value);
-  let els = await this._locate(Locator.field.labelEquals(literal));
+  let els = await this._locate(Locator.field.labelContains(literal));
   if (els.length) return els;
 
   els = await this._locate(Locator.field.byText(literal));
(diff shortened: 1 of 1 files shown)


DavertMik:
Value is not take into account during search...

At first we locate element by name or placeholder or label

```js
  byText: literal => xpathLocator.combine([
    `.//*[self::input | self::textarea | self::select][not(./@type = 'submit' or ./@type = 'image' or ./@type = 'hidden')][(((./@name = ${literal}) or ./@id = //label[contains(normalize-space(string(.)), ${literal})]/@for) or ./@placeholder = ${literal})]`,
    `.//label[contains(normalize-space(string(.)), ${literal})]//.//*[self::input | self::textarea | self::select][not(./@type = 'submit' or ./@type = 'image' or ./@type = 'hidden')]`,
  ]),
```

Then we locate element by a label containing a text

```js
  labelContains: literal => xpathLocator.combine([
    `.//*[self::input | self::textarea | self::select][not(./@type = 'submit' or ./@type = 'image' or ./@type = 'hidden')][(((./@name = ${literal}) or ./@id = //label[contains(normalize-space(string(.)), ${literal})]/@for) or ./@placeholder = ${literal})]`,
    `.//label[contains(normalize-space(string(.)), ${literal})]//.//*[self::input | self::textarea | self::select][not(./@type = 'submit' or ./@type = 'image' or ./@type = 'hidden')]`,
  ]),

```

Issue #1636, opened 2019-05-03
ERROR: ENAMETOOLONG saving screenshots with complex capabilities


GSasu:
#### What are you trying to achieve?
Trying to take screenshots on failures when running in selenoid with complex browser capabilities running in multiple

#### What do you get instead?

```bash
Error: ENAMETOOLONG: name too long, open '/path/to/project/output/smoke__browser_firefox_desiredCapabilities_moz_firefoxOptions_args_purgecaches_window_size_375x667_prefs_general_useragent_override_Mozilla_5_0_iPhone_CPU_iPhone_OS_11_0_like_Mac_OS_X_AppleWebKit_604_1_38_KHTML_like_Gecko_Version_11_0_Mobile_15A372_Safari_604_1_log_level_trace_acceptInsecureCerts_true_selenoidOptions_enableVNC_true_enableLog_true_logName_session_log_log_sessionTimeout_10m__2/record_Add_virtual_selectio_1556886655/index.html
```
The more capabilities you set for your sessions the larger the name gets. Tried with both `uniqueScreenshotNames` set to false and true, no difference.

```js
browsers: [
            {
                browser: 'chrome',
                desiredCapabilities: {
                    'goog:chromeOptions': {
                        args: args: [
                           '--headless',
                           '--disable-gpu',
                           '--no-sandbox',
                           '--disable-setuid-sandbox',
                           '--ignore-certificate-errors',
                           `--user-agent=${(process.env.SCREENSIZE === 'mobile') ? MOBILE_UA : DESKTOP_UA}`,
                            // uncomment the following line if you need the browser console
                            // '--auto-open-devtools-for-tabs',
                        ],
                    },
                    selenoidOptions: {
                         enableVNC: true,
                         enableLog: true,
                         logName: 'session-log.log',
                         sessionTimeout: '10m',
                   },
                },
            },
            {
                browser: 'firefox',
                desiredCapabilities: {
                    'moz:firefoxOptions': {
                        args: [
                             '-headless',
                             '-purgecaches',
                             `--window-size=${getDevice(process.env.SCREENSIZE).width}x`
                             + `${getDevice(process.env.SCREENSIZE).height}`,
                             // uncomment the following line if you need the browser console
                             // '-jsconsole',
                        ],
                        prefs: {
                            'general.useragent.override': (process.env.SCREENSIZE === 'mobile')
                                ? MOBILE_UA : DESKTOP_UA,
                        },
                        log: {
                            level: 'trace',
[...]

PR #1643, opened 2019-05-07
Fixed ss name too long when using complex capabilities in run-multiple

GSasu:
Fixed https://github.com/Codeception/CodeceptJS/issues/1636

Error: ENAMETOOLONG: name too long, open '/path/to/project/output/smoke__browser_firefox_desiredCapabilities_moz_firefoxOptions_args_purgecaches_window_size_375x667_prefs_general_useragent_override_Mozilla_5_0_iPhone_CPU_iPhone_OS_11_0_like_Mac_OS_X_AppleWebKit_604_1_38_KHTML_like_Gecko_Version_11_0_Mobile_15A372_Safari_604_1_log_level_trace_acceptInsecureCerts_true_selenoidOptions_enableVNC_true_enableLog_true_logName_session_log_log_sessionTimeout_10m__2/record_Add_virtual_selectio_1556886655/index.html

### DIFF
--- a/lib/command/run-multiple.js
+++ b/lib/command/run-multiple.js
@@ -121,7 +121,7 @@ function executeRun(runName, runConfig) {
   if (browserConfig.outputName) {
     outputDir += typeof browserConfig.outputName === 'function' ? browserConfig.outputName() : browserConfig.outputName;
   } else {
-    outputDir += JSON.stringify(browserConfig).replace(/[^\d\w]+/g, '_');
+    outputDir += JSON.stringify(browserName).replace(/[^\d\w]+/g, '_');
   }
   outputDir += `_${runId}`;
 


GSasu:
Hmmm, no idea why the protractor tests failed in circleci, locally they passed fine plus nothing was touched on protractor. Any ideas why that failed occured?

DavertMik:
I restarted the build but I don't think that simplifying it to just a browser name is the good solution.
But when we run same test in different window sizes we will get the same output. That's not so good. Maybe we could build a hash out of capabilities? Or shorten the filename?
What do you think?

GSasu:
> I restarted the build but I don't think that simplifying it to just a browser name is the good solution.
> But when we run same test in different window sizes we will get the same output. That's not so good. Maybe we could build a hash out of capabilities? Or shorten the filename?
> What do you think?

Hashing the capabilities sounds like a good plan! Thank you for that, I'll make the change and push them as soon as I get some time again, hopefully today.

DavertMik:
Closed in favor of #1671

PR #1644, opened 2019-05-07
Update ts Locator definitions

LukoyanovE:
Updated `ILocator` definition (not founded use of `value` and removed it, added `id`)

Simplify TS definitions for functions with `locator`.

Fixed tests for parser.
Added tests for locator constructor (not necessary, but why not).

Need some TS developers review for errors, style and improvements

### DIFF
--- a/lib/command/definitions.js
+++ b/lib/command/definitions.js
@@ -29,14 +29,16 @@ declare class ScenarioConfig {
 }
 
 interface ILocator {
+  id?: string;
   xpath?: string;
   css?: string;
   name?: string;
-  value?: string;
   frame?: string;
   android?: string;
   ios?: string;
 }
+
+type LocatorOrString = string | ILocator | Locator;
 
 declare class Helper {
   /** Abstract method to provide required config options */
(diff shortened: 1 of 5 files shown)


LukoyanovE:
maybe `ilocator` should be changed to something like:
```
type ILocator =
  { id: string } |
  { xpath: string } |
  { css: string } |
  { name: string } |
  { frame: string } |
  { android: string } |
  { ios: string } |
  { react: string, props?: {}, state?: {} }
```

DavertMik:
I'm approving, waiting for others to review.

lib/command/definitions.js
```diff
@@ -123,14 +119,10 @@ declare function BeforeSuite(callback: ICodeceptCallback): void;
...
-declare function locate(selector: ILocator): Locator;
-declare function within(selector: string, callback: Function): Promise<any>;
-declare function within(selector: ILocator, callback: Function): Promise<any>;
-declare function session(selector: string, callback: Function): Promise<any>;
-declare function session(selector: ILocator, callback: Function): Promise<any>;
-declare function session(selector: string, config: any, callback: Function): Promise<any>;
-declare function session(selector: ILocator, config: any, callback: Function): Promise<any>;
+declare function locate(selector: LocatorOrString): Locator;
+declare function within(selector: LocatorOrString, callback: Function): Promise<any>;
+declare function session(selector: LocatorOrString, callback: Function): Promise<any>;
```
tsuemura: Is `session` only taking a string as an first argument?

LukoyanovE: Look like You are right.
Don't see any usage of `session` except with string

PR #1654, opened 2019-05-17
Feat/uncheck option for nightmare and protractor

PeterNgTr:
Introduce the `unCheckOption` for `Nightmare` and `Protractor`.


DavertMik:
Thanks, just a few things to improve. It should be `uncheckOption` as it is now in WebDriver and Puppeteer. Also please add some tests to `tests/helpers/webapi` for this method

DavertMik:
Thanks!

PR #1660, opened 2019-05-20
Added Documentation for possible ways to locate elements in React Native apps

DimGun:
I'm pretty new to the CodeceptJS, so maybe I'm missing some obvious approach to do the same thing.
Please suggest if I'm reinventing a wheel.

When I tried to make a test that would work on both platforms in the React Native app, I found out that test works on iOS but fails on Android. After some investigation I figured out that on Android React Native uses `setTag` to set test id, but UIAtomator2 doesn't have any means to query a view using tag.

After some researches I came to the point that the easiest way will be to use Espresso driver, which has built-in support for tags. But I was not able to use it's `view-tag` strategy directly, as `helper/Appium.js` has very strict rules on possible strategies. After several experiments and diving into espresso-appium-driver source code I finally constructed an XPath expression that allows locator to find element via 'view-tag'.

I collected all necessary setup steps as well as some explanation of the problem in the document. I think it could be helpful for other developers.

Please share your opinion on this subject.

PS
I also tried to build a plugin with custom locator for this approach, but seems like `helper/Appium.js` totally ignores data provided by plugin. I think I will fill bug on this shortly.

### DIFF
--- a/docs/mobile-react-native-locators.md
+++ b/docs/mobile-react-native-locators.md
@@ -1,0 +1,64 @@
+## Automating React Native apps
+
+### Problem
+Let's say we have a React Native app with component defined like this
+```html
+<Button testID='someButton'>My button</Button>
+```
+
+If you will try to execute a simple test like
+```js
+I.tap('~someButton')
+```
+it will work correctly on iOS (with XUITest driver), but on Android's UIAutomator2 it will give you an error
+```
+Touch actions like "tap" need at least some kind of position information like "element", "x" or "y" options, you've none given.
+```
+
+This happens because of the way React Native implements `testID` for Android, it puts provided value into the [View tag](https://developer.android.com/reference/android/view/View#tags),
+as could be found in the [source code](https://github.com/facebook/react-native/blob/19a88d7f4addcd9f95fd4908d50db37b3604b5b1/ReactAndroid/src/main/java/com/facebook/react/uimanager/BaseViewManager.java#L114).
+But UIAutomator doesn't have any means to work with view tags.
+
+### Solutions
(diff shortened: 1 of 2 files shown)


DavertMik:
Wow! That's awesome :pray: 
I didn't have experience with React Native, and it's such a piece of good news you made it run and wrote documentation on that.

Could I ask you to extend this guide a bit more, for not just locators, but write a guide for testing React Native apps? Setting up, writing tests, launching tests, writing locators, etc. That would be very interesting.

> After some researches I came to the point that the easiest way will be to use Espresso driver, which has built-in support for tags. But I was not able to use it's view-tag strategy directly, as helper/Appium.js has very strict rules on possible strategies. After several experiments and diving into espresso-appium-driver source code I finally constructed an XPath expression that allows locator to find element via 'view-tag'.

Maybe we can work on patching `helper/Appium.js`? 
I'm open to solutions.

> I also tried to build a plugin with custom locator for this approach, but seems like helper/Appium.js totally ignores data provided by plugin. I think I will fill bug on this shortly.

This can be. Maybe you could share your code, I will try to look at what might go wrong. I'm not testing mobile apps at the moment so it's hard for me to get into context but I would like to help with everything possible. 

Please, contact me in our Slack https://join.slack.com/t/codeceptjs/shared_invite/enQtMzA5OTM4NDM2MzA4LTNiMTE5YzNlMjQyNWYwZjAzODE3ZTI1MmVkYjY0YTYyNDZmMWJjM2VlODg3NTQzYzM2OGE5ZmIwODFjODk1MzI

Or in Telegram "@davert". 

I hope for a productive collaboration!

DimGun:
Hi, thanks for the fast response. I'm a bit busy these days, but I will find time this or next week to extend documentation and will fill mentioned issue. I'll get back to you soon

DavertMik:
For now I am merging this but I have hope for better React Native guide one day!

PR #1666, opened 2019-05-22
introduce grabAccessibilityTestResults method

PeterNgTr:
**grabAccessibilityTestResults**
Retrieves accessibility test results of given URL. If no URL is given, the URL in the config will be used.

```javascript
let results = await I.grabAccessibilityTestResults('http://localhost:8000/');

//Returned results
{ documentTitle: 'TestEd Beta 2.0',
pageUrl: 'http://localhost:8000/',
issues:
[ { code: 'WCAG2AA.Principle3.Guideline3_1.3_1_1.H57.2',
context: '<html><head><title>TestEd Beta 2.0</t...</html>',
message:
'The html element should have a lang or xml:lang attribute which describes the language of the document.',
type: 'error',
typeCode: 1,
selector: 'html' } ] }
```

**Parameters**
- url (optional, default url in the config file)
- string url to grab accessibility test results

### DIFF
--- a/docs/helpers/Nightmare.md
+++ b/docs/helpers/Nightmare.md
@@ -371,6 +371,32 @@ I.fillField({css: 'form#login input[name=username]'}, 'John');
 -   `field`  located by label|name|CSS|XPath|strict locator.
 -   `value`  text value to fill.
 
+### grabAccessibilityTestResults
+
+Retrieves accessibility test results of given URL.
+If no URL is given, the URL in the config will be used.
+
+```js
+let results = await I.grabAccessibilityTestResults('http://localhost:8000/');
+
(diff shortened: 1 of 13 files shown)


DavertMik:
This method doesn't use anything from a helper where it is defined. It also requires a new dependency...
So I don't think it should be the part of browser control helpers.

So it looks like this functionality should be moved to a standalone helper, and maybe to be extended with methods like: `seeThisPageIsAccessible()` or smth like that. Anyway, this should be a new helper. 

Right now I would recommend to create that helper in a new package and add to our list of custom helper. https://github.com/Codeception/CodeceptJS/wiki/Community-Helpers

PR #1668, opened 2019-05-22
(feat): grab perfermance timing data - puppeteer helper

PeterNgTr:
**grabDataFromPerformanceTiming**
Grab the data from performance timing using Navigation Timing API Resumes test execution, so should be used inside an async function with await operator.

```javascript
await I.amOnPage('https://example.com');
let data = await I.grabDataFromPerformanceTiming();
//Returned data
{ // all results are in [ms]
responseEnd: 23,
domInteractive: 44,
domContentLoadedEventEnd: 196,
loadEventEnd: 241
}
```


DavertMik:
Thanks!

PR #1669, opened 2019-05-23
Fixing path already exists error

seta-tuha:
When running multiple testing scenarios, the output coverage test fail with already exist error


DavertMik:
Thanks!

PR #1674, opened 2019-05-26
fix: add Promise support to avoid UnhandledPromiseRejectionWarning

adrielcodeco:
fix: add Promise support to avoid UnhandledPromiseRejectionWarning on #page.on('load')# when #page.$('body')#

### DIFF
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -2181,10 +2181,10 @@ function $XPath(element, selector) {
 function targetCreatedHandler(page) {
   if (!page) return;
   this.withinLocator = null;
-  page.on('load', frame => {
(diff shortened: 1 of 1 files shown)


DavertMik:
So this issue happens when an iframe doesn't have `<body>` tag inside?

adrielcodeco:
maybe too.  

I found this issue when testing my React application.
My scenario is a BDD test, after login when i call I.amOnPage i had a UnhandledPromiseRejectionWarning: "Execution context was destroyed, most likely because of a navigation."
This PR fixed this issue.

DavertMik:
Thanks! I usually fix "execution context" issues with retries. I hope this can make a better job then

PR #1676, opened 2019-05-27
fix link in configuration doc

LukoyanovE:
Just fix one link in docs


DavertMik:
I renamed them to `reports` :smile:

PR #1687, opened 2019-05-30
Update typescript definitions for inject

LukoyanovE:
Added definition of global `inject` function.

Added definition of scenario's `injectDependencies` function and updated `ICodeceptCallback` generation, to support dynamic injection.

Slightly updated definition template generator to be more linter-according.

### DIFF
--- a/lib/command/definitions.js
+++ b/lib/command/definitions.js
@@ -12,20 +12,20 @@ const template = `
 type ICodeceptCallback = (i: CodeceptJS.{{I}}{{callbackParams}}) => void;
 
 declare class FeatureConfig {
-  retry(times:number): FeatureConfig
-  timeout(seconds:number): FeatureConfig
-  config(config:object): FeatureConfig
-  config(helperName:string, config:object): FeatureConfig
+  retry(times: number): FeatureConfig
+  timeout(seconds: number): FeatureConfig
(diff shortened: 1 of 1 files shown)


lib/command/definitions.js
```diff
@@ -9,23 +9,24 @@ const fs = require('fs');
 const path = require('path');
 
 const template = `
-type ICodeceptCallback = (i: CodeceptJS.{{I}}{{callbackParams}}) => void;
+type ICodeceptCallback = (i: CodeceptJS.{{I}}{{callbackParams}}, ...args: any) => void;
```
LukoyanovE: This is for https://codecept.io/pageobjects#dependency-injection

lib/command/definitions.js
```diff
@@ -180,11 +182,7 @@ module.exports = function (genPath, options) {
 
   let definitionsTemplate = template.replace('{{methods}}', methods.join(''));
   definitionsTemplate = definitionsTemplate.replace('{{exportPageObjects}}', exportPageObjects.join('\n'));
-  if (callbackParams.length > 0) {
```
LukoyanovE: `join` does not add join-symbol, if length == 1.
and return `""`, when length ==0

DavertMik:
Thank you!

Issue #1171, opened 2018-07-30
generate bdd snippetsin specific file


osman-mohamad:
#### What are you trying to achieve?
I want to generate my features snippets in specific file
#### What do you get instead?
I can not do this with the current `codeceptjs bdd:snippets` command

### Details
* CodeceptJS version: 1.32
* NodeJS Version: 9.11.0
* Operating System: ubuntu linux 18.04
* Configuration file:

```js
{
  "output": "./tests/output",
  "helpers": {
    "Puppeteer": {
      "windowSize": "1366x768",
      "url": "http://localhost:8081"
    }
  },
  "include": {
    "I": "./tests/steps_file.js"
  },
  "mocha": {},
  "bootstrap": false,
  "teardown": null,
  "hooks": [],
  "gherkin": {
    "features": "./src/**/features/*.feature",
    "steps": [
      "./src/modules/app-module/crud-section/tests/features/step_definitions/regsiter.js",
      "./src/modules/app-module/crud-section/tests/features/step_definitions/list.js",
      "./src/modules/app-module/crud-section/tests/features/step_definitions/update.js",
    ]
  },
  "tests": "./{src,tests}/**/*.test.js",
  "timeout": 10000,
  "name": "ui"
}```

**Comment by anabellaspinelli** [comment]
I think it's not unusual to have steps grouped in different files but codeceptjs generates the snippets always on the first steps file on the list defined in codecept.json > gherkin > steps.

This would be really useful, because right now we have to generate them and then copied to the right steps file. 

👍

**Comment by osman-mohamad** [comment]
exactly @anabellaspinelli  this is what I am talking about:
```bash
codeceptjs  bdd:snippets --path=./src/modules/app-module/crud-section/tests/features/step_definitions/update.js
```
where the `path` is the path to where the snippet will be generated

patrickvale:
I've just to Codeception and I was a bit surprised that gherkin:snippets just seems to generate output on the command line. I was sort of expecting it to go into a file, or at least that to be configurable, much as generate:feature creates a file.

Is there a technical limitation that prevents this or would a PR be welcome?

khanzzirfan:
Hi Team,
has any one got the below command working , generating the codecept gherkin snippets where codecept.conf.js is in different location not a root directory?
// not working
` npx codeceptjs -c ./e2es/config/codecept.conf.js gherkin:snippets --path=./e2es/steps/stepdefinition.js ` 
// working
[...]

PR #1688, opened 2019-05-30
Adding optional path parameter to bdd:snippets command

cthorsen31:
In reference to #1171 this would allow a user to target a specific file to upload snippets do rather than the first file in the steps array. 

Usage "codeceptjs  bdd:snippets [--path=PATH]"

For this to work the file must already exist and be included in the config steps.


DavertMik:
Thanks, this will be useful for me!

PR #1689, opened 2019-05-30
Feature: add target folder parameter to def command

LukoyanovE:
Added a param `--to`.
It will allow to generate file directly in described folder:

`codeceptjs def --to ./tests/typings`
will create file `<dir path>/tests/typings/steps.d.ts`

I am not sure about name of parameter, and should it generate file name or leave `steps.d.ts`?

### DIFF
--- a/bin/codecept.js
+++ b/bin/codecept.js
@@ -36,7 +36,7 @@ program.command('list [path]')
 program.command('def [path]')
   .description('Generates TypeScript definitions for all I actions.')
   .option('-c, --config [file]', 'configuration file to be used')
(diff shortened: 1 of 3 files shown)


pablopaul:
@LukoyanovE `steps.d.ts` as filename is fine.
I sugges to change `--to` to `--destination`

LukoyanovE:
> I suggest to change `--to` to `--destination`

There are one more possible name - "--output"/"-o".
Maybe it should be used, as more  common

pablopaul:
Good idea!

DavertMik:
Thanks! Looks good

PR #1691, opened 2019-06-01
Doc: CodeceptJS Quick Start with Cucumber/Gherkin

gkushang:


DavertMik:
Thank you! Currently, we have this example page to be generated from our Wiki
It's easier than sending pull requests. I updated the Wiki for page for you https://github.com/Codeception/CodeceptJS/wiki/Examples#quick-start-with-cucumbergherkin
This page will be updated automatically on next site rebuild.

Thank you!

PR #1699, opened 2019-06-06
proxy pass

dmred:
added proxy pass to rest helper
very needed and useful for QAs and pentesters

### DIFF
--- a/lib/helper/REST.js
+++ b/lib/helper/REST.js
@@ -126,10 +126,11 @@ class REST extends Helper {
    * @param {*} url
    * @param {object} headers
    */
-  async sendGetRequest(url, headers = {}) {
+  async sendGetRequest(url, headers = {}, proxy = false) {
     const request = {
       baseURL: this._url(url),
       headers,
+      proxy,
     };
     return this._executeRequest(request);
   }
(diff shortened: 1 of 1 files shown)


dmred:
I can't understand, I've just edited REST helper, why are there any problems with Protractor???

LukoyanovE:
> I can't understand, I've just edited REST helper, why are there any problems with Protractor???

I see same error for protractor and webdriver in my and some others PRs.
Looks like a some dependency problem

DavertMik:
Yeah, Protractor stuff is some unrelated and probably it's something with Travis CI.

However, could you pass the proxy stuff inside a config of REST helper?
I don't think that everyone would love to include proxy config on each request

dmred:
> Yeah, Protractor stuff is some unrelated and probably it's something with Travis CI.
> 
> However, could you pass the proxy stuff inside a config of REST helper?
> I don't think that everyone would love to include proxy config on each request

so, I've added by default turned it off, but having ability to change proxy in *_test.js can help us if we need to change proxy lots of times while testing(or parsing) specific APIs

Well, i think the compromise is to add it to HELPER config and to the third param by default turned off

DavertMik:
>  and to the third param by default turned off

I am really against specifying configuration inside a test file itself. 
Case of switching between proxy servers is pretty rare so I'd like not to introduce a new parameter just for it, as we may need it for something else...  

I think you should write a custom helper which allows different proxy configs:

```js
MyHelper: {
  require: './my_helper',
  proxies: {
    server1: {}, // server config
    server2: {}
  }
}
```
Then inside the helper you can change configuration for REST:

```js
// inside a helper
function changeProxy(name)
  this.helpers.REST._setConfig({ proxy: this.config.proxies[name] });
}
```
Is the idea clear?

DavertMik:
No activity, closing

PR #1700, opened 2019-06-06
Fix extra comma in definition without pages

LukoyanovE:
Fixed a silly error made in #1687 

Without pages it generates double commas:
```js
type ICodeceptCallback = (i: CodeceptJS.I, , ...args: any) => void;
```

Also added `IData` interface for `xData` and fixed a couple of linter errors


DavertMik:
Thank you!

Issue #1679, opened 2019-05-28
support object API changed with 2.1.3 release


johnyb:
#### What are you trying to achieve?

One of our support objects used to be an array. It looks similar to this:

```JavaScript
module.exports = ['user number 1', 'user number2'];
```

It's included as "users" and was used in tests:

```JavaScript
Scenario('Test login', function (users, I) {
    I.login(users[0]);
});
```

#### What do you get instead?

The support object is not an array any longer, after the upgrade to 2.1.3. Related commit would be https://github.com/Codeception/CodeceptJS/commit/314fcc13ed4ddbe80f8af3b6fb44712ff3aee12d - this changes the support object to always be an object and just copy all attributes using `Object.assign`. Regarding the commit message, I'm not sure this was intended. Especially since the upgrade was just on "patch level", not a major version bump.

### Details

* CodeceptJS version: 2.1.3
* NodeJS Version: any
* Operating System: any
* Configuration file:

```js
# relevant part is:
    include: {
        users: './e2e/users'
    }
```

Any hints on how to solve this issue? I'm totally fine with changing code on our side but currently have no clue on how to best solve this.

DavertMik:
Ok, I will tell the reasoning and probably we can come up with the solution.

Here is the thing. I'm pushing the idea to use `inject` as a single way to require actors and support objects. However, this has issues when page object requires an actor or another page object which was not initialized yet. This way all objects initialized as empty objects so they can be referenced by others.

So maybe you may find a more elegant solution for this?

johnyb:
> So maybe you may find a more elegant solution for this?

Sure :) It just means a larger change in our code base and I wanted to make sure this change was by intention. I'm all in for pushing the inject idea forward, since I really like the concept.

johnyb:
Mhhh, after playing around with this a little more, I think there should be a very simple solution which does restore the old behaviour. Besides that, it's closer to actual dependency injection, because the reference to the original module is not lost. I'm going to test this a little more and open a MR with the change.

PR #1701, opened 2019-06-06
fix references to support objects when using DI

johnyb:
some change to how the support objects are initialized removed the original
reference to the support object and replaced it a "cloned" object. However,
since Object.assign was used, not all properties have been copied and especially
the original reference was lost.

This implementation uses a proxy object to load the injected modules on demand
and returns the original reference to the module. This should fix all usages
of support objects, where the type is not a plain object.

Fixes: #1679

### DIFF
--- a/lib/container.js
+++ b/lib/container.js
@@ -161,23 +161,23 @@ function createHelpers(config) {
 }
 
 function createSupportObjects(config) {
-  const objects = {};
+  const objects = container.support = new Proxy({}, {
+    get(target, key) {
+      // configured but not in support object, yet: load the module
+      if (key in config && !(key in target)) target[key] = lazyLoad(key);
+      return target[key];
+    },
+  });
 
-  for (const name in config) {
-    objects[name] = {}; // placeholders
-  }
-
-  if (!objects.I) {
-    objects.I = require('./actor')();
+  if (!config.I) {
+    container.support.I = require('./actor')();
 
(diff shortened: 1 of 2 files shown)


DavertMik:
Thank you! I also wanted to use Proxy for that but was not so smart to implement in a correct way

johnyb:
Concerning the broken tests for webdriver, this seems to be due to an upgrade to latest chrome in the selenium standalone images. Therefor I currently have the docker images fixed to `selenium/standalone-chrome:3.141.59-oxygen`, where I.doubleClick is still working for instance. So I guess this has nothing to do with the change of this PR.

DavertMik:
Sure, it's nothing to do with this PR.

If you know how to fix the builds, could you please update .travis.yml and send a patch? 
Much appreciated

DavertMik:
The day has come. I'm merging!

DavertMik:
@johnyb your implementation broke recursive loading of support objects. Proxy was good but still, we need to initialize objects and then use Object.assign. In a new implementation https://github.com/Codeception/CodeceptJS/blob/codecept-221-release/lib/container.js#L213

* I still use proxy to load objects on demand
* I use Object assign but only for real objects
* I added test for loading objects as an array
* I fixed https://github.com/Codeception/CodeceptJS/issues/1795
* I added a test from @Vorobeyko to the core.

johnyb:
Very good, thanks!

PR #1704, opened 2019-06-07
workaround for selenium standalone chrome issues

johnyb:
for some reason, using latest chrome fails the webdriver tests.
Stick to a version of selenium standalone + chrome which is known to work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,9 +21,9 @@ addons:
 services:
 - docker
 before_install:
-- docker pull selenium/standalone-chrome:latest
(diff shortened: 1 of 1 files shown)


johnyb:
As requested in https://github.com/Codeception/CodeceptJS/pull/1701#issuecomment-499558736 - this should make the pipeline green, again. I still haven't found the root cause of this nor an issue in WebDriverIO, Chromedriver or any other component to track.

DavertMik:
Thank you! This should be it!

PR #1707, opened 2019-06-07
Fix field locator - added finding by attribute "label" too

Ilrilan:


DavertMik:
Sorry, those fields do not have `label` attribute. So I can't accept this PR.
I would recommend you to create a custom locator for such type of elements so you could use:

```js
I.fillField('=Label Input', 'xxxx');
I.fillField({ label: 'Label Input' }, 'xxxx');
```

Both can be implemented as described here https://codecept.io/locators#custom-locators

PR #1708, opened 2019-06-07
Record repl commands into history file

hubidu:
Why is it useful:

Because I can copy&paste the commands from the history file into my test after the interactive session.
The ~history file is written to the codeceptjs_dir.


DavertMik:
Yeah, this can be useful. But history file should be saved into `output` directory, so it won't be accidentally added to git. Also, it would be nice if you will print a text message that the commands were saved after the save.

hubidu:
History file will now be saved to output directory and a message will be printed to console

PR #1711, opened 2019-06-10
Allow Control between Specified Window Handles with WebDriver

emmonspired:
Sometimes we are testing the functionality of websites that we cannot control, such as a closed-source managed package of Salesforce, like ZQuotes within our instance. There are popups that either remain open for configuring data on the screen, or that close as a result of clicking a window, and we need more control over which page is being tested with Codecept at any given time. We use WebDriver to accomplish this. We have a workaround in place by using a custom helper, but it would be nice if these methods to control window handle properties on the Webdriver browser object were available in the core WebDeriver API.

- added methods to WebDriver and mocha test to WebDriver_test
- added mocha test 'allow back and forth between handles: #getWindowHandles #getWindowHandle #switchToWindow'

### DIFF
--- a/lib/helper/WebDriver.js
+++ b/lib/helper/WebDriver.js
@@ -1581,6 +1581,41 @@ class WebDriver extends Helper {
     await browser.buttonUp(0);
   }
 
+  /**
+   * Get all Window Handles.
+   * Useful for referencing a specific handle when calling `I.switchToWindow(handle)`
+   *
+   * ```js
+   * const handles = await I.getWindowHandles();
+   * ```
+   */
+  async getWindowHandles() {
+    return this.browser.getWindowHandles();
+  }
+
+  /**
+   * Get the current Window Handle.
+   * Useful for referencing it when calling `I.switchToWindow(handle)`
+   *
+   * ```js
+   * const handle = await I.getWindowHandle();
+   * ```
+   */
+  async getWindowHandle() {
+    return this.browser.getWindowHandle();
+  }
+
+  /**
+   * Switch to the window with a specified handle.
+   *
+   * ```js
+   * await I.switchToWindow( handle );
+   * ```
+   */
+  async switchToWindow(handle) {
+    await this.browser.switchToWindow(handle);
+  }
+
 
   /**
    * Close all tabs except for the current one.
--- a/lib/helper/WebDriver.js
+++ b/lib/helper/WebDriver.js
@@ -1586,10 +1586,10 @@ class WebDriver extends Helper {
    * Useful for referencing a specific handle when calling `I.switchToWindow(handle)`
    *
    * ```js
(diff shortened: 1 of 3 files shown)


lib/helper/WebDriver.js
```diff
@@ -1581,6 +1581,41 @@ class WebDriver extends Helper {
...
 
+  /**
+   * Get all Window Handles.
+   * Useful for referencing a specific handle when calling `I.switchToWindow(handle)`
+   *
+   * ```js
+   * const handles = await I.getWindowHandles();
+   * ```
+   */
+  async getWindowHandles() {
```
DavertMik: Ok, I agree to have this but let's get to standard naming:

```js
const windows = await I.grabAllWindowHandles();
```

lib/helper/WebDriver.js
```diff
@@ -1581,6 +1581,41 @@ class WebDriver extends Helper {
...
+
+  /**
+   * Get the current Window Handle.
+   * Useful for referencing it when calling `I.switchToWindow(handle)`
+   *
+   * ```js
+   * const handle = await I.getWindowHandle();
+   * ```
+   */
+  async getWindowHandle() {
```
DavertMik:
```js
const window = await I.grabCurrentWindowHandle();
```

lib/helper/WebDriver.js
```diff
@@ -1581,6 +1581,41 @@ class WebDriver extends Helper {
...
+    return this.browser.getWindowHandle();
+  }
+
+  /**
+   * Switch to the window with a specified handle.
+   *
+   * ```js
+   * await I.switchToWindow( handle );
```
DavertMik: example should include grab* actions listed above.

DavertMik: so to know where to get those window handles

lib/helper/WebDriver.js
emmonspired: Ok, I agree - pushing new changes now.

lib/helper/WebDriver.js
emmonspired: Another good suggestion. One moment ...

lib/helper/WebDriver.js
emmonspired: Also making these adjustments. Thanks for reviewing.

emmonspired:
> Please rename methods. Also, could you write this use case in `docs/acceptance.md` or into the https://codecept.discourse.group/c/cookbook No sure where it can work better

Added changes to `docs/acceptance.md`

emmonspired:
Added more descriptive comments. Stand by for next push.

emmonspired:
Ok. Ready to be reviewed again @DavertMik .

DavertMik:
Thanks! Looks very comprehensive!

PR #1712, opened 2019-06-11
German translation

hubidu:
I fixed the translation file to use underscore between words and all lowercase words.


DavertMik:
Thank you!

PR #1716, opened 2019-06-13
Fix Run-Multiple when `tests` is undefined

gkushang:
`config.tests` can be Undefined if the test suite only has Gherkin Files. 

When it's undefined, it throws an error: 

```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type undefined
    at assertPath (path.js:39:11)
    at Object.resolve (path.js:1088:7)```


DavertMik:
Thanks!

PR #1717, opened 2019-06-13
Create better Typescript definition file through JSDoc

### LINKED ISSUE #1213: Type annotations should try use JSDoc

If CodeceptJS really want to play nice with TypeScript, than they should try at least try to parse JSDoc, best would be if the files would not by analysed by other tool that the one that will help developers write it. Therefor I would suggest using TS compiler to get hints about types of functions as well as their return types. Also this would give us ability to comment our code (for library, but for users too).

Take this, it is very well annotated, but I can't enable type checking (`//@ts-check`) due to Helper not being in definitions. (fixed in #1206)
``` js

class NetworkHelper extends Helper {

	/**
	 * Gets first request log that passes matcher function.
	 * @param {{ url: string }} key
	 * @param {(log: IrequestLog) => boolean} matcher
	 * @return {IrequestLog}
	 */
	grabRequestLogThatMatches(key, matcher) {
		return this.grabRequestLog(key).instances.find(matcher);
	}

	/**
	 * Tests request log that passes matcher function exists.
	 * @param {{ url: string }} key
	 * @param {(log: IrequestLog) => boolean} matcher
	 * @param {string} matcherDescription Is used to described what matcher does, if test fails.
	 */
	seeRequestLogMatches(key, matcher, matcherDescription) {
		if (!this.grabRequestLogThatMatches(key, matcher)) {
			throw new Error(`No request logs matches: ${matcherDescription}`);
		}
	}
}

/**
 * Storage of requests.
 * - Using `WeakMap` levreges the need to clean up after test, as the reference to the key are lost.
 * @type {WeakMap<{ url: string }, IrequestLog>}
 */
NetworkHelper.requestMap = new WeakMap();

/** @typedef {{ readonly count: number, instances: { url: string, method: import("puppeteer").HttpMethod, headers: import("puppeteer").Headers, postData: string, response: import("puppeteer").Response }[] }} IrequestLog */
```

I would expect the steps.d.ts to look like this:
``` ts
type IrequestLog = {{ readonly count: number, instances: { url: string, method: import("puppeteer").HttpMethod, headers: import("puppeteer").Headers, postData: string, response: import("puppeteer").Response }[] }};

declare namespace CodeceptJS {
  export interface I {
    //other definitions...
	/**
	 * Gets first request log that passes matcher function.
	 */
    grabRequestLogThatMatches(key: { url: string }, matcher: (log: IrequestLog) => boolean) : IrequestLog,
	/**
	 * Tests request log that passes matcher function exists.
	 * @param matcherDescription Is used to described what matcher does, if test fails.
	 */
    seeRequestLogMatchess(key: { url: string }, matcher: (log: IrequestLog) => boolean, matcherDescription: string) : void,
  }
}
```

#### What do you get instead?

``` ts
declare namespace CodeceptJS {
  export interface I {
    //other definitions...
    grabRequestLogThatMatches(key: string, matcher: string) : Promise<string>,
    seeRequestLogMatchess(key: string, matcher: string, matcherDescription: string) : void,
  }
}
```
This is completely different! Unusable I can't do anything with this... IDE will cry and I can either disable it or comment every user made type annotation...

If at leas you have put type `any`, it would be inconvenient, but this is unusable.

### Details

* CodeceptJS version: 1.3.3
* NodeJS Version: 9.11.1
* Operating System: Windows

lemnis:
I am using typescript within codecept, and bumped quite often into incomplete types. This fork adds the possibility to generate typings from the JSDoc annotation. Still, I just started to fix a small part of the files. But would like to get feedback if you like the direction that I am going.

Would fix #1213

Todo:

- [x] Fix path to other definitions files (https://github.com/Codeception/CodeceptJS/pull/1717/files#diff-b38e49bd02b239814a0c2fa24175ecd4R11)
- [x] Make sure that the static definition file is included for users to use.
- [x] Move the module export to a more appropriate place
- [x] Update information on how to generate and extend the definition file
- [x] Better describe my whole plan
- [x] Check if the generated definitions aren't missing anything
- [x] When typescript is used in object and/or custom helpers use tsc to generate definitions
- [x] Check behavior for translations


docs/webapi/checkOption.mustache
```diff
@@ -9,4 +9,4 @@ I.checkOption('I Agree to Terms and Conditions');
 I.checkOption('agree', '//form');
 ```
 @param {string|object} field checkbox located by label | name | CSS | XPath | strict locator.
-@param {string} context (optional, `null` by default) element located by CSS | XPath | strict locator.
\ No newline at end of file
+@param {string} [context=null] (optional, `null` by default) element located by CSS | XPath | strict locator.
```
DavertMik: yeah, technically you are correct here. However, I had issues generating documentation when defining optional types like this. I spend a day trying to fix documentation rebuild but I found that better solution for now not to use JSDoc styles and write manually `optional` word

LukoyanovE:
@lemnis, good beginning.

But there are one problem.
Codeceptjs creates definitions in `lib/command/definitions.js` script.
It grabs some data from helpers, page objects and some others files described in config, and puts it in `steps.d.ts` on `codeceptjs def` command.

If you look on `definitions.js`, you'll see a template with `FeatureConfig` and `ScenarioConfig` manually described. And these manually described definitions are used in final `steps.d.ts` file instead of yours.

So, as an one of possible (first and simplies) solutions, you can import your definitions from `typings/types.d.ts` as text and paste it into these template.
(Also, you need to add your types.d.ts, generated by `def` command, to git if you choose this way)

There are may be others betters solutions.
Like a remake definitions generation for all codeceptjs.
Maybe make one part a "stable", where codeceptjs core functionality is described. And make a user generated part (custom helpers, page objects, etc) a generated by command.
It's a more "right" way, but long and possibly hard.

For example you can make a first way now, merge it, and then we can make step-by-step changes to implement 2nd way.

LukoyanovE:
And, as I understand, this PR will not fully fix #1213, but a few problems it will start to solve

docs/webapi/checkOption.mustache
lemnis: Is the correct implementation of my code fine, or would you like to something else?

lemnis:
Ok, most of it is now (re)written.  Was planning it from the start, but didn't describe it 😇. 

The full output is:
- A static index.d.ts that contains all static definitions who can't be generated from JSDoc
- A generated types.d.ts that contains all definitions from the JSDoc
- A steps.d.ts generated by the user just like before.

@elukoyanov Do you like current behavior, I believe it kinda does what you were writing.

I removed the previous implementation of generating definitions, as a result, functions without JSDoc (e.g. functions in custom helpers) will lose some autocompletion. For example, an argument with the name 'locator' will not automatically be of type 'LocatorOrString' anymore. I don't know if this a big issue.

Things do check:
- Generated definitions of built-in helpers, especially that all async functions return a Promise type
- How to setup / behavior of imported helpers within the generated file.
- Remove all utility functions that were just for generating definitions in the past.
- Generate new docs
- Move building of the built-in helpers to its own npm script and update RoboFile.php accordingly

lib/command/definitions.js
```diff
@@ -2,160 +2,65 @@ const getConfig = require('./utils').getConfig;
...
-  }
-{{exportPageObjects}}
+interface IData {
+  Scenario: IScenario
+  only: {
+    Scenario: IScenario
+  };
 }
 
-declare module "codeceptjs" {
```
lemnis: Removed it because the implementation didn't match index.js,  was it used somewhere?

typings/index.d.ts
```diff
@@ -1,95 +1,61 @@
-// Type definitions for CodeceptJS 1.0.2
+// Type definitions for CodeceptJS 2.1.4
```
lemnis: Is this file still actively used internally?, as I don't see any indication it could have used externally. If used internally, I will update it accordingly so it will support the same code

elukoyanov: Used and updated, except this line. But I think it should be generated by package.json version

elukoyanov: or removed

lib/command/definitions.js
```diff
@@ -2,160 +2,65 @@ const getConfig = require('./utils').getConfig;
 const getTestRoot = require('./utils').getTestRoot;
 const Codecept = require('../codecept');
 const container = require('../container');
-const methodsOfObject = require('../utils').methodsOfObject;
 const output = require('../output');
-const { toTypeDef } = require('../parser');
 const fs = require('fs');
 const path = require('path');
+const { exec } = require('child_process');
```
elukoyanov: Is it really need?
Don't see any uses.

lemnis:
What are all the methods to extend the `I` object?

What I have found:
- Using built-in helpers
- Using custom helpers
- Using steps_file.js
- Added methods (`say` and `retry`) within the `actor.js` file.

Vorobeyko:
@lemnis can you update your branch with the latest master commits? 
I want to use your changes for my project and test it.

lib/command/definitions.js
```diff
@@ -2,160 +2,61 @@ const getConfig = require('./utils').getConfig;
...
-  before(locator: LocatorOrString): Locator;
-  after(locator: LocatorOrString): Locator;
-  withText(text: string): Locator;
-  withAttr(attrs: object): Locator;
-  as(output: string): Locator;
+declare namespace CodeceptJS {
+  export type I = ${customFilePaths.I ? 'ReturnType<typeof I>' : 'WithTranslation<methods>'};
 }
 
+type ICodeceptCallback = (i: CodeceptJS.I${callbackParams}) => void;
```
Vorobeyko: This callback has a hard order of arguments. And CodeceptJS allows you to transfer here PageObjects in an arbitrary order. And when I don't want to use `I` but only PageObject, autocomplete suggests methods for `I`

typings/index.d.ts
```diff
@@ -1,95 +1,95 @@
...
-  first(): Locator;
-  last(): Locator;
-  inside(locator): Locator;
-  before(locator): Locator;
-  after(locator): Locator;
-  withText(text): Locator;
-  withAttr(attr): Locator;
-  as(output): Locator;
-}
+import * as Protractor from "protractor";
```
Vorobeyko: What happens if i dont have protractor in node_modules?

lib/command/definitions.js
Vorobeyko: This callback has a hard order of arguments. And CodeceptJS allows you to transfer here PageObjects in an arbitrary order. And when I don't want to use I but only PageObject, autocomplete suggests methods for I

typings/index.d.ts
Vorobeyko: What happens if i dont have protractor in node_modules?

lib/command/definitions.js
```diff
@@ -169,75 +70,44 @@ module.exports = function (genPath, options) {
...
-          methods.push(`${params}`);
-        }
-      } else {
-        methods.push(`    ${(name)}: (${params}) => any; \n`);
-      }
-    }
+  for (const name in customFilePaths) {
+    const value = customFilePaths[name];
+    const parsedPath = path.parse(value);
+    const relativePath = path.relative(targetFolderPath, path.join(testsPath, parsedPath.dir, parsedPath.name));
```
Vorobeyko: We can use PageObject from custom package. And then we get an error, like:
```
import RadioGroup = require('./../../tests/codeceptjs/configs/@custom_package/codeceptjs/PageObjects/Fields/Core/RadioGroup');
```

but it has to generate so:
```
require('@custom_package/codeceptjs/PageObjects/Fields/Core/RadioGroup');
```

I suggest you to look in container.createHelpers function

typings/index.d.ts
```diff
@@ -1,95 +1,95 @@
...
-    within: any;
-    session: any;
-    DataTable: any;
-    locate: Locator,
-    inject: any,
-    secret: any,
-    by: any;
+type codeceptjs = {
+  codecept: typeof Codecept,
+  container: typeof Container,
```
Vorobeyko: I have error, if use `import { container } from 'codeceptjs'`
Screenshot from 2019-07-30 11-39-49

Vorobeyko: I got error when use: `import { container } from 'codeceptjs'`
Screenshot from 2019-07-30 11-39-49
```
Module '"../../../node_modules/codeceptjs/typings"' has no exported member 'container'.ts(2305)
```

typings/index.d.ts
```diff
@@ -1,95 +1,95 @@
...
-    By: any;
-    ExpectedConditions: any;
-    element: any;
-    $: any;
-    $$: any;
-    browser: any;
-  }
-}
+declare global {
+  const codeceptjs: codeceptjs;
```
Vorobeyko: Autocomplete doesnt work when I import so: 'import codeceptjs from 'codeceptjs''
image

typings/index.d.ts
```diff
@@ -1,95 +1,95 @@
...
+  
+  const codecept_dir: string;
+  const codecept_helper: Helper;
+  const output_dir: string;
 
-declare interface Window {
-  codeceptjs: any;
-}
+  const locate: typeof Locator.build;
+  const inject: typeof Container.support;
```
Vorobeyko: Autocomplete doesnt work when I import so: `const { I, PageObject} = inject()`

typings/types.d.ts
```diff
@@ -0,0 +1,7571 @@
...
+     * Hook executed after each test
+     */
+    _after(): void;
+    /**
+     * Hook provides a test details
+     * Executed in the very beginning of a test
+     *
+     * @param {IHelperVoidFunc} test
+     */
+    _test(test: IHelperVoidFunc): void;
```
Vorobeyko: test must be Mocha.Test types

typings/types.d.ts
```diff
@@ -0,0 +1,7571 @@
...
+     * @param {IHelperVoidFunc} step
+     * @override
+     */
+    _afterStep(step: IHelperVoidFunc): void;
+    /**
+     * Hook executed before each suite
+     *
+     * @param {IHelperVoidFunc} suite
+     */
+    _beforeSuite(suite: IHelperVoidFunc): void;
```
Vorobeyko: suite must be Mocha.Suite types in:

- _beforeSuite
- _afterSuite

typings/types.d.ts
```diff
@@ -0,0 +1,7571 @@
...
+     * @returns { * }
+     */
+    static support(name?: string): any;
+    /**
+     * Get all helpers or get a helper by name
+     *
+     * @api
+     * @param {string} [name]
+     */
+    static helpers(name?: string): void;
```
Vorobeyko: return type must be any

Vorobeyko:
You must update this branch by master. Because to master added new helpers

typings/types.d.ts
```diff
@@ -0,0 +1,7571 @@
...
+     * @param {*} opts
+     */
+    static create(config: any, opts: any): void;
+    /**
+     * Get all plugins
+     *
+     * @api
+     * @param {string} [name]
+     */
+    static plugins(name?: string): void;
```
Vorobeyko: return type must be an any

typings/types.d.ts
```diff
@@ -0,0 +1,7571 @@
...
+     */
+    _init(): void;
+    /**
+     * Hook executed before each test.
+     */
+    _before(): void;
+    /**
+     * Hook executed after each test
+     */
+    _after(): void;
```
Vorobeyko: methods must be a protected.

typings/index.d.ts
lemnis: It should ignore it and give all defined const the 'any' type. Will double-check this behavior.

typings/index.d.ts
lemnis: Should be fixed

lemnis:
@Vorobeyko Thanks for your review! Some issues are resolved now. Hopefully, I will have soon my hands free, to further describe open issues or questions. You are totally welcome to improve the JSDoc and fix the tests.

package.json
```diff
@@ -68,6 +71,7 @@
     "@pollyjs/adapter-puppeteer": "^2.5.0",
     "@pollyjs/core": "^2.5.0",
     "@types/inquirer": "^0.0.35",
+    "@types/mocha": "^5.2.7",
```
Vorobeyko: duplicate dependencies. See 42 line in "dependencies"

Vorobeyko:
@lemnis Сan you give access to your repository?

typings/index.d.ts
Vorobeyko: I got error:
image

Georgegriff:
It looks like this or addresses https://github.com/Codeception/CodeceptJS/issues/1827 (comments in jsdoc fed to types file) could someone confirm and I'll close my issue :) my bad didn't see this pr!

lemnis:
Yes, this will fix #1827 . I am just taking my time to finish this PR ;)

Georgegriff:
@lemnis awesome, thank you :) ! New to CodeceptJs and loving it, having code completion via typescript is fantastic too :)

lib/command/definitions.js
```diff
@@ -2,387 +2,160 @@ const getConfig = require('./utils').getConfig;
...
-  Scenario(title: string, opts: {}, callback: ICodeceptCallback): ScenarioConfig;
-  only: IScenario;
-}
+declare function actor<T extends { [action: string]: Function }>(
+  customSteps: T & ThisType<WithTranslation<methods & T>>
+): WithTranslation<methods & T>;
+declare function inject(): SupportObject;
+declare function inject<T extends keyof SupportObject>(name: T): SupportObject[T];
+declare const codecept_actor: typeof actor;
+declare const Scenario: Scenario;
```
Vorobeyko: Cannot find name 'Feature'
image

lib/command/definitions.js
```diff
@@ -2,387 +2,160 @@ const getConfig = require('./utils').getConfig;
...
+ * @param {Object<string, string>} pathsToType
+ * @param {Object<string, string>} pathsToValue
+ * @returns
+ */
+function getImportString(testsPath, targetFolderPath, pathsToType, pathsToValue) {
+  const importStrings = [];
+
+  for (const name in pathsToType) {
+    const relativePath = getPath(pathsToType[name], targetFolderPath, testsPath);
+    importStrings.push(`type ${name} = typeof import ('./${relativePath}');`);
```
Vorobeyko: The path should not start with dot

image

lib/command/definitions.js
```diff
@@ -2,387 +2,160 @@ const getConfig = require('./utils').getConfig;
...
-  }
-  return methods;
+/**
+ * Returns the relative path from the  to the targeted folder.
+ * @param {string} originalPath
+ * @param { string} targetFolderPath
+ * @param { string} testsPath
+ */
+function getPath(originalPath, targetFolderPath, testsPath) {
+  if (!originalPath.startsWith('.')) return originalPath;
```
Vorobeyko: an import path cannot end with a '.ts' extension

image

lib/command/definitions.js
```diff
@@ -2,387 +2,160 @@ const getConfig = require('./utils').getConfig;
...
-  frame?: string;
-  android?: string;
-  ios?: string;
-}
+type translation = ${JSON.stringify(translations.vocabulary, null, 2)};
+type WithTranslation<T> = T & import('codeceptjs/typings/utils').Translate<T, translation["actions"]>;
+type methods = ${helperNames.length > 0 ? `${helperNames.join(' & ')} & ` : ''} import('codeceptjs/typings/utils').ActorStatic;
 
-type LocatorOrString = string | ILocator | Locator;
+type ICodeceptCallback = (${convertObjectToParameterList(supportObject)}) => void;
```
Vorobeyko: This callback has a hard order of arguments. And CodeceptJS allows you to transfer here PageObjects in an arbitrary order. And when I don't want to use I but only PageObject, autocomplete suggests methods for I

typings/index.d.ts
```diff
@@ -1,95 +1,92 @@
...
-  last(): Locator;
-  inside(locator): Locator;
-  before(locator): Locator;
-  after(locator): Locator;
-  withText(text): Locator;
-  withAttr(attr): Locator;
-  as(output): Locator;
-}
+/// <reference path="./types.d.ts" />
+/// <reference path="../node_modules/webdriverio/webdriverio-core.d.ts" />
```
Vorobeyko: webdriverio may not be in codeceptjs node_modules. But can be in the root of the node_modules folder. 
And webdriverio can be not installed

lib/command/definitions.js
lemnis: This should have been a caching issue, make sure you reload the definition files before testing

lemnis:
Just a short update, the output is changed a bit from last commits.

Starting from this commit **everyone has automatically type checking**, still running `codeceptjs def` will improve the type checking but it is not needed anymore 😃. 

Have still look into the import paths and some other small things.

lib/command/definitions.js
lemnis: Unfortunately, it is impossible to recreate how codeceptjs works. Right know `Scenario`, `xScenario`, `BeforeSuite`, `AfterSuite`, `Background`, `Before` and `After` have an default order that is the same as in the past. But it can be overwritten within Typescript, e.g.
```ts
Scenario<[MyHelper, CodeceptJS.I]>("...", async (myHelper, I) => {
 ...
});
```

package.json
```diff
@@ -34,11 +35,13 @@
     "json-server:graphql": "node test/data/graphql/index.js",
     "lint": "eslint bin/ examples/ lib/ test/ translations/ gulpfile.js",
     "lint-fix": "eslint bin/ examples/ lib/ test/ translations/ gulpfile.js --fix",
-    "docs": "gulp docs",
+    "docs": "robo docs",
```
elukoyanov: What is `robo` for?

Why do you replaced gulp? And did not update dependencies?
```
yarn run v1.17.3
$ robo docs
/bin/sh: robo: command not found
```

I found out only [this package](https://www.npmjs.com/package/robo).
It was update 6 years ago, has 2-5 weekly downloads only, and has not link to github repo?

Do we really need this package?

elukoyanov: I search in code, and found out gulp is not used in code too.
And there are no gulp in dependencies.

What is `docs` command for?

elukoyanov: robo is php task runner.
It's used here for doc generation ><

elukoyanov: ok

I am going to cry a little bit and think what to do about it

elukoyanov:
@lemnis , could you please describe an order of executing.

I make `yarn`, `yarn docs`, `yarn def`.
`def` command generate next output in console:
```
[TSD-JSDoc] Failed to find parent of doclet 'module.exports' using memberof 'module', this is likely due to invalid JSDoc.
[TSD-JSDoc] Failed to find parent of doclet 'container.mocha' using memberof 'container', this is likely due to invalid JSDoc.
[TSD-JSDoc] Failed to find parent of doclet 'module.exports' using memberof 'module', this is likely due to invalid JSDoc.
[TSD-JSDoc] Failed to find parent of doclet 'context.Scenario' using memberof 'context', this is likely due to invalid JSDoc.
[TSD-JSDoc] Unable to resolve type for an unnamed item, this is likely due to invalid JSDoc. Often this is caused by invalid JSDoc on a parameter. Defaulting to `any`.
[TSD-JSDoc] Unable to resolve type for an unnamed item, this is likely due to invalid JSDoc. Often this is caused by invalid JSDoc on a parameter. Defaulting to `any`.
[TSD-JSDoc] Unable to resolve type for an unnamed item, this is likely due to invalid JSDoc. Often this is caused by invalid JSDoc on a parameter. Defaulting to `any`.
```
Is this correct behavior?

Also I cannot successfully run  `yarn build` command. A bunch of TSC errors appears.

What do I need to do to build project?
Node  8.16

lemnis:
@elukoyanov the NPM scripts are kinda confusing and my edit doesn't help. I don't know if `build` or `docs` are used.

I did update the `docs` as it also confused me at first and make it at least run the right command.

You don't need to build anything in reality, so no need to use `yarn build`.

And yes the output while running `yarn defs` is expected, it can be fixed by adding more jsdoc at the needed code, but because it doesn't break or change the output I left it as is.

@DavertMik Is `build` in use? if not, wouldn't it be better to remove it?

elukoyanov:
> @elukoyanov the NPM scripts are kinda confusing and my edit doesn't help. I don't know if `build` or `docs` are used.
> 
> I did update the `docs` as it also confused me at first and make it at least run the right command.
> 
> You don't need to build anything in reality, so no need to use `yarn build`.
> 
> And yes the output while running `yarn defs` is expected, it can be fixed by adding more jsdoc at the needed code, but because it doesn't break or change the output I left it as is.
> 
> @DavertMik Is `build` in use? if not, wouldn't it be better to remove it?

@DavertMik ?

Vorobeyko:
@lemnis Hi, we are testing your pull request. It will be merged asap.

Vorobeyko:
@lemnis Can you create rebase by master, please?

lemnis:
Is Hound configured correctly?

Vorobeyko:
@lemnis  Thank you! This is awesome changes!

PR #1720, opened 2019-06-14
WIP Testcafe Helper

hubidu:
- Just to show what's going on
- Some of the acceptance tests should work, some won't


lib/helper/Testcafe.js
```diff
@@ -0,0 +1,661 @@
...
+
+    return this.t.navigateTo(url)
+      .catch(mapError);
+  }
+
+
+  /**
+   * {{> resizeWindow }}
+   *
+   * Unlike other drivers Puppeteer changes the size of a viewport, not the window!
```
DavertMik: Should be deleted

lib/helper/Testcafe.js
```diff
@@ -0,0 +1,661 @@
...
+  /**
+   * {{> amOnPage }}
+   */
+  async amOnPage(url) {
+    if (!(/^\w+\:\/\//.test(url))) {
+      url = this.options.url + url;
+    }
+
+    return this.t.navigateTo(url)
+      .catch(mapError);
```
DavertMik: Btw, why do we need that `mapError` function? Can we just throw TestCafe exceptions?

lib/helper/Testcafe.js
```diff
@@ -0,0 +1,661 @@
...
+    this.debugSection('_before', 'Starting testcafe browser...');
+
+    this.isRunning = true;
+
+    // TODO Do we have to cleanup the runner
+    const runner = this.testcafe.createRunner();
+    runner
+      .src(this.dummyTestcafeFile)
+      .screenshots(global.output_dir, true)
+      // .video(global.output_dir)
```
DavertMik: I think video should be enabled in config

lib/helper/Testcafe.js
```diff
@@ -0,0 +1,661 @@
...
+      .catch(mapError);
+  }
+
+  /**
+   * {{> fillField }}
+   * {{ react }}
+   */
+  async fillField(field, value) {
+    return this.t
+      .typeText(findElements.call(this, this.context, field), value, { replace: true })
```
DavertMik: Similarly, you can get `appendField` implemented. The same but `replace: false`

lib/helper/Testcafe.js
```diff
@@ -0,0 +1,661 @@
...
+  /**
+   * {{> see }}
+   *
+   * {{ react }}
+   */
+  async see(text, context = null) {
+    let els;
+    if (context) {
+      els = findElements.call(this, this.context, context)
+        .withText(text);
```
DavertMik: `see` should also check for visibility

lib/helper/Testcafe.js
```diff
@@ -0,0 +1,661 @@
...
+    } else {
+      // TODO Check how that is implemented in Puppeteer
+      els = findElements.call(this, this.context, `//*[contains(text(),'${text}')]`);
```
DavertMik: The text should be escaped to be passed into XPath. Maybe you can have `new Locator().withText(text)` ?

lib/helper/Testcafe.js
```diff
@@ -0,0 +1,661 @@
...
+  }
+
+  // TODO Add url assertions
+
+  /**
+   * {{> setCookie }}
+   */
+  async setCookie(cookie) {
+    if (Array.isArray(cookie)) {
+      throw new Error('cookie array is not supported');
```
DavertMik: just recursively call this function...

lib/helper/Testcafe.js
```diff
@@ -0,0 +1,661 @@
...
+
+      const duration = (Date.now() - start) / 1000;
+      if (duration > sec) {
+        throw new Error('waitForFunction timed out');
+      }
+    }
+  }
+}
+
+const elementByXPath = xpath => Selector(() => {
```
DavertMik: wow, I am surprised TestCafe still doesn't have full XPath support :open_mouth:

lib/helper/Testcafe.js
```diff
@@ -0,0 +1,661 @@
...
+  static _checkRequirements() {
+    try {
+      requireg('testcafe');
+    } catch (e) {
+      return ['testcafe@^1.1.0'];
+    }
+  }
+
+  async _startBrowser() {
+    this.dummyTestcafeFile = createTestFile(); // create a dummy test file to get hold of the test controller
```
DavertMik: Can we have it without extra files? By just calling TestCafe API ?

lib/helper/Testcafe.js
```diff
@@ -0,0 +1,661 @@
...
+class Testcafe extends Helper {
+  constructor(config) {
+    super(config);
+
+    this.iteration = 1;
+    this.testcafe = undefined; // testcafe instance
+    this.t = undefined; // testcafe test controller
+    this.dummyTestcafeFile; // generated testcafe test file
+
+    this.context = undefined; // TODO Not sure if this applies to testcafe
```
DavertMik: context is used for `within()` function.
It requires to have `_withinBegin`and `_withinEnd` implemented. 
Inside `_withinBegin` we should define that all next element calls should be started from a specific element (this.context).

test/docker-compose.yml
```diff
@@ -85,6 +85,20 @@ services:
       - ./support:/support
       - node_modules:/node_modules
 
+  test-acceptance.testcafe:
```
DavertMik: This test should be added to https://github.com/Codeception/CodeceptJS/blob/master/.circleci/config.yml#L7 so we could add it to the pipeline

test/docker-compose.yml
```diff
@@ -85,6 +85,20 @@ services:
...
+    build: ..
+    env_file: .env
+    environment:
+      # TODO Add a testcafe tag
+      - CODECEPT_ARGS=-c codecept.Testcafe.js --grep @Puppeteer
```
DavertMik: `--grep @TestCafe`

test/helper/Testcafe_test.js
```diff
@@ -0,0 +1,701 @@
...
+    it('should NOT uncheck option that is NOT currently checked', async () => {
+      await I.amOnPage('/info');
+      await I.uncheckOption('interesting');
+      // Unchecking again should not affect the current 'unchecked' status
+      await I.uncheckOption('interesting');
+      await I.dontSeeCheckboxIsChecked('interesting');
+    });
+  });
+
+  describe('#handleDownloads', () => {
```
DavertMik: I think this is only for Puppeteer, so you can remove that

test/helper/Testcafe_test.js
```diff
@@ -0,0 +1,701 @@
...
+  //     await I.amOnPage('/');
+  //     const res = await I.grabDataFromPerformanceTiming();
+  //     expect(res).to.have.property('responseEnd');
+  //     expect(res).to.have.property('domInteractive');
+  //     expect(res).to.have.property('domContentLoadedEventEnd');
+  //     expect(res).to.have.property('loadEventEnd');
+  //   });
+  // });
+
+  // webApiTests.tests();
```
DavertMik: The most important is to make those tests pass as they are the same for all helpers.
Other tests are helper specific.

test/helper/Testcafe_test.js
```diff
@@ -0,0 +1,701 @@
...
+    });
+
+    it('should open absolute url', async () => {
+      await I.amOnPage(siteUrl);
+      const url = await I.grabCurrentUrl();
+      return url.should.eql(`${siteUrl}/`);
+    });
+  });
+
+  // describe('grabDataFromPerformanceTiming', () => {
```
DavertMik: Puppeteer only - just remove

test/helper/Testcafe_test.js
```diff
@@ -0,0 +1,701 @@
...
+    it('should trigger hover event', () => I.amOnPage('/form/hover')
+      .then(() => I.moveCursorTo('#hover'))
+      .then(() => I.see('Hovered', '#show')));
+
+    it('should not trigger hover event because of the offset is beyond the element', () => I.amOnPage('/form/hover')
+      .then(() => I.moveCursorTo('#hover', 100, 100))
+      .then(() => I.dontSee('Hovered', '#show')));
+  });
+
+  describe('#switchToNextTab, #switchToPreviousTab, #openNewTab, #closeCurrentTab, #closeOtherTabs, #grabNumberOfOpenTabs', () => {
```
DavertMik: I don't think TestCafe supports this. Delete those tests

lib/helper/testcafe/testcafe-utils.js
```diff
@@ -0,0 +1,28 @@
+const fs = require('fs');
+const path = require('path');
+
+const createTestFile = () => {
+  // TODO Generate unique test files (concurrent test runs)
+  const TESTFILE = path.join(__dirname, '_test.js');
```
DavertMik: why do you generate it inside `lib/helper/testcafe`?
if you really need this file - put it into the `output` directory. 
Also, I think it's ok to use some sort UUID or time function to create a file if anyway you delete it in the end. This way you will be able to run parallel tests. 

But again is there a way to work without files?

test/docker-compose.yml
DavertMik: Those tests should be introduced after all unit tests pass

test/helper/Testcafe_test.js
```diff
@@ -0,0 +1,701 @@
...
+      .then(() => I.waitInUrl('about:blank'))
+      .then(() => I.switchToPreviousTab(2))
+      .then(() => I.wait(2))
+      .then(() => I.waitInUrl('/info'))
+      .catch((e) => {
+        assert.equal(e.message, 'There is no ability to switch to previous tab with offset 2');
+      }));
+  });
+
+  describe('popup : #acceptPopup, #seeInPopup, #cancelPopup, #grabPopupText', () => {
```
DavertMik: THose are also Puppeteer specific.
Btw, how TestCafe handles popups?

lib/helper/Testcafe.js
```diff
@@ -0,0 +1,661 @@
...
+      .src(this.dummyTestcafeFile)
+      .screenshots(global.output_dir, true)
+      // .video(global.output_dir)
+      .browsers(this.options.show ? this.options.browser : `${this.options.browser}:headless`)
+      .reporter('minimal')
+      .run({
+        skipJsErrors: true,
+        skipUncaughtErrors: true,
+        quarantineMode: true,
+        selectorTimeout: this.options.waitForTimeout,
```
DavertMik: should be configurable

lib/helper/Testcafe.js
hubidu: will fix

lib/helper/Testcafe.js
hubidu: Testcafe methods throw an error object with lots of information which is quite unreadable if it goes unmapped. However I will see if I can find a better solution.

lib/helper/Testcafe.js
hubidu: Makes sense

lib/helper/Testcafe.js
hubidu: ok

lib/helper/Testcafe.js
hubidu: already done in one of the latest commits

lib/helper/Testcafe.js
hubidu: yeah me too: https://github.com/DevExpress/testcafe/issues/1178

lib/helper/Testcafe.js
hubidu: I dont see how. They don't have a real API. The only thing you can do is run testcafe test files. Therefore this trick. Got it from here https://github.com/rquellh/testcafe-cucumber

lib/helper/Testcafe.js
hubidu: ok

test/helper/Testcafe_test.js
hubidu: Will focus on that

test/helper/Testcafe_test.js
hubidu: I dont know yet

lib/helper/Testcafe.js
hubidu: I can make them configurable, but IMHO these feature do not make so much sense in a codeceptjs context.

lib/helper/testcafe/testcafe-utils.js
hubidu: Will make them uuid and put them to output.

Would be glad if someone finds a way without generating test files, but I could not.

DavertMik:
@hubidu what's current status?
Which methods still need to be implemented? Any issues?

hubidu:
@DavertMik  I am currently making my way through the webapi.js tests as you suggested.
Currently implementing "#selectOption". Had some problems with doubleClick and rightClick which I mysteriously could not get to work (didn't even work when I was clicking manually in the opened browser window). Also faced an issue connected to using xpaths to find a context element and then search by text within this element which I could not get to work either.

Apart from that its going ok.

Do you want to have all tests implemented for a first release or would a subset also do (for an experimental release)?

DavertMik:
> Currently implementing "#selectOption".

From my experience, that's a functionality which is always hard to implement. Good luck with this one!

> Do you want to have all tests implemented for a first release or would a subset also do (for an experimental release)?

A subset of the test would be fine. Inside `webapi` you can use `if (isHelper('TestCafe')) return;` just like we do for similar cases in Nightmare.

And let's put it into experimental release, that's a good idea.

We will also need to work to make it support `within` but that's another story, probably after the helper is merged.

DavertMik:
@hubidu are we ready to merge this helper?
We can mark it as experimental and ask for feedback. I can also take over some of the parts of implementation.

hubidu:
@DavertMik Well I could just skip the remaining tests in webapi.js from line 466 onwards for Testcafe and then we make it ready to be merged into an experimental release?
Afterwards we split the remaining work?

pablopaul:
@hubidu @DavertMik I am here for some testing if experimental release is ready :)

DavertMik:
@hubidu Sure! I think we are ready to merge it and include into the next release. 
We will need further implementation of missing methods and a guide. But let's try it first!

PR #1721, opened 2019-06-17
Fix issue 1161

haily-lgc:
Root cause:
- Throwing an error in Scenario make recorder stopped. Then "after each" hook will restart recorder by using recorder.startUnlessRunning() and its error handler also will be reset to null. Finally, the hook cannot done and hang at this time.

Solution:
- Fix logic code


DavertMik:
Thank you! @Vorobeyko will update the code to include test for this fix and we are ready to release

PR #1723, opened 2019-06-19
Fix issue of "host" configuration not work correctly on Appium.

trinhpham:
Solution: handle webdriverio specs correctly


DavertMik:
Thank you!

PR #1728, opened 2019-06-21
Add S3 AWS Plugin to auto copy output files to S3

Vorobeyko:


DavertMik:
Nice! 
However, I would recommend to remove Amazon S3 from direct dependencies. And ask user to install that manually. Check for that dependency on start, and throw exception asking user to install it.

Vorobeyko:
> Nice!
> However, I would recommend to remove Amazon S3 from direct dependencies. And ask user to install that manually. Check for that dependency on start, and throw exception asking user to install it.

OK!

Vorobeyko:
@DavertMik done =)

package.json
```diff
@@ -53,6 +53,7 @@
     "inquirer": "^6.3.1",
     "js-beautify": "^1.9.1",
     "lodash.merge": "^4.6.1",
+    "mime-types": "^2.1.24",
```
DavertMik: I'd rather not add an extra library for just a plugin.
I see two options here:
1. ask to install mime-types as well as aws-sdk
2. move this plugin into a standalone package (probably the cleanest approach).

I think the option 2 would be the best:
- AWS SDK and mime-types will be installed automatically
- plugin will be separated from the core 

So I think we should consider moving this plugin into `@codeceptjs/s3` package.
We can still add it to official list on the website or to 3rd-party list at https://github.com/Codeception/CodeceptJS/wiki/Community-Helpers

Vorobeyko:
I choose the second option!)

DavertMik:
Moved to https://github.com/Codeception/codeceptjs-plugins

PR #1729, opened 2019-06-21
fix: Tests did not fail when an exception raised in the async bootstrap

Vorobeyko:


lib/hooks.js
```diff
@@ -39,14 +40,39 @@ function loadCustomHook(module) {
...
   } else {
     callable();
     if (done) done();
   }
 }
 
+function callAsync(callable, done = undefined) {
+  let called;
+
+  if (done) called = callable(done);
```
DavertMik: If there is done function there are no `catch` in response

DavertMik:
Thanks!

PR #1733, opened 2019-06-23
feat: Mocking network requests/response using PollyJS

### LINKED ISSUE #1732: feat: Mock request/response in Puppeteer helper via PollyJS.
**Type:** feature
**Helper:** Puppeteer

- Mock request/response of any network call

> Examples
```js
I.mock('GET', '/api/users', 200); 

...

const { status } = I.sendGetRequest('/api/users');
console.log(status) // 200


```

VikalpP:
This will be used to mock the network request/response.
User will need to start Mocking before use it.

resolves #1732


lib/helper/Puppeteer.js
```diff
@@ -1837,6 +1838,60 @@ class Puppeteer extends Helper {
...
+   * ```js
+   * I.mock('GET', '/api/users', 200);
+   * I.mock('ANY', '/secrets/*path', 403);
+   * I.mock('POST', '/secrets', { secrets: 'fakeSecrets' });
+   * ```
+   *
+   * Multiple requests
+   *
+   * ```js
+   * I.mock('GET', ['/secrets', '/v2/secrets'], 403);
```
DavertMik: I :heart: this syntax!

DavertMik:
So far I am impressed! Ok, there are still few failures on Travis:

```
       #startMocking, #mock
         should be able to mock request with 200:
     Error: connect ECONNREFUSED 127.0.0.1:80
```

The sever runs on 8000 port. So I think this can be fixed somehow. 

---

Ok, here are the next steps.
So far this is good but PollyJS is coupled to Puppeteer while in theory it can be implemented for other helpers too. It also requires additional packages which may not be useful for people running tests with webdriverio. 

So the next goal would be to move PollyJS into its own helper, and connect to Puppeteer. 

1. Create a new class in `lib/helper/Polly.js` 
2. Move the implementation of PollyJS to it.
3. Use `_checkRequirements` to check for required libraries (polly adapter, and polly puppeteer) https://github.com/Codeception/CodeceptJS/blob/master/lib/helper/Puppeteer.js#L186
4. Connect to Puppeteer using `this.helpers.Puppeteer` inside a class. https://codecept.io/helpers#webdriver-example

Refer to https://github.com/Percona-Lab/codeceptjs-resemblehelper/blob/master/index.js#L79 as an example. This is a helper to do visual testing connecting to other helpers.

lib/helper/Polly.js
```diff
@@ -0,0 +1,39 @@
...
+    }
+  }
+
+  async _setupForPuppeteer(adapter) {
+    if (!PollyJS) {
+      throw new Error('Polly is not initialized properly!');
+    }
+
+    console.log('value of this.helpers => ', this.helpers);
+    const { page } = this.helpers.Puppeteer;
```
DavertMik:
```js
if (!this.helpers.Puppeteer) {
   throw new Error('Puppeteer is the only supported helper right now');
}
```

lib/helper/Puppeteer.js
```diff
@@ -1837,6 +1838,49 @@ class Puppeteer extends Helper {
...
   }
 
+  /**
+   * Start mocking network requests/responses
+   *
+   * ```js
+   * I.startMocking('Backend APIs');
+   * ```
+   */
+  async startMocking(title = 'Test') {
```
DavertMik: this method should be moved to Polly.js

test/helper/Puppeteer_test.js
```diff
@@ -276,6 +277,25 @@ describe('Puppeteer', function () {
       .then(() => I.seeNumberOfElements('#area1', 1)));
   });
 
+  describe('#startMocking, #mock', () => {
```
DavertMik: This test doesn't work, as axios is useless here.
We mock only requests inside browser, not inside NodeJS.

lib/helper/Puppeteer.js
VikalpP: We have different adapters based on helpers.
If I'll move this method to `Polly.js` then we'll have to guess current helper. 🤔

Well, because of currently we're providing for puppeteer only so it's good to keep separated.

DavertMik: Guessing current helper is fine. See comment above `if (this.helpers.Puppeteer)` should work. 
I think keeping function like this is harmful as it requires 3rd-party module without strictly requiring its installation. 

`require('@pollyjs/adapter-puppeteer');` will just fail when user will try to use without installing any Polly dependencies. And they will be surprised, as they were never asked to install such dependency.

Contrary, if we put all this logic to helpers/Polly,js a dependency will be explicitly listed. 
Also, this is fine if a helper will have adapters for different engines, including Puppeteer.

lib/helper/Polly.js
VikalpP: Done! Added into function `startMocking`.

test/helper/Puppeteer_test.js
VikalpP: Changed to acceptance test.

VikalpP:
**Demo of mocking requests:**


I_mockacceptancetest

VikalpP:
The reason for the test is failed mentioned in issue #1742.

lib/helper/Polly.js
```diff
@@ -0,0 +1,135 @@
...
+    }
+    const data = dataOrStatusCode;
+    return handler.intercept((_, res) => res.send(data));
+  }
+
+  /**
+   * Checks whether mocking is started.
+   */
+  _checkIfMockingStarted() {
+    if (!this.polly || !this.polly.server) {
```
DavertMik: Maybe if mocking is not started, can we start it automatically? 
If (!this.polly) this.startMocking() ?
This can be a nice addition

lib/helper/Polly.js
```diff
@@ -0,0 +1,135 @@
...
+   * I.mock('POST', '/secrets', { secrets: 'fakeSecrets' });
+   * ```
+   *
+   * Multiple requests
+   *
+   * ```js
+   * I.mock('GET', ['/secrets', '/v2/secrets'], 403);
+   * ```
+   */
+  async mock(method, oneOrMoreUrls, dataOrStatusCode) {
```
DavertMik: I just figured out that the name should be more informative. Like `mockRequest`

```js
I.mockRequest()
```
As there can probably be other things that can be mocked...

DavertMik:
Thanks @VikalpP So excited that the feature is almost ready! :sunglasses: 
Almost ready! I merged that pull request. And I have two proposals for the current one.

lib/helper/Polly.js
VikalpP: Yah!, that name of the method will be more meaningful 😉

lib/helper/Polly.js
VikalpP: You mean, whenever user directly use `I.mockRequest()` we'll start internally?

VikalpP: Ok, I've changed to start mocking automatically when `I.mockRequest()` used.

VikalpP:
**Demo of mocking requests:**


ezgif com-crop

VikalpP:
Found a little issue,

Peek 2019-07-05 17-42


I guess unIntercepted network requests are not being passed through after mocking is started.

VikalpP:
Solved by, setting up a default for all network request to let passed through.

peek solved

PR #1738, opened 2019-06-26
fix(multiple): async bootstrapAll must be called before creating chunks

Vorobeyko:
Async bootstrapAll invoked after calling done.
And when I need to override configuration parameters like "tests" I can't do that because bootstrapAll has an asynchronous function that codeceptjs wasn't waiting


VikalpP:
resolves #1745

DavertMik:
Thanks!

PR #1740, opened 2019-06-27
Puppeteer firefox to codeceptjs

ngadiyak:
Adding support of Puppeteer-firefox to Codeceptjs

https://github.com/GoogleChrome/puppeteer/tree/master/experimental/puppeteer-firefox

### DIFF
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -154,6 +154,7 @@ class Puppeteer extends Helper {
 
   _validateConfig(config) {
     const defaults = {
+      browser: 'chrome',
       waitForAction: 100,
       waitForTimeout: 1000,
       fullPageScreenshots: false,
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -45,6 +45,7 @@ const consoleLogStore = new Console();
  *
  * This helper should be configured in codecept.json or codecept.conf.js
  *
+ * * `browser`: (optional, default: chrome) - select browser
  * * `url`: base url of website to be tested
  * * `show`: (optional, default: false) - show Google Chrome window for debug.
  * * `restart`: (optional, default: true) - restart browser between tests.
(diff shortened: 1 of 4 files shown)


DavertMik:
:open_mouth: Wow! Does it really work?

ngadiyak:
i tried it locally (including parallel launch puppeteer-chrome and puppeteer-firefox) and it worked, but i didn't try all of API methods. this is experimental feature, @DavertMik you can try it locally too )

DavertMik:
According to their docs 90% of all methods already implemented. So I think we should pick it up. 
I will try to provide some suggestions for config API and I will merge it.

lib/helper/Puppeteer.js
```diff
@@ -45,6 +45,7 @@ const consoleLogStore = new Console();
  *
  * This helper should be configured in codecept.json or codecept.conf.js
  *
+ * * `browser`: (optional, default: chrome) - select browser
```
DavertMik: - select browser: chrome or firefox (experiemental, requires `puppeteer-firefox` package)

DavertMik: Btw, I don't see where default value for `browser` is acutally set

DavertMik:
@ngadiyak Please add default value for `browser` and I am merging! :+1:

ngadiyak:
https://github.com/ngadiyak/CodeceptJS/blob/dbc2f836cadef27fa4e98608418ebc807054d1c9/lib/helper/Puppeteer.js#L145

DavertMik:
I was thinking to put `browser: 'chrome'` somewhere here https://github.com/ngadiyak/CodeceptJS/blob/dbc2f836cadef27fa4e98608418ebc807054d1c9/lib/helper/Puppeteer.js#L156

ngadiyak:
@DavertMik done

DavertMik:
Ok, I will merge it when I will be releasing a new version

PR #1744, opened 2019-06-28
fix(plugin): stepByStepReport unique report folder

chimurai:
Currently the generated report folder names collide when tests have similar titles.
For example when you have test titles which start with "As a customer I want to..."

Different tests will end up being reported in same folder:
`<dir>/record_As_a_customer_I_want_1561724771/index.html`

With this fix a `md5 hash` is generated to avoid collisions in the reporting folders.
To calculate the md5 hash, `test.file` and `test.title` is used, just in case duplicate titles exist in a project.

After fix:
`<dir>/record_d43336466a7e65faf417d9ea70b3df70/index.html`

### DIFF
--- a/lib/plugin/stepByStepReport.js
+++ b/lib/plugin/stepByStepReport.js
@@ -7,7 +7,8 @@ const fs = require('fs');
 const path = require('path');
 const figures = require('figures');
 const colors = require('chalk');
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks!

PR #1756, opened 2019-07-05
Fix multiple run falsy parameter overriding

### LINKED ISSUE #1754: can not override config when run:multi
`codeceptjs run-multiple xxxx`

the multiple browser config should override original config

#### What do you get instead?

when the config value is a false value, it can not be override

Refer to code here: https://github.com/Codeception/CodeceptJS/blob/5fb0f6a787d93fc641d711fdf4cd74d3132829e0/lib/command/run-multiple.js#L200

when obj[key] === 0, the new value (for example, 1) will not be assigned.

**Possible fix**:
```js
if (obj[key] !== undefined) obj[key] = value;
```

Sample codecept.conf.js
```js
{
  config: {
    helpers: {
      WebDriver: {
        timeouts: 0
      }
    }
    multiple: {
      example: {
        browsers: [{
          timeouts: 1
        }]
      }
    }
  }
}
```

`codeceptjs run-multiple example` , the subprocess received args: `--override '{"helpers": {"WebDriver": {"timeout": 0}}}'` , but we expect `timeout: 1`

### Details

* CodeceptJS version: 2.1.5
* NodeJS Version: 8.15.1
* Operating System: macOS 10.14.5
* WebDriverIO version: 5.10.10

LukoyanovE:
Bug:
If value in config has falsy value, then multiple config does not overwrite original value.

- [x] made fix
- [x] wrote tests

Fixes #1754

### DIFF
--- a/lib/command/run-multiple.js
+++ b/lib/command/run-multiple.js
@@ -8,7 +8,7 @@ const runHook = require('../hooks');
 const event = require('../event');
 const collection = require('./run-multiple/collection');
 const clearString = require('../utils').clearString;
-const replaceValue = require('../utils').replaceValue;
+const replaceValueDeep = require('../utils').replaceValueDeep;
 
 const runner = path.join(__dirname, '/../../bin/codecept');
 let config;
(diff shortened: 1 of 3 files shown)


DavertMik:
Thank you!

PR #1760, opened 2019-07-08
refactor: Use puppeteer config for Polly

VikalpP:
**Problem:** No config `url` specific for Polly helper.
**Solution:** Use Puppeteer helper configs.

resolves #1766

PR #1768, opened 2019-07-09
fix: add preconditions to check for puppeteer page availability

### LINKED ISSUE #1767: Polly: Cannot read property 'setRequestInterception' of null

I am using  I.mockRequest('GET', '/api/users', 200); in my test. 

#### What do you get instead?

```bash
Cannot read property 'setRequestInterception' of null
  ypeError: Cannot read property 'setRequestInterception' of null
      at Polly._disconnectPuppeteer (node_modules\codeceptjs\lib\helper\Polly.js:164:16)
      at Polly.stopMocking (node_modules\codeceptjs\lib\helper\Polly.js:156:16)
      at Polly._after (node_modules\codeceptjs\lib\helper\Polly.js:56:16)
```

The call itself works, but I get the above error in the after hook, when _stopMocking() is called.

I think the cause is that the Puppeteer page has already been closed in the after hook and is no longer available (provided restart is set to true).

### Details

* CodeceptJS version: 2.2.0
* NodeJS Version: 10.x
* Operating System: Windows
* Polly

VikalpP:
Added some preconditions to check for puppeteer page is open or not.

resolves #1767

### DIFF
--- a/lib/helper/Polly.js
+++ b/lib/helper/Polly.js
@@ -67,9 +67,12 @@ class Polly extends Helper {
   // Connect Puppeteer helper to mock future requests.
   async _connectPuppeteer(title) {
     const adapter = require('@pollyjs/adapter-puppeteer');
-
     PollyJS.register(adapter);
+
     const { page } = this.helpers.Puppeteer;
+    if (!page) {
+      throw new Error('Looks like, there is no open tab');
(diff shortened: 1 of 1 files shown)


VikalpP:
This PR depends on #1760.

PR #1769, opened 2019-07-09
Testcafe fixes

hubidu:
Fixes and more helper method implementations


DavertMik:
Seems fine but tests are failing 😞

elukoyanov:
@DavertMik , why is it merged, if tests always fail?

PR #1773, opened 2019-07-12
fix(Puppeteer): checkRequirements failed when use only puppeteer-firefox

Vorobeyko:
When require Puppeteer with puppeteer-firefox and not install puppeteer - I obtains exception:

```
Could not load helper Puppeteer from module '...':
Cannot read property 'browser' of undefined
```

### DIFF
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -193,10 +193,12 @@ class Puppeteer extends Helper {
   }
 
   static _checkRequirements() {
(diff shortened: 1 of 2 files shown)


DavertMik:
I think `checkRequirements` won't work for both packages `puppeteer` and `puppeteer-firefox`. 
A proper solution would be to create `PuppeteerFirefox` helper which extends from Puppeteer.
Then we can easily override `_checkRequirements`.

PR #1778, opened 2019-07-16
fix: dynamic configs not reset after test

### LINKED ISSUE #1776: dynamic configurations are persisted for nested objects

Trying to change config dynamically for one test and have the helper config revert to the values defined in `codecept.conf.js` after the test. Specifically trying to set the user-agent via `args` array options for chrome, although the problem applies to any nested configuration objects.

#### What do you get instead?

Nested configurations are not reset to their original state (prior to using `Scenario().config()`). 

##### codecept.conf.js:

```js
exports.config = {
   tests: 'src/*.test.js',
   output: './output',
   helpers: {
      Puppeteer: {
         waitForNavigation: ['domcontentloaded', 'networkidle0'],
         chrome: {
            restart: true,
            defaultViewport: null,
            args: [
               '--disable-notifications',
               '--no-sandbox',
               '--disable-setuid-sandbox',
            ]
         }
      }
   },
   include: {
      I: './src/custom-steps.js',
   },
};
```

##### test:
```js
Scenario('test something with an android user-agent', async I => {
   /* ... */
}).config({
   chrome: {
      args: [
         '--disable-notifications',
         '--no-sandbox',
         '--disable-setuid-sandbox',
         `--user-agent="Mozilla/5.0 (Linux; Android 6.0; Nexus 5 Build/MRA58N) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/75.0.3770.100 Mobile Safari/537.36"`,
         `--window-size=600,960`,
      ],
   },
});
```

Every subsequent test after this one will open a 600x960 window with an android user-agent. I'm guessing all other chrome options are overwritten as well. Similar behaviour with other helpers.

### Details

* CodeceptJS version: 2.2.0
* NodeJS Version: v8.16.0 
* Operating System: osx 10.14.4

### Comments

This is related to this line:

https://github.com/Codeception/CodeceptJS/blob/38e04f564cbb3e954dc591636c9d2f26a3f8c920/lib/listener/config.js#L20

I'd open a pr  with something along the lines of:
```js
	helper._setConfig(deepMerge(deepClone(oldConfig), config));
```

cviejo:
Nested objects in dynamic config are also reset after a test or suite is done

Resolves #1776


DavertMik:
Thanks!

PR #1783, opened 2019-07-18
Update plugins.md

vitaly87:


DavertMik:
Thanks, this fix should have been applied to `lib/plugins/allure.js`
I will do that manually

PR #1784, opened 2019-07-18
fix: puppeteer scroll to locator

cviejo:
**Problem** 
With the puppeteer helper using `I.scrollTo()` on an element that's well below the viewport won't scroll to the element. The browser will remain in the upper area of the page instead.

**Solution**
The `elementCoordinates` returned by `_clickablePoint()` are **viewport** coordinates  (not window coordinates). Since we have already scrolled to the element via `_scrollIntoViewIfNeeded()`, we need scrollBy (not scrollTo) those coordinates to bring the element to the top the way webdriver does.

**Problem 2** 
When switching to an iframe, scroll is not correctly applied. 

**Solution 2**
Using `executeScript` instead of `this.page.evaluate`, since `executeScript` uses the current context (`_locate`, `_scrollIntoViewIfNeeded` and `_clickablePoint` use the it as well).

### DIFF
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -599,7 +599,7 @@ class Puppeteer extends Helper {
    * {{> scrollPageToTop }}
    */
   scrollPageToTop() {
-    return this.page.evaluate(() => {
+    return this.executeScript(() => {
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks!

PR #1790, opened 2019-07-19
Fix the --path option on gherkin:snippets

asselin:
PR https://github.com/Codeception/CodeceptJS/pull/1688 added support to direct new snippets to a specific file, but forgot to add the `--path` option to the command line parser.  Without this fix, `codeceptjs gherkin:snippets --path=foo` results in an error "error: unknown option `--path'"

### DIFF
--- a/bin/codecept.js
+++ b/bin/codecept.js
@@ -56,6 +56,7 @@ program.command('gherkin:snippets [path]')
   .description('Generate step defintions from steps.')
   .option('--dry-run', "don't save snippets to file")
   .option('-c, --config [file]', 'configuration file to be used')
+  .option('--path [file]', 'file in which to place the new snippets')
   .action(require('../lib/command/gherkin/snippets'));
 
 


asselin:
Please note that the Travis build fail wasn't caused by this PR, the WebDriver tests failed with a filename that is too long for some reason.

asselin:
@DavertMik Do you think maybe we could merge this fix in for the next release?  Also, if you have time, I'd love to hear your feedback on #1801 and #1803 :)

bin/codecept.js
```diff
@@ -56,6 +56,7 @@ program.command('gherkin:snippets [path]')
   .description('Generate step defintions from steps.')
   .option('--dry-run', "don't save snippets to file")
   .option('-c, --config [file]', 'configuration file to be used')
+  .option('--path [file]', 'file in which to place the new snippets')
```
DavertMik: can we rename this to `--output`?
Path not actually represents the meaning

PR #1791, opened 2019-07-20
feat: add 'run-workers' command

### LINKED ISSUE #1775: Parallel execution with workers
We need a better way to run tests in parallel with workers. With workers we can split tests by chunks, and aggregate output into the main process.

VikalpP:
## What it includes?
- Added command `run-workers` to run tests in [workers](https://nodejs.org/docs/latest-v12.x/api/worker_threads.html).

## Description: 
Using `run-workers` command, you will be able to run tests in parallel with workers.
Tests will be divided into groups to run in each worker, and the result from each worker will be combined and shown by each file.

### Other
resolves #1775

### DIFF
--- a/bin/codecept.js
+++ b/bin/codecept.js
@@ -138,6 +138,15 @@ program.command('run-multiple [suites...]')
 
   .action(require('../lib/command/run-multiple'));
 
+program.command('run-workers [otherOptions...]')
+  .description('Executes tests in workers')
+  .option('-n, --workers <value>', 'number of workers', 3)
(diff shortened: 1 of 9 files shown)


VikalpP:
@DavertMik , I don't know the exact reason behind failure of CI.
Can you please have look?🤔

DavertMik:
Thanks @VikalpP . Unfortunately, I can't include it in 2.2.1 release as this is a very huge feature. I will need some additional time to look and review it.

DavertMik:
Replaced in #1791

PR #1792, opened 2019-07-20
fix: Crash when using  xScenario & Scenario.skip with tag

### LINKED ISSUE #1751: Crash when using xScenario with tag
if we mark test with `.tag('@tag');` as `xScenario`, codeceptjs throw  `TypeError: Cannot read property 'tag' of undefined`

Example:
```js
xScenario('title', async (I) => {
    ...
}).tag('ISSUE-001');
```

VikalpP:
By returning the object of Scenario it continues to support other methods provided by Scenario.

resolves #1751

### DIFF
--- a/lib/ui.js
+++ b/lib/ui.js
@@ -138,7 +138,7 @@ module.exports = function (suite) {
      * Pending test case.
      */
     context.xScenario = context.Scenario.skip = function (title) {
-      context.Scenario(title, {});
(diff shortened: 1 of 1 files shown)


VikalpP:
Fix of failing test is #1768.
So, this PR depends on it.

DavertMik:
Thanks!

PR #1794, opened 2019-07-22
 Fix:Data().Scenario().injectDependencies() is not a function

andrerleao:
Fix: "TypeError: scenario.injectDependencies is not a function" on use Data().Scenario().injectDependencies() or Data().Scenario().inject();


DavertMik:
Thanks!

PR #1797, opened 2019-07-22
fix: use filename for uploadFile in webdriver

VikalpP:
#### Problem
Using file content in base64 as an argument to `uploadFile`.
` Error: ENAMETOOLONG: name too long, open... `

#### Solution
Use filename as an argument to `uploadFile`.
#### Why?
Checkout official [doc](https://webdriver.io/blog/2019/06/25/file-upload.html)


VikalpP:
Fix for the failed tests is #1768.
So this PR depends on that PR.

PR #1799, opened 2019-07-23
fix: missing support object definition in steps.d.ts

### LINKED ISSUE #1795: In v2.2.0 broken generating the typescript definitions for PageObject (and another supports)
I trying generate typescript definitions 

#### What do you get instead?
For Support object not generated definitions

And broken `container.support()`. It not return support object

It happened after https://github.com/Codeception/CodeceptJS/pull/1701 

### Details

* CodeceptJS version: 2.2.0
* NodeJS Version: 
* Operating System:
* Protractor || WebDriverIO || Nightmare version (if related)
* Configuration file:

```js
exports.config = {
  tests: './*_test.js',
  output: './output',
  helpers: {
    Puppeteer: {
      url: 'http://localhost'
    }
  },
  include: {
    I: './steps_file.js',
    public: './pages/public.js'
  },
  bootstrap: null,
  mocha: {},
  name: 'test-support-def'
}
```

johnyb:
Provide more proxy traps to inspect support object, so generating the definitions file works fine, again.

Fixes #1795

### DIFF
--- a/lib/command/definitions.js
+++ b/lib/command/definitions.js
@@ -182,15 +182,16 @@ module.exports = function (genPath, options) {
   const supports = container.support(); // return all support objects
   const exportPageObjects = [];
   const callbackParams = [];
-  for (const name in supports) {
-    if (name === 'I') continue;
-    callbackParams.push(`${name}:CodeceptJS.${name}`);
+  // See #1795 and #1799, there's no obvious way to use for-in with Proxies
+  Reflect.ownKeys(supports).forEach((name) => {
+    if (name === 'I') return;
+    callbackParams.push(`${String(name)}:CodeceptJS.${String(name)}`);
     const pageObject = supports[name];
     const pageMethods = addAllMethodsInObject(pageObject, {}, []);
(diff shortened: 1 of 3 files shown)


Vorobeyko:
@johnyb I cant get pageObject from container, but in v2.1.5. it works correctly: 
```
import { container } from 'codeceptjs'
container.support('pageObject')
```

DavertMik:
@Vorobeyko this PR is about TS definitions, not accessing container.
the previous PR changed the behavior to use proxy objects. But this approach can have its pitfalls, we need to investigate it more.

DavertMik:
Thanks!

johnyb:
> @johnyb I cant get pageObject from container, but in v2.1.5. it works correctly:
> 
> ```
> import { container } from 'codeceptjs'
> container.support('pageObject')
> ```
> 
> and when I use
> 
> ```
> const { pageObject } = inject()
> ```

That's indeed unrelated. The first case should even work as expected. I'll take a closer look into the second example using global inject ()

PR #1801, opened 2019-07-24
Fix for snippets command with a .feature file that has special characters

asselin:
Fix issue when running snippets command with a .feature file that has steps with Gherkin special characters (any of  { } ( ) / )
Instead of creating strings, create regular expressions that will match
This fixes the problem where the generated snippets won't match the steps from the original .feature file


asselin:
Note: the CircleCI failure was not due to this PR.  CircleCI failed with a timeout.

DavertMik:
Thank you!

PR #1802, opened 2019-07-25
update mocha to 6

elukoyanov:
- [x] update mocha package version
- [x] remove co-mocha (does not support mocha 6) and generators in tests


package.json
```diff
@@ -54,8 +54,9 @@
     "js-beautify": "^1.9.1",
     "lodash.merge": "^4.6.1",
     "mkdirp": "^0.5.1",
-    "mocha": "^4.1.0",
-    "mocha-junit-reporter": "^1.22.0",
+    "mocha": "6",
```
DavertMik: maybe `"mocha": "^6"` ?

VikalpP: Or try the latest 
```suggestion
    "mocha": "^6.2.0",
```

elukoyanov: techincally, there are no difference between
`6` and `^6` 
`^6.2.0` will not install all under 6.2.0

I don't think that versions equal and higher than 6.2.0 will be removed from package servers.

Minor and patch versions updates are not include breaking changes, so it will not broken.
Only new non-breaking features and bug fixed.

But ok, I'll change to today last `^6.2.0`

elukoyanov: And Codeceptjs does not use lock files.

elukoyanov:
@DavertMik , I look into `mochaFactory`.
There are using of `mocha` with `loadFiles` and `Runner.uncaught` overriding.

In the `ui.js` file `tags`, `inject`, and others are added to `Mocha.Test`, `Mocha.Suite` and others.

Maybe, analogues `CodeceptMocha`/`CodeceptTest`/`CodeceptSuite` should be created as class extension instead of monkey patching Mocha classes themself?

Can do in this or other PR.

lib/plugin/screenshotOnFail.js
```diff
@@ -69,7 +69,11 @@ module.exports = function (config) {
     }
   }
 
-  if (Codeceptjs.config.get().mocha) options.reportDir = Codeceptjs.config.get().mocha.reporterOptions.reportDir;
+  if (Codeceptjs.container.mocha()) {
+    options.reportDir = Codeceptjs.container.mocha().options.reporterOptions
+      && Codeceptjs.container.mocha().options.reporterOptions.reportDir;
```
elukoyanov: Mocha 4 used `options = options || {};` and changed options object directly.

Mocha 6 does use `options = utils.assign({}, mocharc, options || {});` and changes only local copy of options.
So it's not correctly to check param in config, when it does not changes and mocha have actual config in itslef

DavertMik: Ok, then remove this

DavertMik:
> Maybe, analogues CodeceptMocha/CodeceptTest/CodeceptSuite should be created as class extension instead of monkey patching Mocha classes themself?

A good idea. That would be really nice! 

Filenames can be like:
  mocha/test.js
  mocha/suite.js

DavertMik:
I tried your branch locally. Works really great. Waiting to see this merged!

DavertMik:
@elukoyanov are we ready?

elukoyanov:
It can be merged.
I'll make mocha class variation in other PR

DavertMik:
Thank you!

PR #1804, opened 2019-07-25
fix(allure): Failed xml parse when assert msg contains ansi symbol

Vorobeyko:


package.json
```diff
@@ -60,7 +60,8 @@
     "promise-retry": "^1.1.1",
     "requireg": "^0.1.8",
     "resq": "^1.5.0",
-    "sprintf-js": "^1.1.1"
+    "sprintf-js": "^1.1.1",
+    "strip-ansi": "^5.2.0"
```
DavertMik: can this be replaced with a 1 regex string?

Vorobeyko: I dont know) 
maybe it (https://github.com/chalk/ansi-regex/blob/master/index.js#L4)  will help?

Vorobeyko: I fixed it

DavertMik:
Thanks!

PR #1805, opened 2019-07-26
Puppeteer: Speed up tests with large forms

hubidu:
...by making Puppeteers .type(...) delay configurable (option pressKeyDelay)


DavertMik:
Thanks!

PR #1807, opened 2019-07-29
Remove default script timeout config.

### LINKED ISSUE #1529: [Possible BUG] MJSONWP protocol requires type and ms" and you sent {"script": 1000}
Hello,

I'm using Codeceptjs with webdriverio and browserstack (automate) and I’m facing an issue related with MJSONWP endpoint to fetch the timeouts and makes my test suite to crash.

#### What do you get instead?

```bash
Error: Parameters were incorrect. We wanted "MJSONWP protocol requires type and ms" and you sent {"script":1000}
```

```bash
-- FAILURES:

  1) Commit valid answer
       "before each" hook: codeceptjs.before for "Commit a valid answer for an explicit commit block":
     Parameters were incorrect. We wanted "MJSONWP protocol requires type and ms" and you sent {"script":1000}
  rror: Parameters were incorrect. We wanted "MJSONWP protocol requires type and ms" and you sent {"script":1000}
      at getErrorFromResponseBody (node_modules/webdriver/build/utils.js:336:12)
      at Request._callback (node_modules/webdriver/build/request.js:116:64)
      at Request.self.callback (node_modules/request/request.js:185:22)
      at Request.EventEmitter.emit (domain.js:441:20)
      at Request.<anonymous> (node_modules/request/request.js:1161:10)
      at Request.EventEmitter.emit (domain.js:441:20)
      at IncomingMessage.<anonymous> (node_modules/request/request.js:1083:12)
```

### Details

* CodeceptJS version: 2.0.4
* NodeJS Version: 10
* Operating System: Mac OS 10.13
* WebDriverIO 5.6.3
* Using inside browserstack: appium_version: 1.9.1 (Tried with 1.6.5, 1.7.1, etc...)**

* Configuration file:
What I found interesting is that Inside my config file (codeceptjs config file) I do not specify any timeout option.

```js
exports.config = {
  tests: './test_codecept/*.js',
  output: './test_codecept/output',
  helpers: {
    WebDriver: {
      host: 'hub.browserstack.com',
      path: '/wd/hub',
      url: 'http://WEBSITE:8080/renderer',
      user: 'xx',
      key: 'xx',
      browser: 'iphone',
      desiredCapabilities: {
        'os_version' : '11',
        'device' : 'iPhone 8',
        'real_mobile' : 'true',
        'browserstack.local' : 'true',
        'browserstack.debug' : 'true',
        'browserstack.networkLogs' : 'true',
        'browserstack.appium_version' : '1.9.1',
        'browserstack.user' : 'xx',
        'browserstack.key' : 'xx'
      }
    }
  }
}
```

After failing a few times, I tried to specify a timeout option inside the helper like mention in the [w3c WebDriver spec](https://www.w3.org/TR/webdriver/#set-timeouts)
and didn't work neither, appeared that I was sending an empty object to the `sessionId/timeouts`.

```js
timeouts: {
  type: 'script',
  ms: 1000
}
```

After that I tried to setup the timeouts like the error message propose: "type" and "ms", didn't work neither.

```js
timeouts: {
  type: 'script',
  ms: 1000
}
```

Thanks for reading

pablopaul:
On BrowserStack and current latest Appium available `v1.9.1` the following occurs when trying to run a test on a real iOS device:

`Parameters were incorrect. We wanted "MJSONWP protocol requires type and ms" and you sent {"script":1000}`.

I think this is a bug related to Appium on iOS. BrowserStack does not give any ETA when a newer Appium version will be available for iOS on their platform.

In the WebDriver spec the default script timeout is set to `30000 ms`, I think it is safe to fallback to that value. See https://w3c.github.io/webdriver/#timeouts.

Fix #1529

### DIFF
--- a/lib/helper/WebDriver.js
+++ b/lib/helper/WebDriver.js
@@ -393,9 +393,6 @@ class WebDriver extends Helper {
       keepCookies: false,
       keepBrowserState: false,
       deprecationWarnings: false,
-      timeouts: {
-        script: 1000, // ms
-      },
     };
 
 


valishah:
@DavertMik: Can we confirm this PR and merge?

pablopaul:
I digged deeper.

Locally Appium 1.9.1 with CodeceptJS running a test on iOS works fine, this seems to be a BrowserStack **and** Appium specific problem.

I contacted their support but I do not expect a fix in short time.

pablopaul:
@DavertMik @valishah another option here would to fall back to usage of 

```js
browser.timeouts( 'script', 30000 );
browser.timeouts('implicit',5000);
…
```

if CodeceptJS detects run on iOS & BrowserStack - Should we build a workaround like this?

valishah:
Hi @DavertMik: Any thoughts on @pablopaul suggestion?

DavertMik:
This is a breaking change so I will leave this till CodeceptJS 3.0

nkrishna79:
@davert @ Peter - I think we really need to get this fix out. At the moment, I am not able to run iOS tests which kind of is a blocker for me to scale my testing framework. Can you please think about it and release it?

valishah:
Thanks for all the support @DavertMik @pablopaul and others 👍

PR #1812, opened 2019-07-30
Fix TestCafe SwitchTo

KadoBOT:
Testcafe tests were failing sometimes due to a promise being passed to `this.t.switchToIframe`. By adding the await, we make sure that the promise is resolved before passing it.


DavertMik:
Thanks @KadoBOT 
You are battle testing TestCafe helper, that's nice. All feedback and pull requests on using this helper are welcome.

PR #1819, opened 2019-07-31
add geoLocation methods

PeterNgTr:
Introduce two methods:
- getGeoLocation()
- setGeoLocation()

### DIFF
--- a/docs/webapi/getGeoLocation.mustache
+++ b/docs/webapi/getGeoLocation.mustache
@@ -1,0 +1,6 @@
+Return the current geo location 
+
+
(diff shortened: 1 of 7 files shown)


DavertMik:
Maybe to keep general style we should use `grabGeolocation` ?

PeterNgTr:
> Maybe to keep general style we should use `grabGeolocation` ?

Sure, that's nice. I would make the change!

DavertMik:
Accepted!

DavertMik:
Thanks!

PR #1822, opened 2019-08-01
Fire event on I.say() and add return val to step.passed

hubidu:
- Can we fire events for I.say()
- and add the return value of commands (e. g. from REST requests) to step.passed?

Specifically useful for codepress.

### DIFF
--- a/lib/actor.js
+++ b/lib/actor.js
@@ -38,7 +38,7 @@ module.exports = function (obj) {
 
   // add print comment method`
   obj.say = (msg, color = 'cyan') => recorder.add(`say ${msg}`, () => {
-    event.emit('step.say', msg);
+    event.emit('step.comment', msg);
     output.say(msg, `${color}`);
   });
   obj.retry = retryStep;
--- a/lib/actor.js
+++ b/lib/actor.js
@@ -37,7 +37,10 @@ module.exports = function (obj) {
     });
 
   // add print comment method`
-  obj.say = (msg, color = 'cyan') => recorder.add(`say ${msg}`, () => output.say(msg, `${color}`));
+  obj.say = (msg, color = 'cyan') => recorder.add(`say ${msg}`, () => {
+    event.emit('step.say', msg);
+    output.say(msg, `${color}`);
+  });
   obj.retry = retryStep;
 
   return obj;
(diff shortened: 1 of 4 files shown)


Vorobeyko:
Can you write tests?

DavertMik:
Seems fine. The only concern I have here is that this is an event for one command only. 
Maybe we can use it like `step.comment`? If so, we can also send steps in gherkin with this event too.

hubidu:
Would be nice to get events for the gherkin steps too. I can add this too.

DavertMik:
@hubidu yeah, let's do that
we can add this event for gherkin features and (maybe?) metasteps. 
It should be called `step.comment` instead of `step.say`

hubidu:
@DavertMik I changed the event name to step.comment and added gherkin support. However I could only implement it in a way that a gherkin step 'step.comment' event is fired AFTER the first codecept step 'step.started' event.
Do you see a way to improve this?

lib/actor.js
```diff
@@ -37,7 +37,10 @@ module.exports = function (obj) {
     });
 
   // add print comment method`
-  obj.say = (msg, color = 'cyan') => recorder.add(`say ${msg}`, () => output.say(msg, `${color}`));
+  obj.say = (msg, color = 'cyan') => recorder.add(`say ${msg}`, () => {
+    event.emit('step.comment', msg);
```
DavertMik:
```suggestion
    event.emit(event.step.comment, msg);
```
`event.step.comment` should be declared as const inside `events.js`

lib/reporter/cli.js
```diff
@@ -79,6 +79,7 @@ class Cli extends Base {
             printMetaStep(metaStep.metaStep);
           }
           currentMetaStep.unshift(metaStep.toString());
+          event.emit('step.comment', step.metaStep.toString() + (step.metaStep.comment ? step.metaStep.comment : ''));
```
DavertMik:
```suggestion
          event.emit(event.step.comment, step.metaStep.toString() + (step.metaStep.comment ? step.metaStep.comment : ''));
```

DavertMik:
> However I could only implement it in a way that a gherkin step 'step.comment' event is fired AFTER the first codecept step 'step.started' event.

Technically, I don't see why they should go in a different order, so it should not be a problem. Maybe what we need is to remove `event.step.started` for gherkin steps at all, replacing it with step.commet. But it will also require some other updates, in allure plugin, for instance

DavertMik:
Ok, I will finialize this

PR #1824, opened 2019-08-02
Fix Parameter in Scenario Example

pablopaul:
Remove `adminPage` as this parameter is not needed and probably will break the scenario if not present.


DavertMik:
Thanks!

PR #1826, opened 2019-08-05
Update attachFile doc in TestCafe helper doc

quekshuy:
Used documentation from WebdriverIO helper doc to fill in the blanks/update TestCafe's own `attachFile` documentation.


DavertMik:
Ok, I see the bug but the fix should be quite different. I will apply it now

quekshuy:
Thank you.

PR #1829, opened 2019-08-05
fixed element position bug

AnotherAnkor:
Without this change you need everytime and everywhere print "()" brackets.

### DIFF
--- a/lib/locator.js
+++ b/lib/locator.js
@@ -133,7 +133,7 @@ class Locator {
     if (position === 0) {
       throw new Error('0 is not valid element position. XPath expects first element to have index 1');
     }
-    const xpath = sprintf('%s[position()=%s]', this.toXPath(), position);
+    const xpath = sprintf('(%s)[position()=%s]', this.toXPath(), position);
     return new Locator({ xpath });
   }
 


Vorobeyko:
Can you write the tests, please?

AnotherAnkor:
> Can you write the tests, please?

I never did it before. But I can. It's strongly recomended? What framework should I use?

Vorobeyko:
> I never did it before. But I can. It's strongly recomended?
You must to write tests in this file: https://github.com/Codeception/CodeceptJS/blob/master/test/unit/locator_test.js
You can look at existing tests, and write by analogy

>  What framework should I use?
We are use mocha.

AnotherAnkor:
As I can see you have the test which covers this code.
Please, take a look: https://github.com/Codeception/CodeceptJS/blob/192e3e01cf0e23b82d8526febe03cb67ad40e08e/test/unit/locator_test.js#L141

DavertMik:
Ok, seems fine. Thanks!

nitschSB:
@AnotherAnkor and @DavertMik 

this change breaks a lot of selectors! Now there are selectors you may not use anywhere.
For example my universal td selector
```
locate('td')
        .withText(text)
        .at(columnNumber)
        .inside(
          this.tableRows(tableId)
            .withText(text)
            .at(_rowNumber)
        )
```

Puppeteer throws this exception: "Uncaught DOMException: Failed to execute ... is not a valid XPath expression."

**Reason:** Brackets aren't allowed in a "ancestor"-selector.
`//*[ancestor::(//*)[1]]`

By the way `removePrefix` does not work for those kind of selectors as well.

AnotherAnkor:
@nitschsb, sorry about it.
But I never used so terrible code like you showed.
May be more skilled guys will says something about your example?

PR #1842, opened 2019-08-12
fix run-multiple with typescript

z4o4z:
Currently, it is impossible to run tests in parallel with TS, as regex finds only `.js` files. 

I've just updated the regex.


DavertMik:
Thanks! Failing tests unrelated, so I'm merging.

PR #1844, opened 2019-08-13
Update Puppeteer.md

dhuddell:


DavertMik:
Thanks, this fix should have been applied to `lib/helper/Puppeteer.js`, so I copied and pasted it manually :)

dhuddell:
My bad! I'll get that next time.

PR #1846, opened 2019-08-14
Missing enabled true option to activate auto start/stop selenium

NSV:


DavertMik:
Thanks! I always forget about this option :)

PR #1848, opened 2019-08-15
Add NodeJS scope for EventEmitter

elukoyanov:
Error `Cannot find name 'EventEmitter'.ts(2304)` ocures without fix

was added in https://github.com/Codeception/CodeceptJS/pull/1800


DavertMik:
:+1:

PR #1852, opened 2019-08-16
feat: mock requests in webdriver

radhey1851:
#### What does this do? 
- Extends support for mocking requests using PollyJs to WebDriver.
- Renames `Polly` helper to `MockRequest`.


radhey1851:
@DavertMik Can you pls look at this when possible? Thanks

VikalpP:
I see function duplications like, `mockRequest`, `getRouteHandler`, `stopMocking`.
Can you abstract that code?

Suggestion: Use function like,
```js
function _getPollyObject {
    if (this.helpers.WebDriver) {
      return window.polly;
    }
    if (this.helpers.Puppeteer) {
      return this.polly;
    }
}
```

And I guess rest most of the code are same except `Setup`. 😉

VikalpP:
Also, Functions - `appendBaseUrl`, `joinUrl`, `shouldAppendBaseUrl`, `trimUrl`
Those are util function can be found at `../utils` 😅

radhey1851:
> I see function duplications like, `mockRequest`, `getRouteHandler`, `stopMocking`.
> Can you abstract that code?
> 
> Suggestion: Use function like,
> 
> ```js
> function _getPollyObject {
>     if (this.helpers.WebDriver) {
>       return window.polly;
>     }
>     if (this.helpers.Puppeteer) {
>       return this.polly;
>     }
> }
> ```
> 
> And I guess rest most of the code are same except `Setup`.

Yes, the code does the same thing as it does with Puppeteer. Bu everything has to run in the context of the browser, i.e, window global with contains the fetch and XMLHttpRequest methods. And with web driver, All the actions have to run within B

lib/helper/Polly.js
```diff
@@ -134,8 +160,20 @@ class Polly extends Helper {
...
+      oneOrMoreUrls,
+      dataOrStatusCode,
+      additionalData || undefined,
+      this.options.url || webDriverIOConfigUrl,
+    );
+  }
+
   _checkIfMockingStarted() {
-    return this.polly && this.polly.server;
+    return (this.polly && this.polly.server) || this.browser;
```
DavertMik: sorry, I didn't understand why `this.browser` should indicate that mocking started

lib/helper/Polly.js
```diff
@@ -134,8 +160,20 @@ class Polly extends Helper {
     return handler.intercept((_, res) => res.send(data));
   }
 
+  _mockRequestForWebDriver(method, oneOrMoreUrls, dataOrStatusCode, additionalData) {
+    const webDriverIOConfigUrl = this.helpers.WebDriver && this.helpers.WebDriver.options.url;
+    this.browser.execute(
```
DavertMik: this looks like async operation, so probably should be made an async function or at least return a promise

lib/helper/Polly.js
```diff
@@ -168,13 +206,19 @@ class Polly extends Helper {
...
   async stopMocking() {
     if (!this._checkIfMockingStarted()) return;
-    await this._disconnectPuppeteer();
 
-    const { polly } = this;
-    if (!polly) return;
-    await polly.flush();
-    await polly.stop();
-    delete this.polly;
+    if (this.currentDriver === 'Puppeteer') {
```
DavertMik: I think you can make _disconnectPuppeteer and _disconnectWebDriver methods similarly to `_connect*`.

lib/helper/PollyWebDriverExt.js
```diff
@@ -0,0 +1,109 @@
+
+const pollyWebDriver = {
+  setup: (title) => {
+    const fetchUrls = [
+      'https://cdn.jsdelivr.net/npm/@pollyjs/core@2.6.0/dist/umd/pollyjs-core.min.js',
```
DavertMik:
```suggestion
      'https://cdn.jsdelivr.net/npm/@pollyjs/core@2/dist/umd/pollyjs-core.min.js',
```

lib/helper/Polly.js
radhey1851: Because I am creating am storing the browser object when I start mocking with web Driver, so I thought I would use that to see if mocking started, as I don't have access to the Polly object. I see that it isn't very obvious to understand.

lib/helper/Polly.js
radhey1851: yeah, that would be better

lib/helper/PollyWebDriverExt.js
radhey1851: Thanks!

lib/helper/Polly.js
radhey1851: > So I changed it to check and return if polly object has been created in browser context.

lib/helper/Polly.js
```diff
@@ -134,15 +161,32 @@ class Polly extends Helper {
...
+      this.options.url || webDriverIOConfigUrl,
+    );
+  }
+
+  async _checkIfMockingStarted() {
+    if (this.currentDriver === 'Puppeteer') {
+      return (this.polly && this.polly.server);
+    }
+    if (this.currentDriver === 'WebDriver') {
+      return (this.browser && this.browser.execute(pollyWebDriver.isPollyObjectInitialized));
```
radhey1851: @DavertMik, does this work. I am checking in the browser context.

lib/helper/Polly.js
```diff
@@ -65,15 +68,20 @@ class Polly extends Helper {
   }
 
   // Start mocking network requests/responses
-  async _startMocking(title = 'Test') {
-    if (!(this.helpers && this.helpers.Puppeteer)) {
-      throw new Error('Puppeteer is the only supported helper right now');
+  async startMocking(title = 'Test') {
```
DavertMik: Requires documentation for this method, if it goes to public API. 
Please mention that `mockRequest` automatically starts request mocking

lib/helper/Polly.js
```diff
@@ -65,15 +68,20 @@ class Polly extends Helper {
...
+    if (this.helpers.Puppeteer) {
+      await this._connectPuppeteer(title);
+    } else if (this.helpers.WebDriver) {
+      await this._connectWebDriver(title);
     }
-    await this._connectPuppeteer(title);
   }
 
   // Connect Puppeteer helper to mock future requests.
   async _connectPuppeteer(title) {
```
DavertMik: Requires a simple docblock, as this method can be executed from other helpers. And it will be available in documentation...

lib/helper/Polly.js
```diff
@@ -92,6 +100,15 @@ class Polly extends Helper {
     });
   }
 
+  // Connect WebDriver helper to mock future requests.
+  async _connectWebDriver(title) {
```
DavertMik: Requires a simple docblock, as this method can be executed from other helpers. And it will be available in documentation...

lib/helper/Polly.js
radhey1851: Added doc. Is this proper?

DavertMik:
Thanks, I'm merging!

PR #1862, opened 2019-08-27
Ensure configured user agent and/or window size is applied to all pages.

martomo:
Prior only initial page would have configuration applied, new pages/tabs or (incognito) sessions would have default values.


DavertMik:
Thanks, this really was an issue, even to me, but I didn't have a chance to fix that!

Issue #1866, opened 2019-08-30
@AnotherAnkor and @DavertMik


johnyb:
@AnotherAnkor and @DavertMik 

this change (#1829) breaks a lot of selectors! Now there are selectors you may not use anywhere.
For example my universal td selector
```
locate('td')
        .withText(text)
        .at(columnNumber)
        .inside(
          this.tableRows(tableId)
            .withText(text)
            .at(_rowNumber)
        )
```

Puppeteer throws this exception: "Uncaught DOMException: Failed to execute ... is not a valid XPath expression."

**Reason:** Brackets aren't allowed in a "ancestor"-selector.
`//*[ancestor::(//*)[1]]`

By the way `removePrefix` does not work for those kind of selectors as well.

_Originally posted by @nitschSB in https://github.com/Codeception/CodeceptJS/pull/1829#issuecomment-523952794_

AnotherAnkor:
May be you are right.
But without examples I can't understand what's happened.
Please, provide more examples.

AnotherAnkor:
Ok. Sorry about it.
But I never used so terrible code like you showed. Usually I use xpath. And in this case my change work perfect.
You can revert my change if need.

AnotherAnkor:
You are right.
Please, revert my changes.

nitschSB:
I thought a little about my example and your more or less offensive criticism
> But I never used so terrible code like you showed.

I have to admit that my example is not pretty but it used to work. After your changes I may refactor my selectors to something like
```
locate('#tableId')
  .find('tr')
  .at(1)
  .find('td')
  .at(2)
  .withText('specificText')
```
which is better readable and leads to the xpath `((.//*[./@id = 'tableId']//tr)[position()=1]//td)[position()=2][contains(., 'specificText')]`

I also thought about your statement
> Usually I use xpath.

So why do you use the Locator Builder API at all?
I think this somehow is contrary to the Locator Builder API which "provides a fluent builder to compose custom locators in JavaScript". In my opinion you should either use xpath or the Locator Builder. It is possible to mix both and you may need it for some special edge cases, but if you do so you need to know the xpath syntax and the Locator Builder API.

johnyb:
Exactly my thoughts and also the solution I implemented to workaround this issue. Besides from actually breaking the locator API, it at least pointed out some badly designed locators. Out of 670 tests, this "only" broke 4 of them ;)

[...]

PR #1870, opened 2019-09-02
Improve handling of xpath locators with round brackets

nitschSB:
After #1829 the locator functions `.at(i)`, `.first()` and `.last()` create xpath locators with round brackets. This may lead to "is not a valid XPath expression" exceptions (see #1866).

The changes of this PR
 * "allows xpath locators to start with round brackets" for the `isXPath()`-method
 * "prevents round brackets for nested selectors" by refactoring the old six times used  `removePrefix((new Locator(locator, 'css')).toXPath()))` expression to a new method called `convertToSubSelector(locator)` which also throws a specific error if the locator `isXPathStartingWithRoundBrackets`

### DIFF
--- a/lib/locator.js
+++ b/lib/locator.js
@@ -111,17 +111,17 @@ class Locator {
   }
 
   find(locator) {
-    const xpath = sprintf('%s//%s', this.toXPath(), removePrefix((new Locator(locator, 'css')).toXPath()));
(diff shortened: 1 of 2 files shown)


DavertMik:
Thanks!

nitschSB:
This improvement is needed due to #1829 which is controversial. You should have a look at my latest comment in #1866 . We need a decision that is recorded in the documentation.

PR #1877, opened 2019-09-05
Use alias command

hannut91:
Use i instead of install.
Use -D instead of --save-dev.

See also
  - https://docs.npmjs.com/cli/install


DavertMik:
Sorry, I can't merge this

1. I don't see why it makes a difference
2. Code formatting changes in files not related to PR

PR #1886, opened 2019-09-11
add a option to run-workers to preserve suites

nitschSB:
distribution of tests from a suite to different workers leads to multiple execution of BeforeSuite

the new option uses a different approach for the assignment of tests to workers that preserves suites


DavertMik:
Wow, I was thinking of it few days ago! Thank you!

PR #1888, opened 2019-09-12
Update visual.md

gopilaxman87:


DavertMik:
Ok, I got the point but it should be rephrased

PR #1896, opened 2019-09-23
fix(pageObject): Not work when PageObject as class

Vorobeyko:
PageObject file:
```js
class PageObject {
  type (s) {
    console.log(s)
  }

  purgeDomains (s) {
    console.log('purgeDomains')
    console.log(s)
  }
}

module.exports = new PageObject()
```

Test file:
```js
Feature('asdas')

Scenario('qw', async (I, page) => {
  await page.type('asda')
})
```

Result:
```sh
asdas --
  ✖ qw in 2ms

-- FAILURES:

  1) asdas
       qw:
     page.type is not a function
  
```


DavertMik:
Thank you!

PR #1904, opened 2019-09-24
Fix #1903 make a copy of page objects config per step although it is deprecated

### LINKED ISSUE #1903: Page objects from shared library cannot be loaded
I want to load my custom shared library `codeceptjs-base` with page objects to use in several projects.

#### What do you get instead?
The tests fail with the error message that the page object file cannot be loaded.
In the error message i see, that every include gets the codecept config directory prepended.
Somehow this works for the first test fine, but the second test will get this error.

```bash
-- FAILURES:

  1) Test Prepration
       login as admin:
     Could not include object login from module '/Users/lkimmel/Coding/.../codeceptjs-base/lib/pages/login'
Cannot find module '/Users/lkimmel/Coding/.../codeceptjs-base/lib/pages/login'
Require stack:
- /Users/lkimmel/Coding/.../node_modules/codeceptjs/lib/container.js
- /Users/lkimmel/Coding/.../node_modules/codeceptjs/lib/codecept.js
- /Users/lkimmel/Coding/.../node_modules/codeceptjs/bin/codecept.js
```

### Details

* CodeceptJS version: 2.3.2
* NodeJS Version: 12.10.0
* Operating System: MacOs
* Configuration file:

```js
exports.config = {
	...
	include: {
		I: "./steps_file.es6",
		adminNavigation: "codeceptjs-base/lib/pages/admin-navigation",
		...
	},
	...
};
```

### Fix
After I analyzed the code flow, i saw that in `step.js` that `loadSupportObjects` always sets the full path to every config entry under `include`.
That method is declared as deprecated but it is still the way it is done in CodeceptJS.
I provide a quick fix as pull request.

GerroDen:
This creates a deep copy of the `include` part from configuration.

### DIFF
--- a/lib/step.js
+++ b/lib/step.js
@@ -180,7 +180,7 @@ function loadSupportObjects() {
   const config = require('./config');
   const path = require('path');
   support = {
-    ...config.get('include', {})
+    ...config.get('include', {}),
   };
   if (support) {
     for (const name in support) {
--- a/lib/step.js
+++ b/lib/step.js
@@ -179,7 +179,9 @@ function detectMetaStep(stack) {
 function loadSupportObjects() {
   const config = require('./config');
   const path = require('path');
-  support = {...config.get('include', {})};
+  support = {
+    ...config.get('include', {})
+  };
   if (support) {
     for (const name in support) {
       const file = support[name];
--- a/lib/step.js
+++ b/lib/step.js
@@ -179,7 +179,7 @@ function detectMetaStep(stack) {
 function loadSupportObjects() {
   const config = require('./config');
   const path = require('path');
-  support = config.get('include', {});
+  support = {...config.get('include', {})};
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks. I didn't understand how it is related to your issue. Did that fix it?
This part of code used not to load page objects but to analyze stacktraces to detect usage of page objects. Technically it should not affect the way tests are executed

DavertMik:
I don't mind merging it. It does not do any harm :) 
But I don't think this actually fixes issue or how it is related to a problem

GerroDen:
I found that part harmful, as it overrides the paths of every page object file by appending the codecept_dir. After fixing this line, i have no issue with loading page objects from external libraries.

DavertMik:
Ok, thanks!

PR #1908, opened 2019-09-25
Update Node version to 12.10 to support new CodeceptJS features

### LINKED ISSUE #1900: Docker image uses old version of node
It seems that CodeceptJS docker image depends on Node 10.14.0, therefore "run-workers" functionality cannot be used there. Could this be updated or do you use current LTS versions for your images?

triforcely:
Fixes #1900 . Uses NodeJS 12.10.0 because it is currently the latest version.


DavertMik:
Thanks. Forgot to update. Now it will include `run-workers!`

PR #1909, opened 2019-09-25
Introduced config hooks

DavertMik:
Config hooks allow to set functions to update complex configs on the fly.

For instance, you can use them to enable/disable headless modes, window sizes, etc.


PeterNgTr:
@DavertMik do we need the doc for this?

PR #1917, opened 2019-10-01
Added missing commas in autoLogin example

Lagunov-PRO:


DavertMik:
Thanks, I had to update it myself as it requires to change `lib/plugin/autoLogin.js`

PR #1921, opened 2019-10-02
update the docker image to include run-workers

PeterNgTr:
- include the `run-workers`
- update README


DavertMik:
@PeterNgTr it's not documented how to use run-workers

PeterNgTr:
Oops. Forgot to add it.

PR #1923, opened 2019-10-03
Fix broken 'Multiple  Execution'-link

gosua:
# Issue
The configuration page contains a dead link to the multiple execution details.
# Solution
Adjusted link to point to the correct page containing the multiple  execution details.


DavertMik:
Thank you!

Issue #1867, opened 2019-09-02
Element getSize() and getLocation


rebajz:
In codecept (all helpers as WebDriver, Protractor, etc) missing methods for work with element size and location. (see/dontSee and grab).

PeterNgTr:
@rebajz could you try to test this and close this issue accordingly?

PR #1926, opened 2019-10-04
implement grabElementBoundingRect for wdio and puppeteer helpers

PeterNgTr:
https://github.com/Codeception/CodeceptJS/issues/1867
this would add `grabElementBoundingRect`method for WDIO and Puppeteer.

### DIFF
--- a/docs/webapi/grabElementBoundingRect.mustache
+++ b/docs/webapi/grabElementBoundingRect.mustache
@@ -24,7 +24,7 @@ value = {x: 1, y: 2, width: 200, height: 200}
 const value = await I.grabElementBoundingRect('h3', 'width');
 ```
 
-If multiple elements found returns an array of texts.
+If multiple elements found returns an object.
 
 @param {string|object} locator element located by CSS|XPath|strict locator.
 @param {string} elementSize the width or height of the given element.
--- a/docs/webapi/grabElementBoundingRect.mustache
+++ b/docs/webapi/grabElementBoundingRect.mustache
@@ -1,10 +1,10 @@
-Grab the width and height of given locator.
+Grab the width, height, location of given locator.
 Provide `width` or `height`as second param to get your desired prop.
 Resumes test execution, so **should be used inside an async function with `await`** operator.
 
 ```js
-const value = await I.grabElementSize('h3');
-const value = await I.grabElementSize('h3', 'width');
+const value = await I.grabElementBoundingRect('h3');
+const value = await I.grabElementBoundingRect('h3', 'width');
 ```
 
 If multiple elements found returns an array of texts.
--- a/docs/webapi/grabElementBoundingRect.mustache
+++ b/docs/webapi/grabElementBoundingRect.mustache
@@ -3,7 +3,24 @@ Provide `width` or `height`as second param to get your desired prop.
 Resumes test execution, so **should be used inside an async function with `await`** operator.
 
 ```js
+/**
+* return an object
+* one found element
+* { x: 226.5, y: 89, width: 527, height: 220 }
+*
+* multiple found elements
+* {
+*  '0': { x: 296, y: 322.5, width: 387, height: 34 },
+*  '1': { x: 350.734375, y: 393, width: 127.40625, height: 36 },
+*  '2': { x: 489.765625, y: 393, width: 142.5, height: 36 }
+* }
+*/
 const value = await I.grabElementBoundingRect('h3');
+/**
+* return only the width
+* value = 200
+*/
+value = {x: 1, y: 2, width: 200, height: 200}
 const value = await I.grabElementBoundingRect('h3', 'width');
 ```
 
@@ -11,4 +28,4 @@ If multiple elements found returns an array of texts.
 
 @param {string|object} locator element located by CSS|XPath|strict locator.
 @param {string} elementSize the width or height of the given element.
-@returns {Promise<array>} Element size
+@returns {object|number} Element bounding rectangle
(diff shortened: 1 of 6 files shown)


lib/helper/WebDriver.js
```diff
@@ -2055,6 +2055,17 @@ class WebDriver extends Helper {
...
 
+  /**
+   * {{> grabElementSize }}
+   *
+   */
+  async grabElementSize(locator, prop) {
+    const res = await this._locate(locator, true);
+    assertElementExists(res, locator);
+
+    return forEachAsync(res, async el => el.getSize(prop));
```
Vorobeyko: Webdriver return only width and height, but Puppeteer method return x, y, width and height.
Can you make the same return values?

PeterNgTr: Sure, not a problem :)

PeterNgTr: @Vorobeyko thanks bro for your review! I pushed the changes!

lib/helper/Puppeteer.js
```diff
@@ -1985,6 +1985,31 @@ class Puppeteer extends Helper {
...
+    const result = [];
+    els = els.filter(v => v);
+
+    if (prop) {
+      els.forEach((element) => {
+        result.push(element[prop]);
+      });
+    } else {
+      els.forEach((element) => {
+        delete element.x;
```
Vorobeyko: Oh no) 
x and y can be use! My command used this property)) 
I thought you make x and y in webdriverio)))

Vorobeyko: X and y coordinates are important))

PeterNgTr: haha. I'm thinking of renaming this method as `grabElementInfo` to convey better meaning. Cause we want to return the element size and location(x,y). Do you have any suggestions?

Vorobeyko: Maybe grabElementBoundingRect ?

PeterNgTr: sounds nice. I give a go for it.

PeterNgTr: @Vorobeyko done. not both wdio and puppeteer return location, weight and height. :)

lib/helper/Puppeteer.js
```diff
@@ -1985,6 +1985,29 @@ class Puppeteer extends Helper {
   async grabDataFromPerformanceTiming() {
     return perfTiming;
   }
+
+  /**
+   * {{> grabElementSize }}
```
Vorobeyko: grabElementBoundingRect

lib/helper/Puppeteer.js
```diff
@@ -1985,6 +1985,29 @@ class Puppeteer extends Helper {
...
+   *
+   */
+  async grabElementBoundingRect(locator, prop) {
+    let els = await this._locate(locator);
+    els = await Promise.all(els.map(el => el.boundingBox()));
+    const result = [];
+    els = els.filter(v => v);
```
Vorobeyko: what this line do?

lib/helper/WebDriver.js
```diff
@@ -2055,6 +2055,34 @@ class WebDriver extends Helper {
...
+      arr.forEach((element) => {
+        result.push(element[prop]);
+      });
+    } else {
+      arr.forEach((element) => {
+        result.push(element);
+      });
+    }
+
+    return result;
```
Vorobeyko: You can use `find` for find element with prop and immediately return the result, like
```
if (prop) {
      return [arr.find((element) => {...});]
    }
```

If prop is null - return els.

lib/helper/Puppeteer.js
PeterNgTr: this line to remove empty items, null, etc.

docs/webapi/grabElementBoundingRect.mustache
```diff
@@ -0,0 +1,14 @@
...
+```js
+const value = await I.grabElementBoundingRect('h3');
+const value = await I.grabElementBoundingRect('h3', 'width');
+```
+
+If multiple elements found returns an array of texts.
+
+@param {string|object} locator element located by CSS|XPath|strict locator.
+@param {string} elementSize the width or height of the given element.
+@returns {Promise<array>} Element size
```
DavertMik: The structure is not defined here. I would like to see which object is returned

docs/webapi/grabElementBoundingRect.mustache
```diff
@@ -0,0 +1,14 @@
+Grab the width, height, location of given locator.
+Provide `width` or `height`as second param to get your desired prop.
+Resumes test execution, so **should be used inside an async function with `await`** operator.
+
+```js
+const value = await I.grabElementBoundingRect('h3');
```
DavertMik: please add example on using `value`, show which properties it includes.

docs/webapi/grabElementBoundingRect.mustache
```diff
@@ -0,0 +1,14 @@
...
+const value = await I.grabElementBoundingRect('h3', 'width');
+```
+
+If multiple elements found returns an array of texts.
```
DavertMik: I know that we have this feature for `grabTextFrom` but I think it was a design mistake. 
Please make this return an object, not an array, to make the behavior more predictable.

DavertMik:
Sorry, @PeterNgTr I was unclear in the last statement:

> I know that we have this feature for grabTextFrom but I think it was a design mistake.
Please make this return an object, not an array, to make the behavior more predictable.

What I actually meant, this method should always return an object for a ONE element. Not an array.
So this method should take the first matched element and returns an object. If you need a specific object in a list - use `locate().at()`. 

If you don't mind I will merge this PR but change this method to return rect only for the first element

PeterNgTr:
@davertk please go with your suggestion.

DavertMik:
Great! Doing so. If you may need multipleRect this should be additional method... However, I hope you won't need it :smile:

PR #1932, opened 2019-10-08
docs: clarified sentences in angular documentation

lucasmelin:
Modified the wording of the Angular documentation.

Tried to keep the original intent of the documentation, while adding missing particles and rearranging sentences in order to improve readability.


DavertMik:
Thank you!

PR #1936, opened 2019-10-09
TestCafe port's management removed (left on TestCafe itself)

### LINKED ISSUE #1934: How does run-workers work?

I'm tryin to execute command:
> npx codeceptjs run-workers 2 [4, 8]

and I expect to get as much chrome browsers as workers count

#### What do you get instead?

1-2 chrome opens and that's it, no more browser windows (cpu and mem are enough atm)

### Details

* CodeceptJS version: 2.3.1
* NodeJS Version: 12.10.0
* Operating System: MacOS Mojave
* TestCafe: 1.5.0
* Configuration file:

```js
exports.config = {
	tests: './*_test.js',
	output: './output',
	helpers: {
		TestCafe: {
			url: 'myUrl',
			browser: 'chrome',
			show: true
		},
		MailSlurp: {
			require: '@codeceptjs/mailslurp-helper',
			apiKey: 'validApiKey',
			timeout: 60000
		},
	},
	include: {
		"mainPage": "../pages/mainPage.js",
		"osagoPage": "../pages/osagoPage.js",
		"userClicksForgotNumber": "./steps/userClicksForgotNumber.js",
		"userSavesStateByEmail": "./steps/userSavesStateByEmail.js"
	},
	bootstrap: null,
	mocha: {},
	name: 'productName',
	translation: 'ru-RU',
	timeout: 10000
};
```
**UPD 1.**
When I've executed command:
> npx codeceptjs run-workers 2 --verbose

There were spawned multiple errors all containing this: 
> [1] Error | Error: The specified 1341 port is already in use by another program.

**UPD 2.**
I've digged into code, first codeceptjs's, then testCafe's and, boy, what do I see:

```js
async function getValidPort(port) {
    if (port) {
        const isFree = await endpointUtils.isFreePort(port);
        if (!isFree)
            throw new runtime_1.GeneralError(types_1.RUNTIME_ERRORS.portIsNotFree, port);
    }
    else
        port = await endpointUtils.getFreePort();
    return port;
}
```
TestCafe can manage ports by its own, so, why do we need this kind of management: 
```js
// in ctor
this.iteration = 1;
...
// and then
async _startBrowser() {
    this.dummyTestcafeFile = createTestFile(global.output_dir); // create a dummy test file to get hold of the test controller

    this.iteration += 2; // Use different ports for each test run
    // @ts-ignore
    this.testcafe = await createTestCafe('localhost', 1338 + this.iteration, 1339 + this.iteration);
```

Why can't we just pass null to a ports in previous line?

**UPD 3.**
I've made a fork and [applied](https://github.com/orihomie/CodeceptJS/commit/a7668420f4cacf9fd5dadc7b59f560425db8ad0c) my suggestions and that works

orihomie:
Resolves #1934

### DIFF
--- a/lib/helper/TestCafe.js
+++ b/lib/helper/TestCafe.js
@@ -86,7 +86,6 @@ class TestCafe extends Helper {
   constructor(config) {
     super(config);
 
-    this.iteration = 1;
     this.testcafe = undefined; // testcafe instance
     this.t = undefined; // testcafe test controller
     this.dummyTestcafeFile; // generated testcafe test file
@@ -136,7 +135,7 @@ class TestCafe extends Helper {
 
     this.iteration += 2; // Use different ports for each test run
     // @ts-ignore
-    this.testcafe = await createTestCafe('localhost', 1338 + this.iteration, 1339 + this.iteration);
+    this.testcafe = await createTestCafe('localhost', null, null);
 
     this.debugSection('_before', 'Starting testcafe browser...');
 


DavertMik:
Thanks for the patch. Probably it was implemented for earlier versions of TestCafe

orihomie:
Thanks for not delaying :)

PR #1937, opened 2019-10-09
introduce Lighthouse to puppeteer helper

PeterNgTr:


DavertMik:
I think we need more features to get this integration done. I would like to see assertions for lighthouse returned values. Here is an example of what I would like to have:

```js
I.amOnPage('/my-page');
I.calculateLighthouseScore([
      'first-meaningful-paint',
      'speed-index',
      'first-cpu-idle',
      'interactive',
]); // connects to current session, perform audists. A list of audit can be default or overridden.
I.seeAuditScoreIsOk('speed-index', '70'); // checks that speed-index is greater than 70 
```

PeterNgTr:
I think those scores are better to do assertions:

Performance
Accessibility
Best Practices
SEO

DavertMik:
@PeterNgTr 

I would really love to see Lighthouse integration in CodeceptJS but here are the reasons I can't accept this PR:

1. It introduces extra dependency which won't be needed for the majority of users.
2. It doesn't contain any usage examples. So If you are not familiar with Lighthouse you will have no idea how to use this.

It would be nice if you could release this as a standalone helper, bringing more documentation and usage examples. 

> I think, we just grab the values and let the users do the assertions they want based on those scores.

I would also love to see some built-in assertions. Sure, users can do their custom assertions on their own, but the idea for helpers is to provide the most common patterns implemented. So `seePerformanceIsOk` (and others) would be a good addition.

DavertMik:
This is how to use Lighthouse with Puppeteer https://t.co/rvbmD0JIVF

PR #1941, opened 2019-10-10
fix(locator): Fixed speed degradation after #1306

Vorobeyko:


nabeelharri:
The change improved the tests speed, but it's still too slow. I believe there is still something else affecting the performance.  BTW I was using old version of CodecepsJS after upgrading to the latest I had huge performance drop .

PR #1948, opened 2019-10-12
Fix documentation of puppeteerCoverage plugin

lemnis:
Docs were talking about `outputDir` while in reality `coverageDir` was used.


DavertMik:
Thanks!

PR #1952, opened 2019-10-13
Improve handling of pressing keys for Puppeteer and WebDriver

martomo:
* Changed `pressKey` method to resolve issues and extend functionality.
  * Did not properly recognize `'Meta'` (or `'Command'`) as modifier key.
  * Right modifier keys did not work in WebDriver using JsonWireProtocol.
  * `'Shift'` + `<key>` combination would not reflect actual keyboard behavior.
  * Respect sequence with multiple modifier keys passed to `pressKey`.
  * Add support to automatic change operation modifier key based on operating system.
* Updated `pressKey` documentation to show a proper list of available key names which are available **both** in [Puppeteer](https://github.com/GoogleChrome/puppeteer/blob/v1.20.0/lib/USKeyboardLayout.js) and [WebDriver](https://www.w3.org/TR/webdriver/#keyboard-actions).
* Added `pressKeyDown` and `pressKeyUp` methods with similar functionality as updated `pressKey` method to simplify pressing of modifier key in combination with user operation.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -114,17 +114,20 @@ Please try to add corresponding testcase to runner or unit.
 
 Documentation is stored in `/docs` directory in markdown format.
 
-**Documentation for helpers is a part of a source code**. Whenever you need to update docs for a helper do it inside a .js file. Then run
+**Documentation for helpers is a part of a source code**.
+Whenever you need to update docs for a helper do it inside a .js file.
+
+In order to generate new documentation from source code run the following command with [Robo](https://robo.li/):
 
 ```
-npm run docs
+robo docs:helpers
 ```
 
 To update markdown documentation. Shared documentation for helpers are located in `docs/webapi/*.mustache`. Inside a docblock those files can be included like this:
 
 ```js
   /**
-   * {{> ../webapi/click }}
+   * {{> click }}
    */
   click() {
     // ...
--- a/docs/helpers/Appium.md
+++ b/docs/helpers/Appium.md
@@ -225,9 +225,9 @@ I.dontSee('Login', '.nav'); // no login inside .nav element
 Verifies that the specified checkbox is not checked.
 
 ```js
-I.dontSeeeCheckboxIsChedcked('#agree'); // located by ID
-I.dontSeeeCheckboxIsChedcked('I agree to terms'); // located by label
-I.dontSeeeCheckboxIsChedcked('agree'); // located by name
+I.dontSeeCheckboxIsChecked('#agree'); // located by ID
+I.dontSeeCheckboxIsChecked('I agree to terms'); // located by label
+I.dontSeeCheckboxIsChecked('agree'); // located by name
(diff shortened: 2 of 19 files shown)


DavertMik:
A very good job taking on polishing keyboard handling! Thank you.

I will merge this PR but I have few notes for the future contributions.

* Puppeteer.md / WebDriver.md was generated, so you don't need to update them. Because all methods have similar behavior, documentation for them is stored in `docs/webapi/methodName.md` there you should have added `pressKeyUp` and `pressKeyDown`
* Please minimize the pull request to make it simple and focused around one feature. 

Thank you!

martomo:
Just to confirm, next time I just should document in **docs/webapi/methodName.mustache** (along with any additional comments in code that do not belong in template) and there is no need to manually generate updated documentation with Robo? If so, how or when is the documentation updated if not by the author that is pushing documentation updates?

DavertMik:
> Just to confirm, next time I just should document in docs/webapi/methodName.mustache (along with any additional comments in code that do not belong in template) and there is no need to manually generate updated documentation with Robo

Yes

> If so, how or when is the documentation updated if not by the author that is pushing documentation updates?

Documentation is actually updated on a release. I run `robo docs` when releasing a new version.

martomo:
Great, good to know. Thanks for the quick review and response!

DavertMik:
@martomo after looking at the code once more I decided to make some changes. For instance, we do not have booleans as parameters in methods (as they are not readable in console output) so I would remove `detectOS` parameter from a function and move that to helper config. 

I think when you write a test you should use one style, so if you detect os in one place you will probably do that in another... Or use dynamic configuration per test which is also possible.

DavertMik:
Also, `detectOS` will be `false` by default for compatibility reasons

martomo:
@DavertMik Indeed, it makes sense to have `detectOS` set to `false` as default to ensure current tests do not break. However I am not sure about `detectOS` being a helper (or per test) level configuration.

In case the application being tested uses custom shortcut handlers, it might be that within the same test case (run on macOS) that a `'Control'` (mapped to `'Meta'`) key sequence is followed by a key sequence using `'Control'` which requires actual `'Control'` and this would not be achievable with helper (or per test) level configuration.

martomo:
To be more specific, here is an example test scenario which would not work with helper (or per test) level configuration.
```js
Scenario('clear inputs in application dialog with shortcut', (I) => {
  // Map to ['Meta', 'Shift', '1'], being an application shortcut to show dialog
  I.pressKey(['Control', 'Shift', '1']); // detectOS = `true`
  I.fillField('field1', 'Hello');
  I.fillField('field2', 'World');
  // Use actual ['Control', 'C'], application shortcut which clears inputs in dialog
  I.pressKey(['Control', 'C']); // detectOS = `false`
  I.dontSeeInField('field1', 'Hello');
  I.dontSeeInField('field2', 'World');
});
```

PR #1953, opened 2019-10-14
Aria tests

DavertMik:
Optimization for searching via `aria-labelledby` attribute
Replaces #1941


nitschSB:
The fix doesn't solve the performance issues for me.

PR #1956, opened 2019-10-14
2.3.3 release

DavertMik:

PR #1957, opened 2019-10-15
Introduce key alias for pressing modifier depending on operating system

martomo:
* Replacing `detectOS` option with key alias (e.g. `'CommandOrControl'`) which will press `'Control'` or `'Meta'` depending on operating system.
* Additional aliases for `'Meta'` (`'Cmd'`, `'OS'`, `'Super'`) and properly detect `'AltGraph'`.

### DIFF
--- a/docs/webapi/pressKeyDown.mustache
+++ b/docs/webapi/pressKeyDown.mustache
@@ -9,4 +9,3 @@ I.pressKeyUp('Control');
 ```
 
 @param {string} key name of key to press down.
-@param {boolean} detectOS (optional, `true` by default) change operation modifier key based on operating system.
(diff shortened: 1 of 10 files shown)


martomo:
@DavertMik This should address the matter regarding `detectOS` option as pointed out in https://github.com/Codeception/CodeceptJS/pull/1952#issuecomment-541873117, by replacing it with a key alias (`CommandOrControl`).

Would it be possible to include this in the next release (https://github.com/Codeception/CodeceptJS/pull/1956#pullrequestreview-301641099)?
By including this solution in the next release, there would be no need to support `detectOS` option (in helper/test config) for any future releases.

DavertMik:
Thanks! This patch is a way better. However, now I need to merge it on top of a previous one :)

martomo:
Give me a couple minutes and I'll create a new branch based from 2.3.3-release.

martomo:
Closing in favour of #1960, which can be merged directly to release 2.3.3 branch.

PR #1960, opened 2019-10-15
Introduce key alias for pressing modifier depending on operating system

martomo:
* Replacing `detectOS` option with key alias (e.g. `'CommandOrControl'`) which will press `'Control'` or `'Meta'` depending on operating system.
* Additional aliases for `'Meta'` (`'Cmd'`, `'OS'`, `'Super'`) and properly detect `'AltGraph'`.


DavertMik:
Wow, you are awesome! Thank you very much!

PR #1962, opened 2019-10-15
introduce authentication for puppeteer

PeterNgTr:
This PR extends the config to provide the basic authentication when opening the base URL, this will make the URL looks cleaner 😄 

#### Example 5: Target URL with provided basic authentication
 ```js
 {
   helpers: {
       Puppeteer : {
         url: 'http://localhost',
         basicAuth: { username: 'postman', password: 'password' },
         show: true
       }
     }
  }
 ```


lib/helper/Puppeteer.js
```diff
@@ -127,6 +128,22 @@ const consoleLogStore = new Console();
...
  * ```
  *
+* #### Example #5: Target URL with provided basic authentication
+ *
+ * ```js
+ * {
+ *    helpers: {
+ *      Puppeteer : {
+ *        url: 'http://localhost',
+ *        basicAuth: 'username, password',
```
martomo: What about passing an object instead of string?

Benefits of this would be no accidental trimming of white spaces as result of string manipulation, or handling of exceptions when password contains `,` and the I guess possibility of being able to use `secret` for password.
```js
basicAuth: {
  username: 'postman',
  password: secret('withtrailingspace ')
}

Comment (PeterNgTr) [reply lib/helper/Puppeteer.js]: I'm not sure the `secret` thingy, I'll check it out. Thanks man!

Comment (PeterNgTr) [reply lib/helper/Puppeteer.js]: with password, one possibility is passing it as env var. :)

Comment (PeterNgTr) [reply lib/helper/Puppeteer.js]: isn't it the case `secret` just help you mask printed string on log, with this config, you can still see it as plain?

Comment (martomo) [reply lib/helper/Puppeteer.js]: Yes, you are right, from the looks of it `secret` is shown as ['*****'](https://github.com/Codeception/CodeceptJS/blob/39225aebe7b03a44becfe0b58388a7ee9a90a92c/lib/step.js#L84-L85) in step log output.
With your proposed changes it actually removed the hardcoded `user:password@` auth string from the URL, as it is handled in the background by calling the [`authenticate`](https://github.com/GoogleChrome/puppeteer/blob/v1.20.0/docs/api.md#pageauthenticatecredentials) method directly on `Page` instance (which shows no log output as far I can recall).

**Comment by DavertMik** [comment]
I like the idea, can we take it to WebDriver as well?

I also agree that username password should be specified as object

File: lib/helper/Puppeteer.js [hunk 2]
```diff
@@ -513,6 +530,11 @@ class Puppeteer extends Helper {
     if (!(/^\w+\:\/\//.test(url))) {
       url = this.options.url + url;
     }
+
+    if (this.config.basicAuth) {
+      this.page.authenticate(this.config.basicAuth);
```
DavertMik: 1. what if we are not on the base page? If I am running `I.amOnPage('https://github.com')`
2. what if we already authenticated?

PeterNgTr: 1. make a check to see if the current is the same with base page.
2. introduce `isAuthenticated` var for checking if we are already authenticated.

lib/helper/WebDriver.js
```diff
@@ -701,6 +721,10 @@ class WebDriver extends Helper {
    *
    */
   amOnPage(url) {
+    if (this.config.basicAuth) {
```
DavertMik: what if url is `/` ?

```js
I.amOnPage('/');
```

test/helper/Puppeteer_test.js
```diff
@@ -19,6 +14,48 @@ let page;
 let FS;
 const siteUrl = TestHelper.siteUrl();
 
+describe('Puppeteer - BasicAuth', () => {
+  before(() => {
+    global.codecept_dir = path.join(__dirname, '/../data');
+
+    I = new Puppeteer({
+      url: 'https://postman-echo.com',
```
DavertMik: better to use local url

PeterNgTr: @DavertMik but local url doesn't have basic auth?

DavertMik:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones added
============
- test/helper/WebDriver_test.js  4
- test/helper/Puppeteer_test.js  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/CodeceptJS/CodeceptJS/pullRequest?prid=4321469&bid=14709064)

DavertMik:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones added
============
- test/helper/Puppeteer_test.js  3
- test/helper/WebDriver_test.js  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/DavertMik/CodeceptJS/pullRequest?prid=4321469&bid=14709064)

Issue #1439, opened 2019-01-13
Puppeteer - cannot switch to the iframe without an ID


iljao:
I'm trying to wait for an element inside an iframe using puppeteer helper.
The thing is that the iframe doesn't have an ID attribute.
I tried both switchTo and within, but waitForElement() fails with timeout in both cases.
It looks like switching is not happening, because I can locate elements outside the target iframe after switchTo(...).
It works with webdriver and nightmare though, so the issue is with puppeteer I guess.

Has anybody encountered this issue?

### Details

* CodeceptJS version: 2.0.2
* NodeJS Version: 10.13.0
* Operating System: win10

hekard2l:
I ran into StaleElementExceptions with WebDriver, so decided to take a look at this again.

This doesn't work if you don't have names on your frames. It will never check by url.
https://github.com/Codeception/CodeceptJS/blob/d28a3e0aca5f8370aa349af209928a94b6734903/lib/helper/Puppeteer.js#L1835

I think, this function should look by url if it can't find anything by name.

franklsm1:
Adding the name field to my iframe allowed me to switch between my iframe and the main page with no trouble. Before the URL of the iframe was changing behind the scenes so I could never enter into the iframe at the right time, but with name it does not matter I can call `waitForElement` calls again.

GPITRTD:
Please fix this issue, it is open for a long time. It is currently impossible to do any tests in an iFrame with Puppeteer. Both functions: within() and switchTo() are not working (even with an iframe id). As stated before after the switch function, it is possible to find elements outside the iFrame.

**Details**

- CodeceptJS version: v2.3.0
- Node version: v10.16.0
- Operating System: Windows 10

Example:

```
 I.click('button');
 I.waitForVisible('#iframe',8);
 I.switchTo('#iframe');
 I.waitForVisible('#element-outside-iframe', 5);
 I.waitForVisible('#element-in-iframe', 5);
```

Output:

```
"before each" hook: Before for "Check velden zoekscherm":
element (#element-in-iframe) still not visible after 5 sec
waiting for selector "#element-in-iframe" failed: timeout 5000ms exceeded
```

johnyb:
I actually ran into this with a snippet like:

```
    I.waitForElement('.mail-detail-frame');
    within({ frame: '.mail-detail-frame' }, () => {
        I.waitForText('This mail has no content', 3);
    });
```

which uses I.switchTo internally. At least in my situations this patch fixes the problem.

PR #1974, opened 2019-10-18
fix(Puppeteer): simplify switchTo iframe by selector

johnyb:
switchTo did not work as expected in some situations, because the lookup code did not consider every possibility. However, given a ElementHandle, the corresponding frame can be fetched directly using Puppeteer API. This allows to remove a lot of code and should still work as expected.

I tried to add failing tests, but could not come up with a minimal example. I'll try a little harder to come up with something, but wanted to get this PR out of the door.

Fixes: #1439

### DIFF
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -1915,17 +1915,10 @@ class Puppeteer extends Helper {
     // iframe by selector
     const els = await this._locate(locator);
     assertElementExists(els, locator);
+    const contentFrame = await els[0].contentFrame();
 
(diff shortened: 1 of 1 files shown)


johnyb:
At least in my test base, this change actually broke something else, which did work before. Needs investigation if this is related.

DavertMik:
Thanks, I think Puppeteer API evolved since the time we wrote that implementation

DavertMik:
> I tried to add failing tests, but could not come up with a minimal example. I'll try a little harder to come up with something, but wanted to get this PR out of the door.

According to issue a minimal example would be an iframe without id. Did you try this?

PR #1984, opened 2019-10-21
Allow to parm reporter for TestCafe

orihomie:
Allowing to parm reporter through config


lib/helper/TestCafe.js
```diff
@@ -148,7 +148,7 @@ class TestCafe extends Helper {
       .screenshots(global.output_dir, !this.options.disableScreenshots)
       // .video(global.output_dir) // TODO Make this configurable
       .browsers(this.options.show ? this.options.browser : `${this.options.browser}:headless`)
-      .reporter('minimal')
+      .reporter(this.options.reporter ? this.options.reporter : 'minimal')
```
PeterNgTr: would be nice to add documentation to specify which type of reports you can input. https://devexpress.github.io/testcafe/documentation/using-testcafe/common-concepts/reporters.html

PR #1987, opened 2019-10-23
Fixing typo in email testing guide

EricTendian:
MailSlurp was spelled wrong in one place, and the link associated with their guide went to a 404 page. This fixes those issues.


DavertMik:
Thank you! Probably they updated the website so a link got broken.

Issue #1976, opened 2019-10-21
install codeceptjs versi 2.3.3 get respond error "cannot find module './utils"


Kurniawan22:
I am install codeceptjs new version and follow step by step install "https://codecept.io/quickstart"
but get respond error " cannot find module './utils'".

step by step :

- npm init -y
- yarn add codeceptjs puppeteer --save-dev
- npx npx codeceptjs init
   geeting respond = cannot find module './utils'

maybe codeceptjs versi 2,3,3 is corrupt?

sky172839465:
I have the same problem, you can install `parse-function@5.2.11` for a workaround.

hfrijlink:
Thanks @sky172839465 , that worked for me!

pablopaul:
@sky172839465 how did you figured out that this is the dependency which requires the not available module?

sky172839465:
@pablopaul my functional test's error log told me which line throw error

```diff
Error: Cannot find module './utils'
     at Function.Module._resolveFilename (internal/modules/cjs/loader.js:580:15)
     at Function.Module._load (internal/modules/cjs/loader.js:506:25)
     at Module.require (internal/modules/cjs/loader.js:636:17)
     at require (internal/modules/cjs/helpers.js:20:18)
-    at Object.<anonymous> (/xxx/node_modules/parse-function/dist/cjs/index.js:1:731)
     at Module._compile (internal/modules/cjs/loader.js:688:30)
     at Object.Module._extensions..js (internal/modules/cjs/loader.js:699:10)
     at Module.load (internal/modules/cjs/loader.js:598:32)
     at tryModuleLoad (internal/modules/cjs/loader.js:537:12)
     at Function.Module._load (internal/modules/cjs/loader.js:529:3)
```

I have been [report an issue][] to parse-function's repo, hope this issue can be fixed ASAP.

[report an issue]: https://github.com/tunnckoCore/parse-function/issues/202

Kurniawan22:
hi @sky172839465 thanks for you solutions.

tunnckoCore:
Install latest `parse-function` version :)

I'm in times of migration to monorepo for all my stuff and there can be bugs, so frequent updates may be needed for some time. Sorry for the problems :)

PR #1990, opened 2019-10-25
upgrade parse-function to 5.4.x

sky172839465:
`parse-function` has been fix the [issue][] that cause `Cannot find module './utils'` error, so we don't need to keep version in 5.2.x 🎉

[issue]: https://github.com/Codeception/CodeceptJS/issues/1976

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "mocha": "^6.2.0",
     "mocha-junit-reporter": "^1.23.1",
     "ms": "^2.1.2",
-    "parse-function": "5.2.11",
+    "parse-function": "^5.4.3",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -59,7 +59,7 @@
     "mocha": "^6.2.0",
     "mocha-junit-reporter": "^1.23.1",
     "ms": "^2.1.2",
-    "parse-function": "5.2.11",
+    "parse-function": "^5.4.3",
```
PeterNgTr: I think we should use `"parse-function": "5.4.3"` to avoid updating to latest version whenever it is.

sky172839465: Agree! let me update the version.

Issue #1863, opened 2019-08-27
Puppeteer / waitForText method: doesn't works with xpath selector


ko4ergins:
> Provide console output if related. Use `--verbose` mode for more details.

```bash
Text "GitHub" was not found on page after 3 sec
Evaluation failed: TypeError: $XPath is not a function
    at eval (eval at waitForPredicatePageFunction (:2:21), <anonymous>:4:22)
    at eval (eval at waitForPredicatePageFunction (:2:21), <anonymous>:7:11)
    at onRaf (__puppeteer_evaluation_script__:56:33)
    at pollRaf (__puppeteer_evaluation_script__:48:5)
    at waitForPredicatePageFunction (__puppeteer_evaluation_script__:7:18)
```

> Provide test source code if related

```js
Feature('User Story');

Scenario('QA', (I) => {
  I.amOnPage('https://github.com');
  I.waitForText('GitHub', 3, '//*[@class="h1-mktg"]');

});
```

### Details

* CodeceptJS version: 2.3.0
* NodeJS Version: v11.0.0
* Operating System: Linux Mint
* Puppeteer: 1.19.0
![DeepinScrot-2354](https://user-images.githubusercontent.com/19324105/63789558-4bc69800-c900-11e9-8214-f563b2f5aef4.png)

grzegorzprzybylo:
@DavertMik could you take a look on this if you got a second? not sure how important it is (priority) but in some cases XPath selectors are very useful imo. Thanks!

PR #2002, opened 2019-11-04
fix XPath is not a function for waitForText

Heavik:
fix addresses this issue https://github.com/Codeception/CodeceptJS/issues/1863

### DIFF
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -1841,10 +1841,11 @@ class Puppeteer extends Helper {
 
       if (locator.isXPath()) {
         waiter = contextObject.waitForFunction((locator, text, $XPath) => {
+          eval($XPath); // eslint-disable-line no-eval
           const el = $XPath(null, locator);
           if (!el.length) return false;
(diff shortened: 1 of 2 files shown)


lib/helper/Puppeteer.js
```diff
@@ -1841,10 +1841,11 @@ class Puppeteer extends Helper {
 
       if (locator.isXPath()) {
         waiter = contextObject.waitForFunction((locator, text, $XPath) => {
+          eval($XPath); // eslint-disable-line no-eval
```
Vorobeyko: wherefore did you use eval?

Heavik: If we pass $XPath function as is to the Puppeteers waitForFunction it will be resolved as undefined. The possible reason is that $XPath function doesn't exit in Puppeteers browser context, so we need to add $XPath to the scope. The solution can be convert function to string and use eval in the browsers context. Exactly the same approach is used in waitForDetached or waitForEnabled functions. So I decided that eval can be acceptable approach here.

PR #2112, opened 2020-01-02
Refactor: update eslint and it's packages

elukoyanov:
## Motivation/Description of the PR
- Description of this PR, which problem it solves
- A link to the corresponding issue (if applicable).

Updated packages
- eslint
- eslint-config-airbnb-base
- eslint-plugin-import
- eslint-plugin-mocha

## Type of change

- [x] Dependencies changes/updates
- [ ] Breaking changes
- [ ] New functionality
- [ ] Bug fix
- [x] Documentation changes/updates
- [ ] Hot fix
- [ ] Markdown files fix - not related to source code

## Checklist:

- [ ] Tests have been added
- [x] Documentation has been added (Run `npm run docs`)
- [x] Lint checking (Run `npm run lint`)
- [x] Local tests are passed (Run `npm test`)

PR #2120, opened 2020-01-06
refactor: Transform Recorder into class with static methods

elukoyanov:

Applicable helpers:

- [ ] Webdriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe

- A link to the corresponding issue (if applicable).

Make `recorder` a static class instead of module object:
1. `recorder` and it's internal `RecorderSession` include `catch` method. It's not allowed to create non-class methods with reserved word names.
2. Problem described upper creates problem with typings. Typescript deny use of reserved words by default. And it leads to incompatible with most of users TS settings.
3. definition_test fails 
3. `recorder` is kind of entity. And entities should be created as class

## Type of change

- [ ] Breaking changes
- [ ] New functionality
- [x] Bug fix
- [ ] Hot fix

## Checklist:

- [ ] Tests have been added

## Notes
Linter fails on static fields definition: `static sessions = [];`.
Cause old eslint (compatible with ES5, not with ES6) does not support it.
#2112 should be merged first. Or this PR should be to that branch.
Therefore, it's Draft PR until eslint is updated.

Moreover, edit will be need until eslint update.


DavertMik:
Great! Personally, I like the current recorder but I'm not going to oppose progress. :+1:

DavertMik:
@elukoyanov any updates?

DavertMik:
@elukoyanov this PR is still a draft. We have a few more days before releasing CodeceptJS 3
If you want to get it merged please update it, so we could include it into release

elukoyanov:
Static class does not much differs from simple module, so i'll close PR.

Maybe it's better to use singleton object created from class. Maybe no

PR #2134, opened 2020-01-13
Release 2.4.1

PeterNgTr:

Hotfix - add missing lib that prevents codeceptjs from running.


package.json
```diff
@@ -89,6 +89,7 @@
     "@wdio/selenium-standalone-service": "^5.16.10",
     "@wdio/utils": "^5.16.15",
     "apollo-server-express": "^2.9.16",
+    "arrify": "^2.0.1",
```
elukoyanov: No more package errors mentioned?

PeterNgTr: I tried to `npm i arrify` then it worked.

PeterNgTr: my bad, the required lib should stay in `dependencies` instead of `devDependencies`

PR #2157, opened 2020-01-21
(improvement): run acceptance tests with run-workers cmd

PeterNgTr:
Using `run-workers` for acceptance tests


PeterNgTr:
not sure if we need this cause the execution time doesn't reduce much. Guys, what are your thoughts?

DavertMik:
I think it won't work to reduce execution time as circle ci boxes are limited by browsers and by memory
So I don't think we need this

PR #2162, opened 2020-01-24
Release 2.4.2

DavertMik:
Please review this update.

This PR includes fixes to several other pull requests:

* selenoid plugin - updated with better logging and documentation. Added allure support.
* `event.step.comment` removed comments for meta steps. This feature never worked fine and it broke allure reports when @PeterNgTr introduced allure reports for `I.say` commands
* changed and updated documentation in various places

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -531,9 +531,124 @@ Possible config options:
 
 -   `config`  
 
+## selenoid
+
+[Selenoid][8] plugin automatically starts browsers and video recording.
+Works with WebDriver helper.
+
+### Prerequisite
+
+This plugin **requires Docker** to be installed.
+
+> If you have issues starting Selenoid with this plugin consider using official [Configuration Manager][9] tool from Selenoid
+
+### Usage
+
+Selenoid plugin can be started in two ways:
+
+1.  **Automatic** - this plugin will create and manage selenoid container for you.
+2.  **Manual** - you create the conatainer and configure it with a plugin (recommended).
+
+#### Automatic
+
+If you are new to Selenoid and you want plug and play setup use automatic mode.
+
+Add plugin configuration in `codecept.conf.js`:
+
+```js
+plugins: {
+    selenoid: {
+      enabled: true,
+      deletePassed: true,
+      autoCreate: true,
+      autoStart: true,
+      sessionTimeout: '30m',
+      enableVideo: true,
+      enableLog: true,
+    },
+  }
+```
+
+When `autoCreate` is enabled it will pull the [latest Selenoid from DockerHub][10] and start Selenoid automatically.
+It will also create `browsers.json` file required by Selenoid.
+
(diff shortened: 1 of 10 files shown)


lib/plugin/selenoid.js
```diff
@@ -280,11 +236,128 @@ const selenoid = (config) => {
...
+  if (!stdout.includes('selenoid/firefox')) {
+    output.debug('Pulling selenoid/firefox...');
+    resultPromise = exec('docker pull selenoid/firefox:latest')
+      .then(() => output.debug('Pulled in selenoid/chrome'));
+    pulls.push(resultPromise);
+  }
+
+  return Promise.all(pulls).then(() => {
+    console.timeEnd('Pulled containers');
+    output.debug('Disable `autoCreate` in Selenoid plugin and pull in all containers manually to reduce this time');
```
koushikmohan1996: This will take time on first time alone since next time images will be already present and it won't pull the image again. :)

docs/plugins.md
```diff
@@ -531,9 +531,124 @@ Possible config options:
...
+## selenoid
+
+[Selenoid][8] plugin automatically starts browsers and video recording.
+Works with WebDriver helper.
+
+### Prerequisite
+
+This plugin **requires Docker** to be installed.
+
+> If you have issues starting Selenoid with this plugin consider using official [Configuration Manager][9] tool from Selenoid
```
pablopaul: @DavertMik "consider using **the** official"

docs/webdriver.md
```diff
@@ -112,6 +148,12 @@ desiredCapabilities: {
 }
 ```
 
+### Cloud Providers
+
+WebDriver protocol works over HTTP, so you need to have a Selenium Server to be running or other service that will launch a browser for you. That's why you may need to specify `host`, `port`, `protocol`, and `path` parameters.
```
pablopaul: @DavertMik "or **a** other service"

docs/puppeteer.md
```diff
@@ -247,51 +247,6 @@ npx codeceptjs def
 Mocking rules will be kept while a test is running. To stop mocking use `I.stopMocking()` command
 
 
-## Cloud Browsers
```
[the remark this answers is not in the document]
DavertMik: It was never extended to anything more than Aerokube Browsers. 

@codeceptjs/aerokube-plugin has never been downloaded more than 5 times a week, so I think this chapter was not really helpful. 

Probably we need to create a dedicated page for cloud browsers with better content and more alternatives

lib/plugin/selenoid.js
DavertMik: I agree but even pulling and checking images always takes about 2 secs for me. 
Not the big deal but I think it is useful to state that this amount of time can be reduced. Anyway, it is printed in debug mode only...

DavertMik: Anyway, I need to write this better...

PR #2198, opened 2020-02-12
added webdriver type function

### LINKED ISSUE #2185: There should be a method similar to pressKey wich could press multiple keys in sequence

Populating text to places which are not text fields or textareas.

#### What do you get instead?

As I can't use ```fillField()``` to achieve this, in these cases, I end up using multiple lines of ```pressKey()```, or sometimes transforming my string into an array and using a ```forEach()``` on it to solve the problem.

I think that a method like ```pressKeys``` would be ideal to solve problems like this.

```I.pressKeys("My long text here...")```

### PR DESCRIPTION
**Pull request by xMutaGenx**

Adding a type function to Webdriver, when `fillField` is not a valid option to use.

Applicable helpers:

- [x] Webdriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [x] Appium
- [ ] Protractor
- [ ] TestCafe

- Resolves #2185 

## Type of change

- [ ] Breaking changes
- [x] New functionality
- [ ] Bug fix
- [x] Documentation changes/updates
- [ ] Hot fix

## Checklist:

- [x] Tests have been added

### DIFF
--- a/lib/helper/WebDriver.js
+++ b/lib/helper/WebDriver.js
@@ -1605,6 +1605,18 @@ class WebDriver extends Helper {
   }
 
   /**
+   * {{> type }}
+   * Type out given array of keys or a string of text
+   */
+  async type(keys) {
+    if (Array.isArray(keys)) {
+      await this.browser.keys(keys);
+      return;
+    }
+    await this.browser.keys(keys.split(''));
+  }
+
+  /**
    * {{> resizeWindow }}
    * Appium: not tested in web, in apps doesn't work
    */
--- a/test/helper/WebDriver_test.js
+++ b/test/helper/WebDriver_test.js
@@ -305,7 +305,7 @@ describe('WebDriver', function () {
   describe('#type', () => {
     it('should type into a field', async () => {
       await wd.amOnPage('/form/field');
-      await wd.fillField('Name', '');
+      await wd.click('Name');
 
       await wd.type('Type Test');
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: lib/helper/WebDriver.js [hunk 1]
```diff
@@ -1604,6 +1604,18 @@ class WebDriver extends Helper {
...
+   * {{> type }}
+   * Type out given array of keys or a string of text
+   */
+  async type(keys) {
+    if (Array.isArray(keys)) {
+      await this.browser.keys(keys);
+      return;
+    }
+    await this.browser.keys(keys.split(''));
+  }
```
PeterNgTr: could you add tests? Also do we need to focus on any element to type?

xMutaGenx: Hey @PeterNgTr added the tests, and yes, you will need to 'focus' on the element, but that needs to be done by the tester to imitate the journey "I click on a field and I type.."

test/helper/WebDriver_test.js
```diff
@@ -302,6 +302,21 @@ describe('WebDriver', function () {
     });
   });
 
+  describe('#type', () => {
+    it('should type into a field', async () => {
+      await wd.amOnPage('/form/field');
+      await wd.fillField('Name', '');
```
Vorobeyko: Why you use fillField instead click?

DavertMik:
Cool, we will need the same function for other drivers but let's have this one to start!

PR #2204, opened 2020-02-15
fix(command): respect case generate test 

matheo:

`codeceptjs generate:test` is lower casing the input, restricting projects to define their own naming convention for the test files.

Also, for a newbie, it's helpful to have an introductory example showing that it's possible to use paths, and encourage its usage for better categorization of the tests :)

image

## Type of change

- [ ] Breaking changes
- [x] New functionality
- [ ] Bug fix
- [ ] Hot fix

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/command/generate.js
+++ b/lib/command/generate.js
@@ -39,7 +39,7 @@ module.exports.test = function (genPath) {
       message: 'Filename of a test',
       name: 'filename',
       default(answers) {
-        return (answers.feature).replace(' ', '_').toLowerCase() + defaultExt;
+        return (answers.feature).replace(' ', '_') + defaultExt;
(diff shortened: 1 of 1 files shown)


matheo:
@DavertMik here it is :) have a nice weekend!

DavertMik:
Thank you, @matheo

Could you help with another thing? If you use Typescript with CodeceptJS, could you write a small guide about it? Something like docs/typescript.md?
That would really help others to start with TS.

DavertMik:
After a bit of thinking, I will remove default value from test generation
I don't want all users created `Account/forgot_password` tests for everything...

PR #2219, opened 2020-02-21
(fix): add missing supported helpers

PeterNgTr:
We have added 2 more helpers and forgot to add them to supported helpers array of this plugin. 

By separating the supported helpers into a file, we could prevent the missing update on many places, in another words, we just update the new support helper into the list in file and we are all good. One place to update.

Applicable helpers:

- [ ] Webdriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] Appium
- [ ] Protractor
- [x] TestCafe
- [x] Playwright

## Type of change

- [ ] Breaking changes
- [ ] New functionality
- [x] Bug fix
- [ ] Hot fix

## Checklist:
- [ ] Tests have been added

### DIFF
--- a/lib/plugin/autoDelay.js
+++ b/lib/plugin/autoDelay.js
@@ -4,7 +4,7 @@ const recorder = require('../recorder');
 const event = require('../event');
 const log = require('../output').log;
 
-const supportedHelpers = require('./supportedHelpersList');
+const supportedHelpers = require('./standardActingHelpers');
 
 const methodsToDelay = [
   'click',
--- a/lib/plugin/autoDelay.js
+++ b/lib/plugin/autoDelay.js
@@ -11,6 +11,8 @@ const supportedHelpers = [
   'Appium',
   'Nightmare',
   'Puppeteer',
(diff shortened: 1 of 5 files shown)


lib/plugin/autoDelay.js
```diff
@@ -4,14 +4,7 @@ const recorder = require('../recorder');
...
 
-const supportedHelpers = [
-  'WebDriver',
-  'WebDriverIO',
-  'Protractor',
-  'Appium',
-  'Nightmare',
-  'Puppeteer',
-];
+const supportedHelpers = require('./supportedHelpersList');
```
DavertMik: can we specify somehow supported for what? 
in OOP I would use interfaces to describe what kind of behavior is supported but here I have no ideas.

We can name helpers like: `standardActingHelpers` so we assume they have standard actions like `click`, `fillField`, `saveScreenshot`, etc

PeterNgTr: @DavertMik I updated the name as request.

Issue #2150, opened 2020-01-18
waitForTimeout from config file doesnt work in wait* methods by default


Elverslane:
#### What are you trying to achieve?
i want to use any of wait* methods w/o hardcoding timeout because it already declared in config file.
#### What do you get instead?
instead i get TS error 
![Screenshot 2020-01-16 at 14 32 17](https://user-images.githubusercontent.com/17493900/72669744-ae9f0400-3a46-11ea-8ad0-23e855cf4a0e.png)

### Details

* CodeceptJS version: 2.3.6
* Operating System: macos
* puppeteer 

Here's the relevant part from our config:
```
   { 
     helpers: {
        Puppeteer: {
          url: "http://localhost",
          waitForTimeout: 2000,
          waitForAction: 500
    }
  }
}

```

DavertMik:
`sec` should be optional. I think we have error in our typings. 
@elukoyanov can we fix this?

PR #2231, opened 2020-02-25
Improve Scenario DX

### LINKED ISSUE #1972: Improve DX for enigneers using TypeScript (or type checks)

I'm trying to configure our code base in a way that would reliably warn us about type issues.
In particular I'd like to see correct type errors when I inject more than one page object into my Scenario.

This problem makes the CodeceptJS typings not usable in this specific case.

#### What do you get instead?

Instead, whenever I have more than one page object in the Scenario callback, I immediately get a lot of errors.

### Details

Here is a dummy example (pseud-code):

```js
Scenario('dum-dum', (I, pageA, pageC, pageB) => {
  pageA.doA();
  pageB.doB();
  pageC.doC();
});
```

Note that I have purposefully ordered arguments this way (pageC before pageB).

The following code will work fine in runtime but it will fail at "type check" time. 

**Note** that the following is true even if you don't use TypeScript but simply enable typechecking with the following `jsconfig.json`

```json
{
  "compilerOptions": {
    "checkJs": true
  }
}
```

There reason this code fails comes from the way CodeceptJS injects page objects and how that affects `ICodeceptCallback` type (steps.d.ts).

CodeceptJS uses [fn-args](https://github.com/sindresorhus/fn-args) to inject page objects. It reads function's arguments and it expects exact match between argument name and page object name declared in the config file.

The problem with this approach is that the generated `ICodeceptCallback` expects arguments to be passed in a particular order. In context of this example it could look similar to:

```ts
type ICodeceptCallback = (i?: CodeceptJS.I, current?:any, pageA: PageA, pageB: PageB, pageC: PageC, ...args: any) => void;
```

See that it expects the 3rd argument to be of type `PageA`. However, in runtime (and plain, unchecked javascript) it's totally fine to actually put pageC there instead.

* CodeceptJS version: 2.3.2

## Proposed solution

All those problems would disappear if `ICodeceptCallback` expected just two arguments:

- `I`
- and `custom`

Where `custom` would be an object containing all user-defined page objects, fragments, etc.

This way the order of parameters would no longer matter and type checking would be happy :)

If done this way this example scenario would look like so (I'm using destructuring):

```js
Scenario('dum-dum', (I, { pageA, pageC, pageB }) => {
});
```

and generated `ICodeceptCallback` would become

```ts
type ICodeceptCallback = (i?: CodeceptJS.I, custom?: { pageA?: PageA, pageB?: PageB, pageC?: PageC }) => void;
```

## Workaround

Possible workaround for people with the same problem is to use `inject` function at the top of a testing suite.

Instead of doing

```js
Scenario('x', (I, page) => {})
```

do
```
const { page } = inject();
Scenario('x', (I) => {})
```

Although I don't know whether that will be the same with regards to page object's state.

Vorobeyko:
- Resolves #1972, #2150, #2330 
Applicable helpers:

- [ ] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] puppeteerCoverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] wdio

## Type of change

- [x] :fire: Breaking changes
- [ ] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [x] Tests have been added


lib/parser.js
```diff
@@ -1,15 +1,20 @@
 function _interopDefault(ex) { return (ex && (typeof ex === 'object') && 'default' in ex) ? ex.default : ex; }
-const parser = _interopDefault(require('parse-function'))({ ecmaVersion: 2017 });
+const acorn = require('acorn');
+const parser = _interopDefault(require('parse-function'))({ parse: acorn.parse, ecmaVersion: 11, plugins: ['objectRestSpread'] });
```
PeterNgTr: isn't it the `parse-function` lib that gave us troubles twice in the past?

test/data/sandbox/configs/pageObjects/first_test.js
```diff
@@ -1,12 +1,12 @@
 Feature('PageObject');
 
-Scenario('@ClassPageObject', async (I, classpage) => {
+Scenario('@ClassPageObject', async ({ I, classpage }) => {
```
PeterNgTr: if there is way to to input with or without curly brackets would be nice, users don't need to migrate their current scripts ^^

lib/parser.js
DavertMik: yeah, one of the reason of this PR is to remove it, as I understand

test/data/sandbox/configs/pageObjects/first_test.js
DavertMik: We will provide a migration script for them.

package.json
```diff
@@ -84,10 +84,12 @@
     "@pollyjs/adapter-puppeteer": "^2.6.3",
     "@pollyjs/core": "^2.6.3",
     "@types/inquirer": "^0.0.35",
+    "@types/jscodeshift": "^0.6.3",
     "@types/node": "^8.10.54",
     "@wdio/sauce-service": "^5.16.10",
     "@wdio/selenium-standalone-service": "^5.16.10",
     "@wdio/utils": "^5.16.15",
+    "acorn": "^7.1.0",
```
Vorobeyko: need to remove

lib/scenario.js
```diff
@@ -185,21 +186,22 @@ module.exports.suiteTeardown = function (suite) {
...
-      throw new Error(`Object of type ${obj} is not defined in container`);
+    if (!objects[key]) {
+      throw new Error(`Object of type ${key} is not defined in container`);
     }
-    testArguments.push(container.support(obj));
+    testArgs[key] = container.support(key);
   }
-  return testArguments;
+  return testArgs;
 };
```
PeterNgTr: @Vorobeyko  This change will kill the test issue.

```js
const getInjectedArguments = (fn, test) => {
  const container = require('./container');
  const testArgs = {};
  const params = parser.getParams(fn) || [];
  const objects = container.support();
  for (const key of params) {
    testArgs[key] = {};
    if (test && test.inject && test.inject[key]) {
      // @FIX: need fix got inject
      testArgs[key] = test.inject[key];
      continue;
    }
    if (!objects[key]) {
      throw new Error(`Object of type ${key} is not defined in container`);
    }
    testArgs[key] = container.support(key);
  }

  return testArgs;
};
```

docs/webapi/waitForVisible.mustache
```diff
@@ -6,4 +6,4 @@ I.waitForVisible('#popup');
 ```
 
 @param {CodeceptJS.LocatorOrString} locator element located by CSS|XPath|strict locator.
-@param {number} [sec=1] (optional, `1` by default) time in seconds to wait
```
elukoyanov, line 9: why it's removed?

test/runner/definitions_test.js
```diff
@@ -54,6 +54,15 @@ describe('Definitions', function () {
     });
   });
 
+  xit('callback order typings', (done) => {
```
elukoyanov, line 57: why it's skipped

PR #2250, opened 2020-03-04
commentSteps plugin

DavertMik:

This PR adds a better way to document the test flow by adding another function for writing a comment:

```js
Scenario('global var', (I) => {
  __`Prepare user base`;
  I.print('other thins');

  __`Update data`;
  I.print('do some things');

  __`Check the result`;
  I.print('see everything works');
});
```
A comment written inside `__` template literal will be printed as top-level nested step in output.

Screenshot 2020-03-04 at 01 45 47

This way it is easy to split test by blocks or use BDD

P.S. `__` is my personal preference. Any suggestions of naming that function are welcome.

## Type of change

- [ ] :fire: Breaking changes
- [ ] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [x] Tests have been added

### DIFF
--- a/lib/codecept.js
+++ b/lib/codecept.js
@@ -76,6 +76,7 @@ class Codecept {
       global.DataTable = require('./data/table');
       global.locate = locator => require('./locator').build(locator);
       global.inject = container.support;
+      global.share = container.share;
       global.secret = require('./secret').secret;
       global.codeceptjs = require('./index'); // load all objects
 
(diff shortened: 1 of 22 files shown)


PeterNgTr:
@DavertMik what do you think about this? The second var will be parsed instead. If there is no second var, we just basically print the string in first var. I would see it's cleaner way. 😄 
```js
Scenario('global var', (I) => {
  I.print('other thins',  'Prepare user base');
  I.print('do some things',   'Update data');
  I.print('see everything works', 'Check the result');
});
```

DavertMik:
Merged to 2.6.0

PR #2253, opened 2020-03-06
Screenshot session fix

### LINKED ISSUE #2251: I.saveScreenshot is confused by multiple sessions
I am writing tests that involve multiple sessions using the `session()` method made available by CodeceptJS.  As I write these tests I often take screenshots to ensure the tests are performing as I expect.

#### What do you get instead?
It seems that after running steps in a different session than the main session subsequent calls to I.saveScreenshot show the last session used, not the main session as I would expect.

```bash
$ npx codeceptjs run --steps --verbose
CodeceptJS v2.5.0
Using test root "/Users/[REDACTED]/Projects/session-test"
Helpers: Puppeteer
Plugins: screenshotOnFail, retryFailedStep

session --
    [1] Starting recording promises
    Emitted | suite.before ([object Object])
  test something
    Emitted | test.before ([object Object])
    Emitted | test.start ([object Object])
    Emitted | step.before (I am on page "/")
    Emitted | step.after (I am on page "/")
    Emitted | step.before (I fill field "input[name="q"]", "This is User 1")
    Emitted | step.after (I fill field "input[name="q"]", "This is User 1")
    Emitted | step.before (I save screenshot "user1.png")
    Emitted | step.after (I save screenshot "user1.png")
    Emitted | step.before (I save screenshot "user1_2.png")
    Emitted | step.after (I save screenshot "user1_2.png")
    Emitted | step.start (I am on page "/")
    I am on page "/"
    › [Url] https://www.google.com/?gws_rd=ssl
    Emitted | step.passed (I am on page "/")
    Emitted | step.finish (I am on page "/")
    Emitted | step.start (I fill field "input[name="q"]", "This is User 1")
    I fill field "input[name="q"]", "This is User 1"
    Emitted | step.passed (I fill field "input[name="q"]", "This is User 1")
    Emitted | step.finish (I fill field "input[name="q"]", "This is User 1")
    Emitted | step.start (I save screenshot "user1.png")
    I save screenshot "user1.png"
    › Screenshot is saving to /Users/[REDACTED]/Projects/session-test/output/user1.png
    Emitted | step.passed (I save screenshot "user1.png")
    Emitted | step.finish (I save screenshot "user1.png")
 › [Incognito Tab] opened
    [1] Starting <session:user2> session
    Emitted | step.before (I am on page "/")
    Emitted | step.after (I am on page "/")
    Emitted | step.before (I fill field "input[name="q"]", "This is User 2")
    Emitted | step.after (I fill field "input[name="q"]", "This is User 2")
    Emitted | step.before (I save screenshot "user2.png")
    Emitted | step.after (I save screenshot "user2.png")
    Emitted | step.start (user2: I am on page "/")
    user2: I am on page "/"
    › [Url] https://www.google.com/?gws_rd=ssl
    Emitted | step.passed (user2: I am on page "/")
    Emitted | step.finish (user2: I am on page "/")
    Emitted | step.start (user2: I fill field "input[name="q"]", "This is User 2")
    user2: I fill field "input[name="q"]", "This is User 2"
    Emitted | step.passed (user2: I fill field "input[name="q"]", "This is User 2")
    Emitted | step.finish (user2: I fill field "input[name="q"]", "This is User 2")
    Emitted | step.start (user2: I save screenshot "user2.png")
    user2: I save screenshot "user2.png"
    › Screenshot is saving to /Users/[REDACTED]/Projects/session-test/output/user2.png
    Emitted | step.passed (user2: I save screenshot "user2.png")
    Emitted | step.finish (user2: I save screenshot "user2.png")
    [1] <session:user2> Finalize <session:user2> session
    Emitted | step.start (I save screenshot "user1_2.png")
    I save screenshot "user1_2.png"
    › Screenshot is saving to /Users/[REDACTED]/Projects/session-test/output/user1_2.png
    Emitted | step.passed (I save screenshot "user1_2.png")
    Emitted | step.finish (I save screenshot "user1_2.png")
    Emitted | test.passed ([object Object])
    Emitted | test.finish ([object Object])
  ✔ OK in 2915ms

    Emitted | test.after ([object Object])
    Emitted | suite.after ([object Object])

  OK  | 1 passed   // 3s
    Emitted | global.result ([object Object])
    Emitted | global.after ([object Object])
```

```js
Feature('session');

Scenario('test something', (I) => {
  const searchSel = 'input[name="q"]';

  I.amOnPage('/');
  I.fillField(searchSel, 'This is User 1');
  I.saveScreenshot('user1.png');

  session('user2', () => {
    I.amOnPage('/');
    I.fillField(searchSel, 'This is User 2');
    I.saveScreenshot('user2.png');
  });

  I.saveScreenshot('user1_2.png');
});

```

### Details

* CodeceptJS version: 2.5.0
* NodeJS Version: 12.5.0
* Operating System: macOS Catalina (10.15.3)
* Helper: puppeteer
* Configuration file:

```js
const { setHeadlessWhen } = require('@codeceptjs/configure');

// HEADLESS=true npx codecept run

exports.config = {
  tests: './*_test.js',
  output: './output',
  helpers: {
    Puppeteer: {
      url: 'http://www.google.com',
      show: false
    }
  },
  include: {
    I: './steps_file.js'
  },
  bootstrap: null,
  mocha: {},
  name: 'session-test',
  plugins: {
    retryFailedStep: {
      enabled: true
    },
    screenshotOnFail: {
      enabled: true
    }
  }
}
```

ChexWarrior:
- This PR lets Puppeteer successfully determine the correct page object to use the `screenshot` method on when there are multiple sessions.  Previously Puppeteer would just use the last session opened for a screenshot.
- Resolve #2251 

I'd like some constructive criticism on the fix and any advice on any edge cases I may have missed. I'm not ready to say this is finished yet, I'd like to see if I can add a good test for this and I still want to test its effect on the **screenshotOnFail** plugin.  Also I believe I can add a similar fix for Playwright which is fairly similar to Puppeteer.

Applicable helpers:

- [ ] WebDriver
- [x] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [x] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] puppeteerCoverage
- [ ] retryFailedStep
- [x] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] wdio

## Type of change

- [ ] :fire: Breaking changes
- [x] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/helper/Playwright.js
+++ b/lib/helper/Playwright.js
@@ -338,12 +338,19 @@ class Playwright extends Helper {
         this.sessionPages[this.activeSessionName] = existingPages[0];
         return this._setPage(this.sessionPages[this.activeSessionName]);
       },
-      restoreVars: async () => {
+      restoreVars: async (session) => {
         this.withinLocator = null;
-        this.activeSessionName = '';
         this.browserContext = defaultContext;
-        const existingPages = await this.browserContext.pages();
-        await this._setPage(existingPages[0]);
+
+        if (!session) {
+          this.activeSessionName = '';
+          const existingPages = await this.browserContext.pages();
+          await this._setPage(existingPages[0]);
+        } else {
+          this.activeSessionName = session;
+          await this._setPage(this.sessionPages[this.activeSessionName]);
+        }
+
         return this._waitForAction();
       },
     };
--- a/lib/helper/Playwright.js
+++ b/lib/helper/Playwright.js
@@ -1550,6 +1550,18 @@ class Playwright extends Helper {
 
     this.debug(`Screenshot is saving to ${outputFile}`);
 
+    if (this.activeSessionName) {
+      const activeSessionPage = this.sessionPages[this.activeSessionName];
+
+      if (activeSessionPage) {
+        return activeSessionPage.screenshot({
+          path: outputFile,
+          fullPage: fullPageOption,
+          type: 'png',
+        });
+      }
+    }
+
     return this.page.screenshot({ path: outputFile, fullPage: fullPageOption, type: 'png' });
   }
 
--- a/lib/helper/Playwright.js
+++ b/lib/helper/Playwright.js
@@ -1544,10 +1544,7 @@ class Playwright extends Helper {
     const outputFile = screenshotOutputFolder(fileName);
 
     this.debug(`Screenshot is saving to ${outputFile}`);
-    const openPages = await this.browserContext.pages();
-    if (openPages.length > 1) {
-      this.page = openPages[openPages.length - 1];
-    }
+
     return this.page.screenshot({ path: outputFile, fullPage: fullPageOption, type: 'png' });
   }
 
--- a/lib/helper/Playwright.js
+++ b/lib/helper/Playwright.js
@@ -285,8 +285,10 @@ class Playwright extends Helper {
     contexts.shift();
 
     // ensure current page is in default context
-    const existingPages = await this.browserContext.pages();
-    await this._setPage(existingPages[0]);
+    if (this.page) {
+      const existingPages = await this.browserContext.pages();
+      await this._setPage(existingPages[0]);
+    }
 
     await Promise.all(contexts.map(c => c.close()));
 
--- a/lib/helper/Playwright.js
+++ b/lib/helper/Playwright.js
@@ -185,6 +185,8 @@ class Playwright extends Helper {
     this.isRemoteBrowser = false;
     this.isRunning = false;
     this.isAuthenticated = false;
+    this.sessionPages = {};
+    this.activeSessionName = '';
 
     // override defaults with config
     this._setConfig(config);
@@ -316,8 +318,9 @@ class Playwright extends Helper {
   _session() {
     const defaultContext = this.browserContext;
     return {
-      start: async () => {
+      start: async (sessionName = '') => {
         this.debugSection('Incognito Tab', 'opened');
+        this.activeSessionName = sessionName;
 
         const bc = await this.browser.newContext();
         const page = await bc.newPage();
@@ -332,10 +335,12 @@ class Playwright extends Helper {
       loadVars: async (context) => {
         this.browserContext = context;
         const existingPages = await context.pages();
-        return this._setPage(existingPages[0]);
+        this.sessionPages[this.activeSessionName] = existingPages[0];
+        return this._setPage(this.sessionPages[this.activeSessionName]);
       },
       restoreVars: async () => {
         this.withinLocator = null;
+        this.activeSessionName = '';
         this.browserContext = defaultContext;
         const existingPages = await this.browserContext.pages();
         await this._setPage(existingPages[0]);
--- a/lib/helper/Playwright.js
+++ b/lib/helper/Playwright.js
@@ -284,17 +284,17 @@ class Playwright extends Helper {
     const contexts = await this.browser.contexts();
     contexts.shift();
 
-    // ensure current page is in default context
-    if (this.page) {
-      const existingPages = await this.browserContext.pages();
-      await this._setPage(existingPages[0]);
-    }
-
     await Promise.all(contexts.map(c => c.close()));
 
     if (this.options.restart) {
       this.isRunning = false;
       return this._stopBrowser();
+    }
+
+    // ensure current page is in default context
+    if (this.page) {
+      const existingPages = await this.browserContext.pages();
+      await this._setPage(existingPages[0]);
     }
 
     if (this.options.keepBrowserState) return;
--- a/lib/helper/Playwright.js
+++ b/lib/helper/Playwright.js
@@ -284,6 +284,10 @@ class Playwright extends Helper {
     const contexts = await this.browser.contexts();
     contexts.shift();
 
+    // ensure current page is in default context
+    const existingPages = await this.browserContext.pages();
+    await this._setPage(existingPages[0]);
+
     await Promise.all(contexts.map(c => c.close()));
 
     if (this.options.restart) {
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -295,9 +295,10 @@ class Puppeteer extends Helper {
     return {
       start: async (sessionName = '') => {
         this.debugSection('Incognito Tab', 'opened');
+        this.activeSessionName = sessionName;
 
         const bc = await this.browser.createIncognitoBrowserContext();
-        this.sessionPages[sessionName] = await bc.newPage();
+        await bc.newPage();
 
         // Create a new page inside context.
         return bc;
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -261,17 +261,17 @@ class Puppeteer extends Helper {
     const contexts = this.browser.browserContexts();
     const defaultCtx = contexts.shift();
 
-    // ensure this.page is from default context
-    if (this.page) {
-      const existingPages = defaultCtx.targets().filter(t => t.type() === 'page');
-      await this._setPage(await existingPages[0].page());
-    }
-
     await Promise.all(contexts.map(c => c.close()));
 
     if (this.options.restart) {
       this.isRunning = false;
       return this._stopBrowser();
+    }
+
+    // ensure this.page is from default context
+    if (this.page) {
+      const existingPages = defaultCtx.targets().filter(t => t.type() === 'page');
+      await this._setPage(await existingPages[0].page());
     }
 
     if (this.options.keepBrowserState) return;
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -291,7 +291,7 @@ class Puppeteer extends Helper {
 
   _session() {
     return {
-      start: async () => {
+      start: async (sessionName = '') => {
         this.debugSection('Incognito Tab', 'opened');
 
         const bc = await this.browser.createIncognitoBrowserContext();
@@ -1614,6 +1614,7 @@ class Puppeteer extends Helper {
     const outputFile = screenshotOutputFolder(fileName);
 
     this.debug(`Screenshot is saving to ${outputFile}`);
+
     const openSessions = await this.browser.pages();
     if (openSessions.length > 1) {
       this.page = await this.browser.targets()[this.browser.targets().length - 1].page();
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -313,6 +313,7 @@ class Puppeteer extends Helper {
       },
       restoreVars: async () => {
         this.withinLocator = null;
+        this.activeSessionName = '';
         const existingPages = await this.browser.targets().filter(t => t.type() === 'page');
         await this._setPage(await existingPages[0].page());
         return this._waitForAction();
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -291,7 +291,7 @@ class Puppeteer extends Helper {
 
   _session() {
     return {
-      start: async () => {
+      start: async (config = {}) => {
         this.debugSection('Incognito Tab', 'opened');
 
         const bc = await this.browser.createIncognitoBrowserContext();
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -173,8 +173,6 @@ class Puppeteer extends Helper {
     this.isRemoteBrowser = false;
     this.isRunning = false;
     this.isAuthenticated = false;
-    this.sessionPages = {};
-    this.activeSessionName = '';
 
     // override defaults with config
     this._setConfig(config);
@@ -293,9 +291,8 @@ class Puppeteer extends Helper {
 
   _session() {
     return {
-      start: async (sessionName = '') => {
+      start: async () => {
         this.debugSection('Incognito Tab', 'opened');
-        this.activeSessionName = sessionName;
 
         const bc = await this.browser.createIncognitoBrowserContext();
         await bc.newPage();
@@ -308,8 +305,8 @@ class Puppeteer extends Helper {
       },
       loadVars: async (context) => {
         const existingPages = context.targets().filter(t => t.type() === 'page');
-        this.sessionPages[this.activeSessionName] = await existingPages[0].page();
-        return this._setPage(this.sessionPages[this.activeSessionName]);
+
+        return this._setPage(await existingPages[0].page());
       },
       restoreVars: async () => {
         this.withinLocator = null;
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -173,6 +173,8 @@ class Puppeteer extends Helper {
     this.isRemoteBrowser = false;
     this.isRunning = false;
     this.isAuthenticated = false;
+    this.sessionPages = {};
+    this.activeSessionName = '';
 
     // override defaults with config
     this._setConfig(config);
@@ -291,8 +293,9 @@ class Puppeteer extends Helper {
 
   _session() {
     return {
-      start: async () => {
+      start: async (sessionName = '') => {
         this.debugSection('Incognito Tab', 'opened');
+        this.activeSessionName = sessionName;
 
         const bc = await this.browser.createIncognitoBrowserContext();
         await bc.newPage();
@@ -305,8 +308,8 @@ class Puppeteer extends Helper {
       },
       loadVars: async (context) => {
         const existingPages = context.targets().filter(t => t.type() === 'page');
-
-        return this._setPage(await existingPages[0].page());
+        this.sessionPages[this.activeSessionName] = await existingPages[0].page();
+        return this._setPage(this.sessionPages[this.activeSessionName]);
       },
       restoreVars: async () => {
         this.withinLocator = null;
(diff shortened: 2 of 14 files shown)


Vorobeyko:
@ChexWarrior Can you write tests please?

ChexWarrior:
@Vorobeyko Sure thing, I'm reviewing the existing tests to see how it all works and get some ideas.  

One idea I have is to test that screenshots in different sessions (on different pages) are actually different by hashing the contents of each image ([SO link](https://stackoverflow.com/a/2488269)).  I'm assuming the best place for this type of test would be in the Puppeteer helper test.

PeterNgTr:
maybe you can consider a similar approach for Playwright helper. :)

DavertMik:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones removed
==============
+ test/helper/Puppeteer_test.js  -5
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/DavertMik/CodeceptJS/pullRequest?prid=5135646&bid=16874332)

ChexWarrior:
I pushed up some commits involving setup for getting the `session()` method to run properly in the `Puppeteer_test.js` file.  This mostly involved ensuring the Puppeteer helper was created using the `container.create` method and starting the `recorder` before the tests actually ran. If there was a cleaner (better) way to do this please let me know.

ChexWarrior:
So as I work through building these tests I've realized my original approach isn't working, in fact I think a much simpler approach will work.  Since each new page's context created by a session is stored in the `savedSessions[sessionName].vars` we really only have to ensure we are setting the pages properly with `_setPage`. I'll report back with more.

DavertMik:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones removed
==============
+ test/helper/Puppeteer_test.js  -4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/DavertMik/CodeceptJS/pullRequest?prid=5135646&bid=16874332)

ChexWarrior:
Okay, running into an issue that I'm assuming is caused by the environment in which the helper tests are running.  

The bug I started this PR for was that `saveScreenshot` would always use the [last page created for a screenshot](https://github.com/Codeception/CodeceptJS/blob/b5ec7b103f5027afa02a5ee8de0346998748ed1e/lib/helper/Puppeteer.js#L1617-L1620) instead of the page created in the session the `saveScreenshot` method was called in.  When I started working on this issue I thought the solution would be to track the different sessions being created by `session` within the Puppeteer helper instance itself.  When a session would load I would retrieve the page created for that session and use it in the `saveScreenshot` method.  However, after working on the tests I realized there was a much simpler way to achieve this goal. Every session created in Puppeteer creates its own incognito browser context and uses that to create a new page for the session.  The created browser context is stored in a `vars` property unique to each created session.  If a particular session is reused later the code simply passes this stored browser context into the `session.loadVars` method, finds the page created in this context and then runs the `Puppeteer._setPage` method to update the "active" page object.  When a session ends the code runs `session.restoreVars` method which is supposed to grab the original default page and runs `Puppeteer_setPage` method on that.  I modified this method to use the `browser.defaultBrowserContext()` method to ensure that we only search for pages on the originally created browserContext, it was originally using browser.targets() which grabbed all pages (targets) created across all browser contexts.  Finally, I fixed the original bug by removing any logic about other session pages, since `_setPage` was being called properly within the session methods `saveScreenshot` would always take an image of the proper page.  

**Current Issue:**  I have a test in `Puppeteer_test.js` that just takes screenshots in different sessions right now (no actual testing part yet) you can see it [here](https://github.com/ChexWarrior/CodeceptJS/blob/9b114eae2f41c563181a900710525cdb8cae46a3/test/helper/Puppeteer_test.js#L128-L144). As I mentioned [before](https://github.com/Codeception/CodeceptJS/pull/2253#issuecomment-597114456), to get the `session()` method to work I had to include several other classes (`session`, `container`, `recorder`) and modify how the Puppeteer helper was [created](https://github.com/ChexWarrior/CodeceptJS/blob/9b114eae2f41c563181a900710525cdb8cae46a3/test/helper/Puppeteer_test.js#L74-L96).  When I run the actual test it looks like the screenshots are not being taken properly, however the actual issue is that the session doesn't set the proper page until after the I.amOnPage method executes.  This is a bit hard to explain, but if you run a debugger during that test you'll see the first `I.amOnPage` method in the initial 'user2' session runs before the `session.loadVars` method actually runs the `_setPage` method to the correct page.  I'm not sure why, but I assume its with how the session builds the method passed to the recorder object.  I'm still wrapping my head around how that part works.  

What makes me think my solution actually works however is that if I run a script using my modified CodeceptJS code that performs the same actions as the test I described above it works like it supposed to.  To reproduce you can clone my branch and run the following script:

```js
Feature('session');

Scenario('test something', async (I) => {
  I.amOnPage('/');
  I.saveScreenshot('session_user1_1.png', true);

  session('user2', () => {
    I.amOnPage('/info');
    I.saveScreenshot('session_user2_1.png', true);
  });

  I.saveScreenshot('session_user1_2.png', true);

  session('user2', async () => {
    I.saveScreenshot('session_user2_2.png', true);
  });
});
```
codecept.conf.js
```js
const { setHeadlessWhen } = require('@codeceptjs/configure');

// HEADLESS=true npx codecept run

exports.config = {
  tests: './*_test.js',
  output: './output',
  helpers: {
    Puppeteer: {
      url: 'http://localhost:8000',
      show: false
    }
  },
  bootstrap: null,
  mocha: {},
  name: 'session-test',
}
```

I know that's a big dump of info, but I wanted to get it up here before stopping working on it today.

ChexWarrior:
*sigh* I realize now that I should create a runner test for this that actually executes a codecept js test, that would take care of any setup....

ChexWarrior:
Well, I realize now that maybe I'm creating the wrong kind of test here, testing `session` requires several other objects that aren't used else where by the Puppeteer helper tests and attempting to shoehorn them in doesn't seem like its going to work easily.  My next thought was creating a runner test, since that actually executes `codecept run` and the session object would be available. Unfortunately, those tests don't seem to be meant for running specific helper tests.  I feel as though I'm in the middle of two kinds of tests: I need to ensure that `session()` works properly but specifically with the Puppeteer helper.  Any ideas on where this test should be placed? 

@Vorobeyko @PeterNgTr

PeterNgTr:
@ChexWarrior well, I just thought it is more related to `sessions` tests. As we just want to capture the screenshot of certain session. And basically, puppeteer is just one of the helpers, so it's not really the only puppeteer helper, like say, you use playwright helper and you want achieve the same thing, in the end of the day, it's session related, right? My 2 cent thought.

ChexWarrior:
@PeterNgTr I agree, this is a test about sessions, however the actual fix is Puppeteer (and Playwright eventually) specific... That is to fix the bug I actually removed code within the `saveScreenshot` method on the Puppeteer helper and to test this I need to create sessions using the Puppeteer helper.  So I think I will proceed with create a runner test that uses the Puppeteer helper unless someone has another idea.

ChexWarrior:
Adding these tests are `runner` does seem wrong though, none of the other tests use "real" helpers and none depend on the sample application.  But again, it doesn't look like the Helper tests are actually setup to run the `session` method, does it make more sense to attempt to add this in? I have some progress on this currently, but like I mentioned earlier its not quite right and causes some issues with the `recorder` object.

ChexWarrior:
Okay I figured it out, looks like the acceptance tests are exactly where I needed to put this sort of test, sorry for the confusion -that's what I get for not reading through all the code first... 

@PeterNgTr I've add a test for my fix in the `test/acceptance/session_test.js` file  called **screenshots reflect the current page of current session**.  Does this seem acceptable?

ChexWarrior:
Looks like my test is failing the Playwright helper, which makes sense, I'll start working on that helper's fix now...

DavertMik:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones removed
==============
+ lib/helper/Puppeteer.js  -1
+ lib/helper/Playwright.js  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/DavertMik/CodeceptJS/pullRequest?prid=5135646&bid=16874332)

DavertMik:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones removed
==============
+ lib/helper/Playwright.js  -1
+ lib/helper/Puppeteer.js  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/DavertMik/CodeceptJS/pullRequest?prid=5135646&bid=16874332)

DavertMik:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones removed
==============
+ lib/helper/Puppeteer.js  -1
+ lib/helper/Playwright.js  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/DavertMik/CodeceptJS/pullRequest?prid=5135646&bid=16874332)

DavertMik:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones removed
==============
+ lib/helper/Puppeteer.js  -1
+ lib/helper/Playwright.js  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/DavertMik/CodeceptJS/pullRequest?prid=5135646&bid=16874332)

DavertMik:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones added
============
- test/acceptance/codecept.WebDriver.js  1
- test/acceptance/codecept.WebDriverIO.js  1
         

Clones removed
==============
+ lib/helper/Playwright.js  -1
+ lib/helper/Puppeteer.js  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/DavertMik/CodeceptJS/pullRequest?prid=5135646&bid=16874332)

test/acceptance/session_test.js
```diff
@@ -13,6 +13,35 @@ Scenario('simple session @WebDriverIO @Protractor @Puppeteer @Playwright', (I) =
   I.seeInCurrentUrl('/info');
 });
 
+Scenario('screenshots reflect the current page of current session @WebDriverIO @Protractor @Puppeteer @Playwright', async (I) => {
```
PeterNgTr: is your fix covered all those helpers, if not just provide the one you covered like `@Puppeteer @Playwright`?

DavertMik:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones removed
==============
+ lib/helper/Playwright.js  -1
+ lib/helper/Puppeteer.js  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/DavertMik/CodeceptJS/pullRequest?prid=5135646&bid=16874332)

DavertMik:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones removed
==============
+ lib/helper/Playwright.js  -1
+ lib/helper/Puppeteer.js  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/DavertMik/CodeceptJS/pullRequest?prid=5135646&bid=16874332)

test/acceptance/session_test.js
ChexWarrior: Done!

DavertMik:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones removed
==============
+ lib/helper/Playwright.js  -1
+ lib/helper/Puppeteer.js  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/DavertMik/CodeceptJS/pullRequest?prid=5135646&bid=16874332)

ChexWarrior:
@PeterNgTr Looks like an unrelated test failed, something to do with ApiDataFactory.

PeterNgTr:
> @PeterNgTr Looks like an unrelated test failed, something to do with ApiDataFactory.

Yeah @ChexWarrior, that test is intermittently failed. I think it is tracked.

ChexWarrior:
My fix seems to revert the issue in #1266, looking into it...

ChexWarrior:
Is there a built in way to ignore a step failure? To test the **screenshotOnFail** plugin I'd like to fail a step and check if the screenshot taken is equal to a screenshot I took of the page just before the failure.

ChexWarrior:
Just to summarize, the real issue here is getting Puppeteer/Playwright to take a screenshot of the correct page when something fails ala the **screenshotOnFail** plugin. 

My original issue of screenshots being confused by the session was due to the fix in #1266. Removing that fix gets rid of the issue, however to fix the problem with **screenshotOnFail** we need to make the helpers responsible for getting the correct page.  

I do this by storing each page created in a session to a [new `sessionPages` property](https://github.com/ChexWarrior/CodeceptJS/blob/084214a6d8ed55d8a1ff94f31c446dcb53c76472/lib/helper/Puppeteer.js#L176-L177) in each helper.  When a session starts [I pass the `sessionName` var](https://github.com/ChexWarrior/CodeceptJS/blob/084214a6d8ed55d8a1ff94f31c446dcb53c76472/lib/session.js#L68) into the start method and [set the new `activeSessionName` property](https://github.com/ChexWarrior/CodeceptJS/blob/084214a6d8ed55d8a1ff94f31c446dcb53c76472/lib/helper/Puppeteer.js#L298) on the helper.  During the `session.loadVar` method[ I store the newly created page in `sessionPages[activeSessionName]`](https://github.com/ChexWarrior/CodeceptJS/blob/084214a6d8ed55d8a1ff94f31c446dcb53c76472/lib/helper/Puppeteer.js#L311).  When `saveScreenshot` is called it [checks if `activeSessionName` is truthy and if so uses the stored page for the screenshot](https://github.com/ChexWarrior/CodeceptJS/blob/084214a6d8ed55d8a1ff94f31c446dcb53c76472/lib/helper/Puppeteer.js#L1627-L1637).  Finally when the `session.restoreVars` method is called [the `activeSessionName` is reset](https://github.com/ChexWarrior/CodeceptJS/blob/084214a6d8ed55d8a1ff94f31c446dcb53c76472/lib/helper/Puppeteer.js#L316).  

The problem is when an error occurs `session.restoreVars` is called before the **screenshotOnFail** plugin takes the screenshot which causes `activeSesisonName` to be reset resulting in a screenshot of the main (non-session) page.  My idea for a solution is to pass a flag to restoreVars that gets flipped when an error occurs which will indicate **not** to wipe the `activeSessionName`.

DavertMik:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones added
============
- lib/helper/Puppeteer.js  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/DavertMik/CodeceptJS/pullRequest?prid=5135646&bid=16874332)

DavertMik:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones added
============
- lib/helper/Puppeteer.js  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/DavertMik/CodeceptJS/pullRequest?prid=5135646&bid=16874332)

ChexWarrior:
So I've gotten the **screenshotOnFail** plugin to take a screenshot of the correct session in Puppeteer and Playwright.  When there is an error during a session step I pass the current `sessionName` to the `session.restoreVars` method.  Within `restoreVars` I set the current page to whatever page is stored for that session, if no session is passed then I simply revert to the default (non-session) page.  I also had to make some changes in the helper's `_after` methods to ensure the default page is set when this method is called.

Currently I'm trying to do two things:
1. Track down exactly why the Playwright Tests are failing, they pass with no issues locally, but that could be due to environmental differences between my machine and GitHub Actions.  

2.  Write a good test for the *screenshotOnFail* plugin.

DavertMik:
Thanks, @ChexWarrior 
I will try to use the similar solution for WebDriver so we can come up in next release with unified solution. I also will try to fix those Playwright failures which probably not related to your commit but to overall Playwright stability

lib/helper/Puppeteer.js
```diff
@@ -291,11 +300,13 @@ class Puppeteer extends Helper {
 
   _session() {
     return {
-      start: async () => {
+      start: async (sessionName = '') => {
```
DavertMik: I know this was not used yet by Puppeteer but I think it is a good idea if we could pass in config as well.
For instance, in WebDriver we can pass in window size to a new session.
For Puppeteer we also have room for improvement - we can pass in mobile emulation modes, network, etc.

So I propose to have 2 arguments here: session name & config.

lib/helper/Playwright.js
```diff
@@ -282,6 +284,12 @@ class Playwright extends Helper {
     const contexts = await this.browser.contexts();
     contexts.shift();
 
+    // ensure current page is in default context
+    if (this.page) {
+      const existingPages = await this.browserContext.pages();
+      await this._setPage(existingPages[0]);
```
DavertMik: Looks like this line causes Travis to fail 
page is set at the moment it is closed.
To be explicit - running `page.$('body')` throws error, again, because page is not defined anymore.

Please fix this.

lib/helper/Puppeteer.js
ChexWarrior: Thanks for the feedback, I've added the parameters to the Puppeteer start method.

lib/helper/Playwright.js
ChexWarrior: I believe the problem was since the Helper was set to restart the `_stopBrowser()` method was being called after I tried to reset the page to the default session, moving my `_setPage` to occur after that check seems to fix the issue.

DavertMik:
Thank you!

PR #2296, opened 2020-03-30
(enh): Introduce yaml config file for semaphore 2.0

PeterNgTr:
- Providing semaphore yml

### DIFF
--- a/.semaphore/appiumTests.yml
+++ b/.semaphore/appiumTests.yml
@@ -1,32 +1,21 @@
 version: v1.0
-
-name: Appium tests for CodeceptJS
-
+name: Appium tests pipeline
 agent:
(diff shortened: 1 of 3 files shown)


PeterNgTr:
I checked this on my fork, the tests are triggered. Basically, it needs @DavertMik to convert the semaphore to v2.0 and add the saucelabs creds.

PeterNgTr:
@pablopaul it’s the semaphore 2.0, they introduced the yml config file.

pablopaul:
@PeterNgTr do we already use Semaphore v1?

PeterNgTr:
> @PeterNgTr do we already use Semaphore v1?

yeah, the current is semaphore v1

PR #2343, opened 2020-04-16
feat: add bddStep before/after step hook

wojtkowiak:
I was looking for a way to create a better BDD reporter but could not get a hook to a bdd steps. This adds a way to hook before/after a bdd step. Hope I did not mess this up as this is my first try here.

Applicable helpers:

- [ ] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] puppeteerCoverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] wdio

## Type of change

- [ ] :fire: Breaking changes
- [ ] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [X] Tests have been added

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -374,12 +374,14 @@ Available events:
 * `event.suite.before(suite)` - *async* before a suite
 * `event.suite.after(suite)` - *async* after a suite
 * `event.step.before(step)` - *async* when the step is scheduled for execution
-* `event.step.after(step)`- *async* after a step
+* `event.step.after(step)` - *async* after a step
 * `event.step.started(step)` - *sync* when step starts.
 * `event.step.passed(step)` - *sync* when step passed.
 * `event.step.failed(step, err)` - *sync* when step failed.
 * `event.step.finished(step)` - *sync* when step finishes.
 * `event.step.comment(step)` - *sync* fired for comments like `I.say`.
+* `event.bddStep.before(bddStep)` - *async* when the gherkin step is scheduled for execution
+* `event.bddStep.after(bddStep)` - *async* after a gherkin step
 * `event.all.before` - before running tests
 * `event.all.after` - after running tests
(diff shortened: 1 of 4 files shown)


DavertMik:
Yeah, actually there is a way to get BDD step out there. 
`step` object has `metaStep` property which will be either a page object or gherkin step.
this is how we present nested structure:

step -> pageObject -> BDD step

or

step -> bddStep

so you can get on upper level by checking metaStep property. 

Maybe your API is cleaner than that. I agree.

But I have the following questions:

1. why do you need that API
2. does it really work?

wojtkowiak:
> step object has metaStep property which will be either a page object or gherkin step.
Yes, but this only works when you have any `I` methods in the gherkin step, I do have some gherkin steps that only run queries and the `step.before` is not fired then (maybe I've messed sth up)

> But I have the following questions:

> why do you need that API
> does it really work?

I was lacking a report in which I would see only gherkin steps passed/failed.
It does work on my fork that is used in our tests.

DavertMik:
Hi, after thinking a lot on this, I think this is a good idea and nice implementation.
I am going to take this into CodeceptJS 3.0 branch so we could present this as major feature

ktryniszewski-mdsol:
@wojtkowiak Was also looking into better bdd reporting. Curious tho, if no helper methods such as `I` are in bdd step, how do you check pass/fail status for step? Since no event listener fired for pass/fail?

PR #2347, opened 2020-04-20
Added a change for handling http authentication

theTainted:
Added a change for handling http authentication, this now works similarly to webdriver

- This change helps in handling http (basic) authentication fro TestCafe helper

Applicable helpers:

- [ ] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [x] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] puppeteerCoverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] wdio

## Type of change

- [ ] :fire: Breaking changes
- [ ] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/helper/TestCafe.js
+++ b/lib/helper/TestCafe.js
@@ -308,8 +308,13 @@ class TestCafe extends Helper {
    * {{> amOnPage }}
    */
   async amOnPage(url) {
-    if (!(/^\w+\:\/\//.test(url))) {
-      url = this.options.url + url;
+  let split_url;
+    if (this.config.basicAuth) {
+      if (url.startsWith('/')) {
+        url = this.config.url + url;
+      }
+      split_url = url.split('//');
+      url = `${split_url[0]}//${this.config.basicAuth.username}:${this.config.basicAuth.password}@${split_url[1]}`;
     }
 
     return this.t.navigateTo(url)


lib/helper/TestCafe.js
```diff
@@ -308,8 +308,13 @@ class TestCafe extends Helper {
    * {{> amOnPage }}
    */
   async amOnPage(url) {
-    if (!(/^\w+\:\/\//.test(url))) {
```
DavertMik, line 311: this was the code of adding baseUrl to relative url but it was removed
any reason for that?

theTainted: Apologies for the delay in responding.
The code is inspired from the web-driver change that has been implemented. When 
`  if (!(/^\w+\:\/\//.test(url))) {` was present along with the change I made, then amOnPage would fail if I tired to access an url without authentication.
However now I am thinking if there is a way to use the `httpAuth({` which is present in Testcafe. Do you have any suggestions on how we can use that? if its possible then I will like to change based on that.

theTainted: @DavertMik 
```
async amOnPage(url) {
    let split_url;
    if (this.config.basicAuth) {
      if (!(/^\w+\:\/\//.test(url))) {
        url = this.options.url + url;
      }
      split_url = url.split('//');
      url = `${split_url[0]}//${this.config.basicAuth.username}:${this.config.basicAuth.password}@${split_url[1]}`;
    } else {
      url = this.options.url + url;
    }
    return this.t.navigateTo(url)
      .catch(mapError);
  }
```
i have tried including else condition in case config doesnot have a basic Authentication present, and have also included the regex u had to pass the url. If you think this is fine  I will create a new PR.

PR #2390, opened 2020-05-13
Allow retries to be seen via --debug instead of the more verbose --verbose

rctay:
- Description of this PR, which problem it solves:

  Retries are a great way to allow E2E tests to be robust against flakiness caused by things like network delays, and it would be helpful to be aware that retries are being attempted when diagnosing a failure, or even knowing about the health of one's test suite.

- ~~Resolves #issueId (if applicable)~~

Applicable helpers:

- [ ] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] puppeteerCoverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] wdio

## Type of change

- [ ] :fire: Breaking changes
- [ ] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/recorder.js
+++ b/lib/recorder.js
@@ -180,7 +180,7 @@ module.exports = {
       }
 
       return promiseRetry(Object.assign(defaultRetryOptions, retryOpts), (retry, number) => {
-        if (number > 1) output.log(`${currentQueue()}Retrying... Attempt #${number}`);
+        if (number > 1) output.debug(`${currentQueue()}Retrying... Attempt #${number}`);
 
         const retryRules = this.retries.reverse();
 
--- a/lib/recorder.js
+++ b/lib/recorder.js
@@ -1,7 +1,7 @@
 const debug = require('debug')('codeceptjs');
 const promiseRetry = require('promise-retry');
 
-const log = require('./output').log;
+const output = require('./output');
 
 let promise;
 let running = false;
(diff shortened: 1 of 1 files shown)


lib/recorder.js
```diff
@@ -1,7 +1,7 @@
 const debug = require('debug')('codeceptjs');
 const promiseRetry = require('promise-retry');
 
-const log = require('./output').log;
+const output = require('./output');
```
PeterNgTr, line 4: I think `const { log } = require('./output');` will work and you don't need to change much lines of code.

DavertMik:
Good idea! I was thinking about it as well but decided to solve it in a different manner.

In CodeceptJS 3 we will clean up `--verbose` log to provide only valuable information. 
`--debug` output will be left to helpers output (like current Url, grabber results, etc).
I mean, retries can create a mess for each step if they are logged in debug mode.
So they still will be in `verbose` but more visible to end user.

PR #2418, opened 2020-06-01
Fixed waitNumberOfVisibleElements for Webdriver

ilangv:
`waitNumberOfVisibleElements` is broken in WebDriver helpers

**Usecase:** 
When a locator returns 2 elements with 1 visible and other is not ,
Calling `I.waitNumberOfVisibleElements(locator,1)` fails everytime. 
The root cause is that the helper functions get an array of visibility booleans but does not filter them by value `true` 

Applicable helpers:

- [x] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] puppeteerCoverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] wdio

## Type of change

- [ ] :fire: Breaking changes
- [x] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/helper/WebDriver.js
+++ b/lib/helper/WebDriver.js
@@ -2133,6 +2133,7 @@ class WebDriver extends Helper {
         let selected = await forEachAsync(res, async el => el.isDisplayed());
 
         if (!Array.isArray(selected)) selected = [selected];
+        selected = selected.filter(val => val === true);
         return selected.length === num;
       }, aSec * 1000, `The number of elements (${new Locator(locator)}) is not ${num} after ${aSec} sec`);
     }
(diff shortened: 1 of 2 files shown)


lib/helper/WebDriver.js
```diff
@@ -2133,6 +2133,7 @@ class WebDriver extends Helper {
         let selected = await forEachAsync(res, async el => el.isDisplayed());
```
PeterNgTr, line 2133: I thought the visible elements are filtered here, aren't they? `res, async el => el.isDisplayed()`

ilangv: Hi @PeterNgTr . No they are not filtered here. `el.isDisplayed()` returns a Boolean and this function is missing a filter by `true` .
I noticed that the filtering is done in [grabNumberOfVisibleElements](https://github.com/codecept-js/CodeceptJS/blob/master/lib/helper/WebDriver.js#L1403) but missing for this function.

DavertMik:
Thank you! That was a good catch

PR #2481, opened 2020-07-08
tests(step-log): Custom pretty logs for method args

Vorobeyko:
If helper method gets args of object type, our logs become not readable. Like: 
```js
I wait for visible { property: 'value', ... }
```

This MR adds a feature which allows pass args with custom parse-function for print pretty args. Example:
Create function `humanizeArgs` which return some string from your value or your data. 
```js
// PageObject.js
module.exports = {
  _init() {
    this.value = 'Logs Page Value';
  },

  humanizeArgs() {
    return this.value;
  },
};
```

In Custom Helper creates function which gets object with some value.
```js
// CustomHelper.js
class CustomHelper extends Helper {
  getHumanizeArgs(objectArgs) {
    console.log(objectArgs.value);
  }
}

module.exports = CustomHelper;
```
And call this custom method on your scenario
```js
// Scenario.test.js
Feature('Filesystem');

Scenario('Print correct arg message', ({ I, LogsPage }) => {
  I.getHumanizeArgs(LogsPage);
});
```

In result we will get pretty console log, like:
```js
I get humanize args Logs Page Value
```

Instead:
```js
I get humanize args {\"value\":\"Logs Page Value\"}
```

### DIFF
--- a/lib/step.js
+++ b/lib/step.js
@@ -112,6 +112,8 @@ class Step {
         return '*****';
       } else if (arg.toString && arg.toString() !== '[object Object]') {
         return arg.toString();
+      } else if (typeof arg === 'object' && arg.humanizeArgs) {
+        return arg.humanizeArgs();
       } else if (typeof arg === 'object') {
         return JSON.stringify(arg);
       }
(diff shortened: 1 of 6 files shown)


DavertMik:
why not just implement `toString()` for this object, as it is currently supported behavior for pretty printng?

Issue #2483, opened 2020-07-10
TypeScript configuration prevents asynchronous tests from executing properly


shanplourde:
I'm using CodeceptJS (2.6.x) and TypeScript to write our tests. Asynchronous tests are not running properly with the CodeceptJS tsconfig.json. TypeScript is configured to transpile to es2015. With this setting, async and await get transpiled to generators, which the CodeceptJS test runner is not able to handle properly. Tests with generators make the test runner end tests prematurely, and test failures are not reported back to the test runner. This pretty much makes asynchronous tests completely unstable and unreliable. 

Setting tsconfig compilerOptions.target to es2017 works in that async and await remain in the transpiled code.

TypeScript code example with compilerOptions.target set to es2015:

```typescript
const foo = async() => { 

    console.log('async foo');
    return 'abc';
}

const test = foo().then(result => { 
    console.log('test result', test);
});
```

Transpiled down to es2015 code:

```javascript
"use strict";
var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, generator) {
    function adopt(value) { return value instanceof P ? value : new P(function (resolve) { resolve(value); }); }
    return new (P || (P = Promise))(function (resolve, reject) {
        function fulfilled(value) { try { step(generator.next(value)); } catch (e) { reject(e); } }
        function rejected(value) { try { step(generator["throw"](value)); } catch (e) { reject(e); } }
        function step(result) { result.done ? resolve(result.value) : adopt(result.value).then(fulfilled, rejected); }
        step((generator = generator.apply(thisArg, _arguments || [])).next());
    });
};
const foo = () => __awaiter(void 0, void 0, void 0, function* () {
    console.log('async foo');
    return 'abc';
});
const test = foo().then(result => {
    console.log('test result', test);
});
```

Transpiled down to es2017 code:

```
"use strict";
const foo = async () => {
    console.log('async foo');
    return 'abc';
};
const test = foo().then(result => {
    console.log('test result', test);
});
```

PR #2484, opened 2020-07-10
Update compilerOptions.target to es2017

shanplourde:
This resolves (#2483), since async and await are no long transpiled to ES generators

This will allow CodeceptJS TypeScript asynchronous tests to run correctly. Errors are correctly reported back to the CodeceptJS test runner with this change. Asynchronous scenarios are correctly determined as completed with this updated. 

Note that without this configuration change at this time, all TypeScript projects will not work properly if they use asynchronous Scenarios. This would impact all helpers. 

## Type of change
- [X ] :bug: Bug fix

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -10,7 +10,7 @@
     "types": [
         "node"
     ],
-    "target": "es2015"
+    "target": "es2017"
   },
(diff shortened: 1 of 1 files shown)


DavertMik:
Thank you!

PR #2494, opened 2020-07-17
Rearranged typeof check to handle null parameter

### LINKED ISSUE #2493: Invoking custom helper I.methodName(null) method resulted into crash due to a logic error in lib/step.js

I am trying to invoke a custom helper `I.methodName(null)` method with null argument to assert **null** value present in one of the json field.

#### What do you get instead?

```bash
     Cannot read property 'toString' of null
  ypeError: Cannot read property 'toString' of null
      at G:\CodeceptJS-Test\node_modules\codeceptjs\lib\step.js:115:22
      at Array.map (<anonymous>)
      at Step.humanizeArgs (node_modules\codeceptjs\lib\step.js:98:22)
      at Step.toString (node_modules\codeceptjs\lib\step.js:133:67)
      at Object.emit (node_modules\codeceptjs\lib\event.js:112:24)
```

My Helper class looks like below:

```js
// eslint-disable-next-line no-undef
const Helper     = codecept_helper;
const { expect } = require( 'chai' );

class ChaiExtraHelper extends Helper{

  assertNull( data ){

    return expect( data ).to.be.null;
  
  }

}

module.exports = ChaiExtraHelper;

```

My test looks like below:
```
I.assertNull(null);
```

### Details

* CodeceptJS version: 2.6.6
* NodeJS Version: 12.13.0
* Operating System: Windows 10
* puppeteer || webdriverio || protractor || testcafe version (if related)

sjana-aj:
Error:
Cannot read property 'toString' of null
  typeError: Cannot read property 'toString' of null
      at G:\CodeceptJS-Test\node_modules\codeceptjs\lib\step.js:115:22

Solution: Rearranged typeof check to handle crash when helper method is called with null parameter.

- Resolves #2493 .

Applicable helpers:

- [x] WebDriver
- [x] Puppeteer
- [x] Nightmare
- [x] REST
- [x] FileHelper
- [x] Appium
- [x] Protractor
- [x] TestCafe
- [x] Playwright

## Type of change

- [ ] :fire: Breaking changes
- [x] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/step.js
+++ b/lib/step.js
@@ -112,10 +112,10 @@ class Step {
         return `${arg}`;
       } else if (arg instanceof Secret) {
         return '*****';
-      } else if (arg.toString && arg.toString() !== '[object Object]') {
-        return arg.toString();
       } else if (typeof arg === 'object') {
         return JSON.stringify(arg);
+      } else if (arg.toString && arg.toString() !== '[object Object]') {
+        return arg.toString();
       }
       return arg;
     }).join(', ');
--- a/lib/step.js
+++ b/lib/step.js
@@ -95,10 +95,10 @@ class Step {
 
   /** @return {string} */
   humanizeArgs() {
-    if ( !args ) {
-      return '';
-    }
     return this.args.map((arg) => {
+      if ( !arg ) {
+        return '';
+      }
       if (typeof arg === 'string') {
         return `"${arg}"`;
       }
--- a/lib/step.js
+++ b/lib/step.js
@@ -95,6 +95,9 @@ class Step {
 
   /** @return {string} */
   humanizeArgs() {
+    if ( !args ) {
+      return '';
+    }
     return this.args.map((arg) => {
       if (typeof arg === 'string') {
         return `"${arg}"`;
@@ -112,10 +115,10 @@ class Step {
         return `${arg}`;
       } else if (arg instanceof Secret) {
         return '*****';
-      } else if (typeof arg === 'object') {
-        return JSON.stringify(arg);
       } else if (arg.toString && arg.toString() !== '[object Object]') {
         return arg.toString();
+      } else if (typeof arg === 'object') {
+        return JSON.stringify(arg);
       }
       return arg;
     }).join(', ');


lib/step.js
```diff
@@ -112,10 +112,10 @@ class Step {
         return `${arg}`;
       } else if (arg instanceof Secret) {
         return '*****';
-      } else if (arg.toString && arg.toString() !== '[object Object]') {
-        return arg.toString();
       } else if (typeof arg === 'object') {
         return JSON.stringify(arg);
+      } else if (arg.toString && arg.toString() !== '[object Object]') {
```
Vorobeyko, line 117: You can checks that arg is not null (undefined) at the beginning. 
`arg && arg.toString && arg.toString() !== '[object Object]'`
And not rearrange conditions.

lib/step.js
```diff
@@ -112,10 +112,10 @@ class Step {
         return `${arg}`;
       } else if (arg instanceof Secret) {
         return '*****';
-      } else if (arg.toString && arg.toString() !== '[object Object]') {
-        return arg.toString();
       } else if (typeof arg === 'object') {
```
Vorobeyko, line 115: And if arg is null, this line will return true, because null it is object ) 
And `humanizeArgs` return sting like `"null"`. 

Maybe we checks of null as first condition?

sjana-aj: @Vorobeyko `else if (typeof arg === 'object')` handles when arg is sent as null and `humanizeArgs` return sting `"null"` but the real issue here is that `else if (arg.toString && arg.toString() !== '[object Object]')` check comes before `else if (typeof arg === 'object')` check and breaks the condition as `arg.toString()` fails to work on `arg=null` value, so I thought of not adding any extra check and rearrange our checks a bit which will not hopefully create any side effect and handles the null condition well without having to apply any additional check. Would love to hear your feedback on this pls.

lib/step.js
sjana-aj: @Vorobeyko pls see my comment below

lib/step.js
sjana-aj: @Vorobeyko just wanted to check if you found time to look through my comment pls

DavertMik:
Sorry, @sjana-aj 
@Vorobeyko might be on vacation and I think I will need his approval to merge this PR
I hope we will do that soon

sjana-aj:
np at all @DavertMik , thanks for letting me know 👍

lib/step.js
Vorobeyko: @sjana-aj Hi, sorry for late. 
I think you must to add additionally checks on top like:
```js
if (!args) {
  return ''
}
```

And this all. 
Args never been null or undefined on line `arg.toString && arg.toString() !== '[object Object]'`

sjana-aj:
@Vorobeyko thanks for your comments. I just made the changes, can you pls check again?

lib/step.js
```diff
@@ -95,6 +95,9 @@ class Step {
 
   /** @return {string} */
   humanizeArgs() {
+    if ( !args ) {
```
Vorobeyko, line 98: this.args

sjana-aj: I see, updated method now, pls check again

Vorobeyko: hm... interested, why some CI jobs not started)

sjana-aj: yup, I noticed that too. Is there some log available to investigate the cause?

DavertMik:
Thank you!

sjana-aj:
Thank you @DavertMik 👍

Issue #2498, opened 2020-07-23
small "h" tags, unreadable


Arhell:
I suggest changing them a little to larger sizes.
https://codecept.io/plugins

PR #2501, opened 2020-07-25
update plugins page "h" tags for more readable

Arhell:
issue #2498


DavertMik:
Unfortunately, there is a problem here - plugins documentation is taken from `lib/plugin/**.js` files and concatenated
so changing this page will have no effect, you need to edit documentation for all plugins in that folder

Arhell:
@DavertMik done

DavertMik:
Thanks!

PR #2506, opened 2020-07-28
Release 2.6.7

pablopaul:


docs/helpers/Appium.md
```diff
@@ -1280,6 +1280,17 @@ I.rightClick('Click me', '.context');
 -   `locator` **([string][4] \| [object][6])** clickable element located by CSS|XPath|strict locator.
 -   `context` **([string][4]? | [object][6])** (optional, `null` by default) element located by CSS|XPath|strict locator.{{ react }} (optional, default `null`)
 
+### forceRightClick
```
pablopaul, line 1283: @DavertMik is it correct to have this new method in Appium as well? If yes, because Appium inherits all WebDriver methods?

DavertMik: No, in Appium we can't emulate clicks

Issue #2508, opened 2020-07-29
Screenshots On Failure Are Not Saved When Using REST Helper


KateKorneeva:
#### What are you trying to achieve?
Have a screenshot saved on failure https://codecept.io/basics/#screenshot-on-failure

#### What do you get instead?
Screenshot is not saved in `/output`
The bug was introduced in `2.6.7`. In `2.6.6` it works fine

### Details

* CodeceptJS version: 2.6.7
* NodeJS Version: v12.16.1

pablopaul:
@KateKorneeva must be something specific with the project configuration, here is an example with CodeceptJS v 2.6.7 which saves screenshot on failure:

Pipeline run: https://gitlab.com/paulvincent/codeceptjs-bug-report-template/-/jobs/663207933
Failure screenshot: https://gitlab.com/paulvincent/codeceptjs-bug-report-template/-/jobs/663207933/artifacts/file/ci_artifacts/Bug_1.failed.png

pablopaul:
The screenshot can´t be saved if the REST helper is configured after a helper which supports the `saveScreenshot()` method, so that it becomes unavailable.

See this failing pipeline: https://gitlab.com/paulvincent/codeceptjs-bug-report-template/-/jobs/663639282

```
 › <screenshotOnFail> Test failed, saving screenshot
 › <TypeError: helper.saveScreenshot is not a function> undefined
```

The [whole logic to determine "the current helper" in screenshotOnFail](https://github.com/codeceptjs/CodeceptJS/blob/master/lib/plugin/screenshotOnFail.js#L51) looks pretty error prone to me, and with [the latest addition](https://github.com/codeceptjs/CodeceptJS/compare/2.6.6...2.6.7#diff-058e73421cf0c9a35663d3c36775ea7cR11) of `REST` into `standardActingHelpers.js` it is not compatible anymore.

@PeterNgTr do you have an idea for a good fix? Can you help with advise @DavertMik pleae ?

PeterNgTr:
Oh no, so probably my last PR to add REST helper into that standard acting helpers. I would guess instead of adding it to the standard list, adding that REST helper into the delay plugin only. How does it sound? @pablopaul

pablopaul:
@PeterNgTr sounds good to me, thanks. Are you going to open a PR?

PeterNgTr:
> @PeterNgTr sounds good to me, thanks. Are you going to open a PR?

[...]

PR #2527, opened 2020-08-09
Fix: Fail "helper.saveScreenshot is not a function" on stepByStepReport plugin.

andrerleao:

Fix: Fail "helper.saveScreenshot is not a function" when using simultaneous:
 - helpers [REST, WebDriver]
 - plugins [autoDelay, stepByStepReport]

Original message fail: <Can't save step screenshot: TypeError: helper.saveScreenshot is not a function> undefined.

Related #2508,  #2513 

Applicable helpers:

- [x] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [x] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [x] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] puppeteerCoverage
- [ ] retryFailedStep
- [x] screenshotOnFail
- [ ] selenoid
- [x] stepByStepReport
- [ ] wdio

## Type of change

- [ ] :fire: Breaking changes
- [x] :bug: Bug fix
- [x] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [x] Tests have been added

### DIFF
--- a/lib/plugin/autoDelay.js
+++ b/lib/plugin/autoDelay.js
@@ -3,7 +3,7 @@ const store = require('../store');
 const recorder = require('../recorder');
 const event = require('../event');
 const log = require('../output').log;
-const supportedHelpers = JSON.parse(JSON.stringify(require('./standardActingHelpers')));
+const supportedHelpers = require('./standardActingHelpers').slice();
 
 const methodsToDelay = [
   'click',
--- a/lib/plugin/autoDelay.js
+++ b/lib/plugin/autoDelay.js
@@ -3,7 +3,7 @@ const store = require('../store');
 const recorder = require('../recorder');
 const event = require('../event');
 const log = require('../output').log;
-const supportedHelpers = require('./standardActingHelpers');
+const supportedHelpers = JSON.parse(JSON.stringify(require('./standardActingHelpers')));
 
 const methodsToDelay = [
   'click',


andrerleao:
If a variable by reference is changed, all other plugins will have their values ​​changed as well. Then, adding 'REST' in autoDelay.js, stepByStepReport.js will have 'REST' to.

I only changed the "supportedHelpers" to variable by value, allowing to change values only in autoDelay.js.

PeterNgTr:
oh, interesting @andrerleao. I didn’t know this 😅 .

PeterNgTr:
Probably we consider to add test for this.

lib/plugin/autoDelay.js
```diff
@@ -3,7 +3,7 @@ const store = require('../store');
 const recorder = require('../recorder');
 const event = require('../event');
 const log = require('../output').log;
-const supportedHelpers = require('./standardActingHelpers');
+const supportedHelpers = JSON.parse(JSON.stringify(require('./standardActingHelpers')));
```
PeterNgTr, line 6:
```suggestion
const supportedHelpers = require('./standardActingHelpers').slice();
```

how about using `.slice()` cause it doesn’t change the original array. :)

andrerleao: Oh! @PeterNgTr - More clean, now, I didn´t know this ! 😅

DavertMik:
We have removed REST helper from a list so this PR is not needed anymore

PeterNgTr:
hey @DavertMik regarding this `If a variable by reference is changed, all other plugins will have their values ​​changed as well. Then, adding 'REST' in autoDelay.js, stepByStepReport.js will have 'REST' to.` I thought this PR is still needed?

andrerleao:
_> We removed the REST helper from a list, so this PR is no longer needed_

Hello @DavertMik, this is the problem!

In the changes to remove the REST helper, we added this bug :'(

In the autoDelay.js file, the **supoortedHelpers** variable is declared by reference.

> If a variable by reference is changed, all other plug-ins will have their values ​​changed as well, so by adding REST helper to the autoDelay.js file, the plugin, stepByStepReport.js will have 'REST' as well.

In this PR we changed the declaration of the variable to be by **value**, so we changed the list only in this plugin.

Please reconsider this PR.****

PR #2556, opened 2020-08-24
Fix: Using Dynamic Configs Cuts WaitForTimeout by 1000 Each Time

### LINKED ISSUE #1995: Using Dynamic Configs Cuts WaitForTimeout by 1000 Each Time

We are looking to use dynamic configs to be able to test a second site within our larger suite, which has a different base URL but shares a lot of common UI elements.

#### What do you get instead?

Every time we use `.config` on a `Feature`, it causes the config to get revalidated and the `waitForTimeout` property's value appears to get divided by 1000 every time, making it so that each feature gets a smaller and smaller timeout, going from 10 sec to 0.01 sec, to 0.00001 sec, and so on.

Here's some sample output from one of our tests:

```bash
-- FAILURES:

1) Admin - FEATURE 1 NAME
  SCENARIO NAME
  element (.reveal-modal-bg) still visible after 0.01 sec

2) Admin - FEATURE 1 NAME
  SCENARIO NAME
  element (.reveal-modal-bg) still visible after 0.01 sec

3) Admin - FEATURE 2 NAME
  SCENARIO NAME
  element ({css: .fixed-width-message}) still not visible after 0.00001 sec

...

6) Admin - FEATURE 3 NAME
  SCENARIO NAME
  element ({css: .cdk-global-overlay-wrapper form}) still not visible after 1e-8 sec

...

9) Admin - FEATURE 4 NAME
  SCENARIO NAME
  element ({css: communities-list table > tbody > tr:nth-child(1)}) still not visible after 1.0000000000000001e-11 sec
```

Here is some sample source code of how we use `.config`. We confirmed that removing this fixed the issue:

```js
Feature('Admin - XXXXX').config({
    // Required for all admin tests
    url: process.env.CI ? CONSTANTS.frontendUrls.admin.internal : CONSTANTS.frontendUrls.admin.external
});
```

### Details

* CodeceptJS version: **Codecept v2.3.5**
* NodeJS Version: **Node v12.10.0**
* Operating System: **Ubuntu 18.04.03 LTS**
* Protractor || WebDriverIO || Nightmare version: **Webdriver 5.13.2**
* Configuration file: The only important bit is that under our helpers we specify:

```js
waitForTimeout: 10000
```

mirao:
- Solves cutting of `waitForTimeout` in scenario with dynamic config.
I considered fixing in helper (WebDriver/Appium/Protractor) instead of fix in listener, by skipping `config.waitForTimeout /= 1000;` if `waitForTimeout` from original config matches current value * 1000. But the current fix in listener IMHO looks better.  
The fix seems to be needed for WebDriver/Appium/Protractor only. Other helpers don't convert ms to sec in `_setConfig()`
- Resolves #1995 

Applicable helpers:

- [x] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [x] Appium
- [x] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] puppeteerCoverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] wdio

## Type of change

- [ ] :fire: Breaking changes
- [x] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [x] Tests have been added

### DIFF
--- a/lib/listener/config.js
+++ b/lib/listener/config.js
@@ -16,6 +16,11 @@ module.exports = function () {
     event.dispatcher.on(event[type].before, (context = {}) => {
       function updateHelperConfig(helper, config) {
         const oldConfig = { ...helper.options };
+        if (['WebDriver', 'Appium', 'Protractor'].includes(helper.constructor.name)) {
+          // Allows next helper._setConfig() conversion back from ms to sec
+          oldConfig.waitForTimeout *= 1000;
+        }
+
         try {
           helper._setConfig(deepMerge(deepClone(oldConfig), config));
           debug(`[${ucfirst(type)} Config] ${helper.constructor.name} ${JSON.stringify(config)}`);
--- a/test/acceptance/config_test.js
+++ b/test/acceptance/config_test.js
@@ -37,3 +37,18 @@ Scenario('change config 6 @WebDriverIO @Puppeteer @Playwright @Protractor @Night
   await new Promise(r => setTimeout(r, 50));
   return { url: 'https://github.com' };
 });
+
+const assert = require('assert');
+const webDriver = require('../codecept/lib/container').helpers('WebDriver');
+
+Scenario('Check custom waitForTimeout @WebDriverIO', () => {
+  assert.strictEqual(webDriver.options.waitForTimeout, 15);
+}).config({ waitForTimeout: 15000 });
+
+Scenario('Check default waitForTimeout @WebDriverIO', () => {
+  assert.strictEqual(webDriver.options.waitForTimeout, 1); // Check that previous scenario reset the timeout when finished
+});
+
+Scenario('Check that default waitForTimeout is still set @WebDriverIO', () => {
+  assert.strictEqual(webDriver.options.waitForTimeout, 1);
+});
--- a/test/acceptance/config_test.js
+++ b/test/acceptance/config_test.js
@@ -39,7 +39,14 @@ Scenario('change config 6 @WebDriverIO @Puppeteer @Playwright @Protractor @Night
 });
 
 const assert = require('assert');
-const webDriver = require('../codecept/lib/container').helpers('WebDriver');
+
+let container;
+try {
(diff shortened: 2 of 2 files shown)


mirao:
I don't know what to do with the failed test [Playwright Tests / build (12.x) (pull_request)](https://github.com/codeceptjs/CodeceptJS/pull/2556/checks?check_run_id=1020970129).

E.g. if I start WebDriver tests in the folder "test" using `docker-compose run --rm test-acceptance.webdriverio` then `require('../codecept/lib/container')` works well.

But if I start them from project folder as `./bin/codecept.js run -c test/acceptance/codecept.WebDriver.js --grep @WebDriverIO --debug` (similar way Playwright tests run) then `require('../codecept/lib/container')` fails and `require('../../lib/container')` must be used instead.

mirao:
> I don't know what to do with the failed test [Playwright Tests / build (12.x) (pull_request)]

Maybe I could move the definition `const webDriver = require('../codecept/lib/container').helpers('WebDriver');` to those new scenarios. It helps to fix Playwright tests.

mirao:
Fixed loading of container module in Playwright tests. The tests are still [failing](https://github.com/codeceptjs/CodeceptJS/pull/2556/checks?check_run_id=1022344990) but it's not caused by my changes, I believe.

mirao:
> Sorry, I don't like this patch. It breaks the segregation of concerns principle.

I understand. Unfortunately I'm not sure how to fix better. I'm giving a chance to another developer :smile: .
At least the tests I created might be used (if someone fix the issue itself)

DavertMik:
Yeah, let's close it for now. 
As we can't get it done without BC break

Issue #2579, opened 2020-09-11
CodeCeptJS 3.0 beta I.say I.retry (default actor methods) undefined when add steps_file


Georgegriff:
#### What are you trying to achieve?

I added a test case to show the problem:
https://github.com/codeceptjs/CodeceptJS/pull/2580

Following the docs to add
![image](https://user-images.githubusercontent.com/9056958/92914711-cab34380-f423-11ea-9572-c882d0fbd7d2.png)
With the following format
![image](https://user-images.githubusercontent.com/9056958/92914807-e0286d80-f423-11ea-9d8e-004e37ef2da4.png)

I expect to have my methods, and the default methods

#### What do you get instead?
![image](https://user-images.githubusercontent.com/9056958/92915031-17971a00-f424-11ea-8fa4-16492fd06ce8.png)

I am missing the default methods e.g. I.say. I.retry, but my methods are there
> Provide console output if related. Use `--verbose` mode for more details.

```bash
# paste output here
```

> Provide test source code if related

```js
// paste test
```

### Details

* CodeceptJS version:3.0.0-beta.4
* NodeJS Version:
* Operating System:
* puppeteer || webdriverio || protractor || testcafe version (if related)
* Configuration file:

```js
# paste config here
```

Georgegriff:
Im currently trying to reproduce this with a test case in codeceptjs repo

PR #2582, opened 2020-09-12
fix(#2579): say and retry methods now work correctly

Vorobeyko:

Applicable helpers:

- [ ] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] puppeteerCoverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] wdio

## Type of change

- [ ] :fire: Breaking changes
- [ ] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/container.js
+++ b/lib/container.js
@@ -331,6 +331,20 @@ function loadSupportObject(modulePath, supportObjectName) {
   }
   try {
     const obj = require(modulePath);
+
+    if (typeof obj === 'function' && supportObjectName === 'I') {
+      const fobj = obj();
+
+      if (fobj.constructor.name === 'Actor') {
+        const methods = getObjectMethods(fobj);
+        Object.keys(methods)
+          .forEach(key => {
+            fobj[key] = methods[key];
+          });
+
+        return methods;
+      }
+    }
     if (typeof obj !== 'function'
         && Object.getPrototypeOf(obj) !== Object.prototype
         && !Array.isArray(obj)
--- a/lib/container.js
+++ b/lib/container.js
@@ -332,7 +332,7 @@ function loadSupportObject(modulePath, supportObjectName) {
   try {
     const obj = require(modulePath);
 
-    if (typeof obj === 'function' && supportObjectName === 'I') {
+    if (typeof obj === 'function') {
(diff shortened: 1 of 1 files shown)


lib/container.js
```diff
@@ -331,6 +331,20 @@ function loadSupportObject(modulePath, supportObjectName) {
   }
   try {
     const obj = require(modulePath);
+
+    if (typeof obj === 'function' && supportObjectName === 'I') {
+      const fobj = obj();
+
+      if (fobj.constructor.name === 'Actor') {
+        const methods = getObjectMethods(fobj);
```
Georgegriff, line 339: Wondering if is it 100% required that people implementing a steps file must use the `actor` object I seem to think that it's possible to implement a steps file without doing that i.e. just return an object, and if so if we've subtly breaking people with 3.0

Georgegriff: Checked my code was wrong think maybe consumers using actor object is required like documented

Georgegriff (approved):

lib/container.js
```diff
@@ -331,6 +331,20 @@ function loadSupportObject(modulePath, supportObjectName) {
   }
   try {
     const obj = require(modulePath);
+
+    if (typeof obj === 'function' && supportObjectName === 'I') {
```
DavertMik, line 335:
```suggestion
    if (typeof obj === 'function') {
```
Do we need to hardcode `I` here?
I think not

lib/container.js
DavertMik: Good quetsion - technically yes, you can have anything. But without an actor, it doesn't have much sense

DavertMik (approved):

PR #2607, opened 2020-10-02
Fixed error handling in Scenario.js

haveac1gar:
This PR will fix next behavior:
1. Add to codeceptjs.config.js some Include with incorrect name of unexisting file like that:
```
    include: {
      someObject: 'unexisted-file.js',
    },
```
2. Use this object in Scenario **without any Data()** methods (this is important):
```
Scenario('my scenario', async ({ I, someObject }) => {
  await someObject.method()
})
```
3. Got an error - this is OK
4. Got an **exit code 0** - this is the problem.

Also, if test looks like this:
```
Scenario('my scenario', ({ I, someObject }) => {
  someObject.method()
})
```
We got **exit code 1**, what is ok.

The problem is in **getInjectedArguments**, which error doesn't throw into `recorder.catch` and `recorder.throw`.

Applicable helpers:

- [ ] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] puppeteerCoverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] wdio

## Type of change

- [ ] :fire: Breaking changes
- [x] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/scenario.js
+++ b/lib/scenario.js
@@ -67,21 +67,26 @@ module.exports.test = (test) => {
         });
       };
 
+      let injectedArguments;
       try {
-        testFn.call(test, getInjectedArguments(testFn, test)).then(() => {
-          // Promise.resolve().then(() => testFn.call(test, getInjectedArguments(testFn, test))).then(() => {
-          recorder.add('fire test.passed', () => {
-            event.emit(event.test.passed, test);
-            event.emit(event.test.finished, test);
-          });
-          recorder.add('finish test', () => done());
-          recorder.catch();
-        }).catch(catchError);
+        injectedArguments = getInjectedArguments(testFn, test);
       } catch (e) {
         catchError(e);
+        return;
       }
+
+      testFn.call(test, injectedArguments).then(() => {
+        // Promise.resolve().then(() => testFn.call(test, getInjectedArguments(testFn, test))).then(() => {
+        recorder.add('fire test.passed', () => {
+          event.emit(event.test.passed, test);
+          event.emit(event.test.finished, test);
+        });
+        recorder.add('finish test', () => done());
+        recorder.catch();
+      }).catch(catchError);
       return;
     }
+
     try {
       event.emit(event.test.started, test);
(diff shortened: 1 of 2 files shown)


DavertMik:
Sorry I can't approve it without any description

lib/scenario.js
```diff
@@ -54,7 +54,7 @@ module.exports.test = (test) => {
 
     if (isAsyncFunction(testFn)) {
       event.emit(event.test.started, test);
-      testFn.call(test, getInjectedArguments(testFn, test)).then(() => {
+      Promise.resolve().then(() => testFn.call(test, getInjectedArguments(testFn, test))).then(() => {
```
DavertMik, line 57: I don't think this is needed testFn is AsyncFunction
by definition async function returns a promise
so you dont need to start a new promise here

haveac1gar: By promisifying this call we can catch any error from **getInjectedArguments**, which is core of my problem. Instead of promisifying `testFn.call` of course we can wrap **getInjectedArguments** in try / catch block, but this is isconsistent behavior. If we do that, we also should throw error in another `recorder.throw` / `recorder.catch` block, and I dont think this is ok: we already have `recorder.throw` / `recorder.catch` block in `.catch`.

Btw, few strings down we are wrapping all `testFn.call` in try / catch with **getInjectedArguments**, so why we cant promisify it here?

haveac1gar: Added description!

DavertMik (approved):

Arhell (approved):

PR #2691, opened 2020-11-23
 Fix method loop and adding report support for run-workers command

Ameterezu:
When the method __ is called, an endless loop occurs because the added step is recursively appended to the end of the chain.
The added condition fixes method loop.

Add reporting support for run-workers command. 
Can be used to get junit reports.
Example command:
npx codeceptjs run-workers 2 --reporter mocha-junit-reporter

- Fixes [1938](https://github.com/codeceptjs/CodeceptJS/issues/1938)


DavertMik:
Thank you for this fix!

PR #2730, opened 2020-12-21
fix: catch init failures during run

wagoid:

When running run-multiple and there's an error during initialization (such as an exception when loading helpers for example), codeceptjs would return a 0 exit code, because the error becomes an unhandled promise rejection.

With this change, the error is handled by run.js, and the process returns a non 0 exit code.

## Type of change

- [x] :bug: Bug fix

## Checklist:

- [x] Tests have been added


DavertMik:
Thanks!

PR #2774, opened 2021-01-20
Add command to get list of tests in json format

Ameterezu:
## Motivation
Added the ability to get a list of tests in json format.
Filtering of tests through grep and fgrep is supported.
Export of the list to a file is supported.
New functionality will allow to control the launch of tests remotely by it's name.

**Command:**
`npx codeceptjs list-of-tests --grep "My test name" --file`

## Type of change

## Checklist:

- [ ] Tests have been added


DavertMik:
Thanks for the PR!
However, I don't think it should be part of CodeceptJS main package. It adds an independent command. I think it can be published as a standalone package on npm and added to the list of community driven packages https://github.com/codeceptjs/CodeceptJS/wiki/Community-Helpers-&-Plugins

lyubenkov:
@DavertMik Thanks for your detailed feedback, but I still don't understand why json format can't be part of the main package. 
Codeceptjs has in-build ability to export list of tests in a plain text format (dry-run). But such format can't be processed easily afterwards. Isn't it better to support json out of the box?

PR #2872, opened 2021-04-15
Add is shadow locator to playwright

patrjam:
In actual solution is a problem with use shadow elements in tests under Playwright driver, it returns error `page.$$: selector: expected string, got object`. This PR resolves this bug and returns elements to right syntax working with Playwright.

Applicable helpers:
- [x] Playwright

Applicable plugins:
NONE

## Type of change
- [x] :bug: Bug fix

## Checklist:

- [ok] Tests have been added
- [not needed] Documentation has been added (Run `npm run docs`)
- [ok] Lint checking (Run `npm run lint`)
- [ok] Local tests are passed (Run `npm test`)

### DIFF
--- a/lib/helper/Playwright.js
+++ b/lib/helper/Playwright.js
@@ -2219,6 +2219,10 @@ function buildLocatorString(locator) {
   } if (locator.isXPath()) {
     // dont rely on heuristics of playwright for figuring out xpath
     return `xpath=${locator.value}`;
+  } if (locator.isShadow()) {
+    // playwright supports specific shadow syntax without commas
+    const els = locator.value.toString();
+    return els.replace(/,/g, ' ');
   }
   return locator.simplify();
 }
--- a/package.json
+++ b/package.json
@@ -121,7 +121,7 @@
     "mocha-parallel-tests": "^2.3.0",
     "nightmare": "^3.0.2",
     "nodemon": "^1.19.4",
-    "playwright": "^1.9.1",
+    "playwright": "^1.10.0",
     "protractor": "^5.4.4",
     "puppeteer": "^8.0.0",
     "qrcode-terminal": "^0.12.0",
--- a/test/data/app/controllers.php
+++ b/test/data/app/controllers.php
@@ -175,6 +175,13 @@ class iframe_nested {
     }
 }
 
+class shadow_dom {
+    public function GET()
+    {
+        include __DIR__.'/view/shadow_dom.php';
+    }
+}
+
 class facebookController {
     function GET($matches) {
         include __DIR__.'/view/facebook.php';
--- a/test/data/app/index.php
+++ b/test/data/app/index.php
@@ -40,6 +40,7 @@ $urls = array(
     '/spinner' => 'spinner',
     '/iframe' => 'iframe',
     '/iframe_nested' => 'iframe_nested',
+    '/shadow_dom' => 'shadow_dom',
     '/dynamic' => 'dynamic',
     '/timeout' => 'timeout',
     '/download' => 'download',
--- a/test/data/app/view/shadow_dom.php
+++ b/test/data/app/view/shadow_dom.php
@@ -1,0 +1,47 @@
+<h1>Simple Shadow DOM</h1>
+
+<script>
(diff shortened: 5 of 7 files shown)


gkushang:
@patrjam Playwright has exclusive support for the Shadow DOM which won't required an array of DOM hierarchy. Shouldn't that work for you? Or are you looking for the Parity with Webdriver?

Georgegriff:
I agree playwright works automatically without special syntax CSS selector auto pierce shadow Dom in playwright

patrjam:
Sorry for misunderstanding, my idea is to have some parity with Webdriver (e.g. in our company we use Webdriver + Playwright for testing web also with shadow elements). For this case we need to have duplicate tests/code just because of different approach of syntax of shadow element use, if we want use Webdriver or Playwright. 
As another pros I see, on the first view for QEs/QAs is more clear, element is in shadow root. This PR is something like a syntax sugar for users, who uses Webdriver and Playwright both for some reason and don't want to duplicate their code just because of this. I just want to unify some standard.

And what I talk about:

Standard (only) Playwright syntax:
```
Scenario('Shadow test', async ({ I }) => {
  I.amOnPage('http://the-internet.herokuapp.com/shadowdom');
  I.seeElement('my-paragraph');
  I.waitForVisible('my-paragraph', 5);
  I.dontSee('My default text', 'my-paragraph p' );
  I.seeElement('my-paragraph span' );
  I.see("Let's have some different text!", 'my-paragraph span');
  I.see("Let's have some different text!", 'my-paragraph li' );
})
```

Versus unify "codeceptjs" syntax usable for more drivers:
```
Scenario('Shadow test', async ({ I }) => {
  I.amOnPage('http://the-internet.herokuapp.com/shadowdom')
  I.seeElement({ shadow: ['my-paragraph'] });
  I.waitForVisible({ shadow: ['my-paragraph'] }, 5);
  I.dontSee('My default text', { shadow: ['my-paragraph', 'p'] });
  I.seeElement({ shadow: ['my-paragraph', 'span'] });
  I.see("Let's have some different text!", { shadow: ['my-paragraph', 'span'] });
  I.see("Let's have some different text!", { shadow: ['my-paragraph', 'li'] });
})
```

Georgegriff:
I understand now but this needs to be thought about in terms of direction if this is desirable, @DavertMik what do you think?


As an FYI: the latest release offered another alternative for webdriver using custom selectors 

The documentation could do with being improved here, but you can implement your own custom locators in web driver https://codecept.io/locators/#custom-locators
Example shadow Dom custom locators you could import from a library I maintain https://github.com/Georgegriff/query-selector-shadow-dom/blob/d0a6384615896f9dac9bc6c6d5aa5d81ef61811e/examples/webdriverio/deeply-nested.js#L2

Georgegriff:
Also @DavertMik something we should factor into decisions around shadow Dom is that web driver are in the process of looking at adding native shadow Dom support inspired by my library :) https://github.com/webdriverio/webdriverio/pull/6709#issuecomment-821255893

gkushang:
I expected the motivation could be to get the parity with WebDriverIO. I implemented the support for `{ shadow: }` initially in CodeceptJS until I came across @Georgegriff's [query-selector-shadow-dom](https://github.com/Georgegriff/query-selector-shadow-dom/tree/d0a6384615896f9dac9bc6c6d5aa5d81ef61811e). And, now I recently committed the Parity (both Playwright & WebdriverIO) in CodeceptJS which uses above library but yet to release by @DavertMik.

You certainly do not want to implement this. Instead you can follow the below solution,

Below works for both Playwright and WebdriverIO. The Idea is to not have DOM Hierarchy (like Shadow API)
```
I.click('any-shadow-element') // works with both Playwright and WebdriverIO. No hierarchy is required. 
```

Checkout more info here: https://engineering.salesforce.com/automation-made-easy-for-web-components-shadow-roots-b23016779ce8

Working Solution: https://github.com/salesforce/codeceptjs-bdd/tree/develop/examples/webcomponents-playwright-webdriver-example#webcomponents-playwright--webdriver-e2e-example

DavertMik:
I agree with @gkushang

PR #3006, opened 2021-08-28
[Feature] Adds coverage plugin for `Playwright`

anirudh-modi:
- `Playwright` exposes its own coverage API which enables us to collect coverage, https://playwright.dev/docs/api/class-coverage#coverage-stop-js-coverage. The PR aims to enable coverage collection for `Playwright`. As there was existing code for `puppeteerCoverage`, simply enabling `Playwright` did the trick for us.. 
Hence a refactor was also added to prevent usage for existing `puppeteerCoverage` plugin

Applicable helpers:

- [ ] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] puppeteerCoverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [ ] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added


lib/plugin/playwrightCoverage.js
```diff
@@ -0,0 +1,36 @@
+const coverage = require('../coverage');
+
+/**
+ * Dumps playwright code coverage after every test.
+ *
+ * #### Configuration
+ *
+ * Configuration can either be taken from a corresponding helper (deprecated) or a from plugin config (recommended).
```
DavertMik, line 8:
```suggestion
```

This line is confusing. Better not to start with something that has an alternative on first line. 
If the heper's config is deprecated let's not mention it at all

lib/plugin/playwrightCoverage.js
```diff
@@ -0,0 +1,36 @@
...
+ * ```
+ *
+ * Possible config options:
+ *
+ * * `coverageDir`: directory to dump coverage files
+ * * `uniqueFileName`: generate a unique filename by adding uuid
+ *
+ *  First of all, your mileage may vary!
+ *
+ *  To work, you need the client javascript code to be NOT uglified. They need to be built in "development" mode.
```
DavertMik, line 25:
```suggestion
 *  To work, you need the client JavaScript code to be NOT uglified. It needs to be built in "development" mode.
```

lib/plugin/playwrightCoverage.js
```diff
@@ -0,0 +1,36 @@
...
+ *  And the end of your tests, you'll get a directory full of coverage per test run.  Now what?
```
DavertMik, line 26:
```suggestion
 *  And the end of your tests, you'll get a directory with code coverage collected per test run. 
```

lib/coverage.js
```diff
@@ -0,0 +1,126 @@
...
+      'starting coverage',
+      async () => {
+        try {
+          if (!coverageRunning) {
+            debug('--> starting coverage <--');
+            coverageRunning = true;
+            await helper.page.coverage.startJSCoverage();
+          }
+        } catch (err) {
+          console.error(err);
```
DavertMik, line 83: ok... what if page is not opened yet? what if indeed browser is not started? 
Do we really need an error printed here?

1. add a check that `page` object exists in page and `coverage` object exists there too
2. print only errors that are related to coverage... I don't know what `startJSCoverage ` can throw but maybe there is something we can handle here

DavertMik (changes requested):
Thanks, overall looks good to me. I have few questions.

Do we need PlaywrightCoverage plugin at all?
If this is 100% the same as PuppeteerCoverage the only difference there is documentation. Maybe you could unify that under `coverage` plugin and provide it as the main tool for coverage collection.

Also, we can deprecated `puppeteerCoverage` for this reason.

I would also ask you if possible to add a chapter to https://github.com/codeceptjs/CodeceptJS/blob/3.x/docs/playwright.md mentioning 

* how to start collecting code coverage
* how to convert it using Istanbul
* how to view code coverage

So to add all user friendly docs there. If possible add some pictures to text. Btw, you can use GitHub issues as storage for pictures. I would really prefer to see some pics there because covergae is a very visual thing and users better understand it if you could make a screenshot of code coverage collected.

lib/coverage.js
anirudh-modi: @DavertMik Let me revisit this entire file a little objectively...

As this was an existing code, I assumed that some of the edge cases might already have been handled.. Will do the appropriate refactor..

anirudh-modi:
@DavertMik 
> Do we need PlaywrightCoverage plugin at all?
If this is 100% the same as PuppeteerCoverage the only difference there is documentation. Maybe you could unify that under coverage plugin and provide it as the main tool for coverage collection.

We actually don't need both plugins.. as removing `puppeteerCoverage` plugin would cause breaking change, hence I kept them the same.. But this also seems to be reasonable that we can mention deprecation of `puppeteerCoverage` in favor of `coverage`

as for the new plugin does the below API works?

```
coverage: {
    enabled: true,
    helper: "Playwright" // this can be a string, which can be a single value from the list of supported plugins
}
```

docs/plugins.md
```diff
@@ -536,6 +536,41 @@ Enable it manually on each run via `-p` option:
...
+Possible config options:
+
+-   `coverageDir`: directory to dump coverage files
+-   `uniqueFileName`: generate a unique filename by adding uuid
+
+    First of all, your mileage may vary!
+
+    To work, you need the client javascript code to be NOT uglified. They need to be built in "development" mode.
+    And the end of your tests, you'll get a directory full of coverage per test run.  Now what?
+    You'll need to convert the coverage code to something istanbul can read.
```
gkushang, line 564: Would it makes sense to add the Conversion to Istanbul as well?

anirudh-modi: I am inclined that might be a fair suggestion. As, if we do so `codecept` will be exposing a single unified `istanbul` format, which I believe is widely accepted and then users can covert appropriately.

@DavertMik thoughts on this as the original implementation didn't do the same, I want to clarify if there was any reason for the same or not?

DavertMik (approved):

DavertMik:
Thanks!

PR #3010, opened 2021-08-31
screenshotOnFail plugin change: Check for helpers' types rather than names

leonardonelson91:
This change slightly modifies the way the 'screenshotOnFail' plugin initializes its code. Currently, there is a verification for the supported helpers at the plugin initialization, comparing it by name. This does not allow for custom implementations of helpers like Appium, WebDriver etc on custom helpers that extend them, i.e.

class MyHelper extends Appium

The above will not activate the screenshotOnFail plugin as its not named as 'Appium', even though it inherits all of Appium's helper (including screenshot) functions

Applicable helpers:

- [x] WebDriver
- [x] Puppeteer
- [x] Nightmare
- [ ] REST
- [ ] FileHelper
- [x] Appium
- [x] Protractor
- [x] TestCafe
- [x] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [x] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [x] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/codecept.js
+++ b/lib/codecept.js
@@ -130,16 +130,16 @@ class Codecept {
     let patterns = [pattern];
     if (!pattern) {
       patterns = [];
-      
+
       // If the user wants to test a specific set of test files as an array or string.
-      if (this.config.tests && !this.opts.features){
-        if(Array.isArray(this.config.tests)){
+      if (this.config.tests && !this.opts.features) {
+        if (Array.isArray(this.config.tests)) {
           patterns.push(...this.config.tests);
-        }else{
+        } else {
           patterns.push(this.config.tests);
         }
       }
-      
+
       if (this.config.gherkin.features && !this.opts.tests) {
         if (Array.isArray(this.config.gherkin.features)) {
           this.config.gherkin.features.forEach(feature => {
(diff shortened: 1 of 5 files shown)


DavertMik (approved):
Thanks, I like the idea. Let me thing a bit more of it if this affects other plugins, because it may

DavertMik (changes requested):
Thanks but seems to break other plugins.
Also, there might be 3rd party plugins that rely on the helper's name.

I agree that the change is needed.

But I think you should keep current behavior for BC just in case. And create a new file like 

`standardActingHelperClasses`

Then refactor plugins to use the new way of detecting a helper.

Arhell (approved):

DavertMik:
No updates for a month, unfortunately I have to close it.

PR #3034, opened 2021-09-09
docs: adding testrail-plugin report

PeterNgTr:
- Adding the testrail plugin report

## Type of change

### DIFF
--- a/docs/reports.md
+++ b/docs/reports.md
@@ -382,7 +382,9 @@ This will give you cli with steps in console and HTML report in `output` directo
 
 Testrail integration with CodeceptJS is now so seamless. The test run is created automatically afterwards. The screenshots of failed tests are also attached to test results.
 
-To use this plugin:
+Try to use [codeceptjs-testrail](https://www.npmjs.com/package/codeceptjs-testrail) plugin
+
+Install it via NPM:
 
 ```sh
 npm i codeceptjs-testrail --save
--- a/docs/reports.md
+++ b/docs/reports.md
@@ -380,125 +380,15 @@ This will give you cli with steps in console and HTML report in `output` directo
 
 ## Testrail
 
-Testrail integration with CodeceptJS is never simple like this. The test run is created automatically after the execution. The screenshots of failed tests are also attached to test results.
+Testrail integration with CodeceptJS is now so seamless. The test run is created automatically afterwards. The screenshots of failed tests are also attached to test results.
 
-![Attachemnt for failed case](http://g.recordit.co/ajaa2QRlnW.gif)
-
-Now there is new feature, add the configuration to test run of test plan
-![Attachemnt for failed case](http://g.recordit.co/uQLvQUq7cT.gif)
-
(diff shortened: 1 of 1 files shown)


Arhell (approved):

docs/reports.md
```diff
@@ -376,3 +376,129 @@ npx codeceptjs run --reporter mocha-multi
...
+
+## Testrail
+
+Testrail integration with CodeceptJS is never simple like this. The test run is created automatically after the execution. The screenshots of failed tests are also attached to test results.
+
+![Attachemnt for failed case](http://g.recordit.co/ajaa2QRlnW.gif)
+
+Now there is new feature, add the configuration to test run of test plan
+![Attachemnt for failed case](http://g.recordit.co/uQLvQUq7cT.gif)
+
```
DavertMik, line 389: Lets stop here, not going into details any further. Just provide a link to Github so all interested readers could take the steps from it. Also it will ensure that installation and configuration wont be outdated

docs/reports.md
```diff
@@ -376,3 +376,129 @@ npx codeceptjs run --reporter mocha-multi
 ```
 
 This will give you cli with steps in console and HTML report in `output` directory.
+
+
+## Testrail
+
+Testrail integration with CodeceptJS is never simple like this. The test run is created automatically after the execution. The screenshots of failed tests are also attached to test results.
```
DavertMik, line 383: Please dont use "simple" in documention. Especially with "never" 😁 try to use more neutral style

DavertMik (changes requested):
Thanks. Take a look into comments

docs/reports.md
PeterNgTr: thanks @DavertMik . I fixed this issue. But out of curiosity, why can't we use 'simple' in documentation? any context behind that?

DavertMik: I can't find the exact quote at this moment but I read a blogpost which advocated the apprach to not using words like "simple" and "easy"

1. in current context it means nothing - if it is simple the user will figure this out by following docs
2. it can't be as simple to someone as you think :smile:

DavertMik:
@PeterNgTr please include the link to your repo! So users would see where to follow the information!

PeterNgTr:
> @PeterNgTr please include the link to your repo! So users would see where to follow the information!

Just pushed a fix for it 😄. Thanks @DavertMik

docs/reports.md
PeterNgTr: > I can't find the exact quote at this moment but I read a blogpost which advocated the apprach to not using words like "simple" and "easy"
> 
> 
> 
> 1. in current context it means nothing - if it is simple the user will figure this out by following docs
> 
> 2. it can't be as simple to someone as you think :smile: 

Aha, I found this https://www.knowledgeowl.com/home/dont-say-simply-jim-fisher. I guess it's the thing you mentioned.

DavertMik: Thank you! The right one

PR #3038, opened 2021-09-13
[FEATURE] plugin screenshotOnFail should allow other file extensions

jpmschuler:
Resemble.js allows for various file extensions, while most step away from JPG due to the artefacts, PNG seems common. However resemble.js allows WEBP as well. 
While `I.saveScreenshot` and `I.seeVisualDiff `allow the file extension to be set by filename, in plugin `screenshotOnFail` this is hardcoded. I suggest to allow an override via an option key.

Using webp drastically reduces artefacts size, at best to about one third, thus the artefact size in CI/CD is improved perhaps even circumventing limits (e.g. uploading artefacts to GitLab Pages often has a size limit).

Applicable plugins:
- [X] screenshotOnFail

## Type of change

## Checklist:
- [ ] Tests have been added

(none of these work out of the box)

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -650,6 +650,7 @@ Possible config options:
 
 -   `uniqueScreenshotNames`: use unique names for screenshot. Default: false.
 -   `fullPageScreenshots`: make full page screenshots. Default: false.
+-   `imageExtension`: define image file format. Default: png.
 
 ### Parameters
 
--- a/lib/plugin/screenshotOnFail.js
+++ b/lib/plugin/screenshotOnFail.js
@@ -75,6 +75,7 @@ module.exports = function (config) {
   event.dispatcher.on(event.test.failed, (test) => {
     recorder.add('screenshot of failed test', async () => {
       let fileName = clearString(test.title);
+      const imageExtension =  options.imageExtension ? options.imageExtension : 'png';
       // This prevent data driven to be included in the failed screenshot file name
       if (fileName.indexOf('{') !== -1) {
         fileName = fileName.substr(0, (fileName.indexOf('{') - 3)).trim();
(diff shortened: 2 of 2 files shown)


lib/plugin/screenshotOnFail.js
```diff
@@ -75,16 +75,17 @@ module.exports = function (config) {
   event.dispatcher.on(event.test.failed, (test) => {
     recorder.add('screenshot of failed test', async () => {
       let fileName = clearString(test.title);
+      const imageExtension =  options.imageExtension ? options.imageExtension : 'png';
```
PeterNgTr, line 78: Shall we handle the case which user provides the invalid extension?

I guess it's better to create a list of supporting image extensions and basically we loop through that list.

docs/plugins.md
```diff
@@ -650,6 +650,7 @@ Possible config options:
 
 -   `uniqueScreenshotNames`: use unique names for screenshot. Default: false.
 -   `fullPageScreenshots`: make full page screenshots. Default: false.
+-   `imageExtension`: define image file format. Default: png.
```
PeterNgTr, line 653: Could we mention the supporting image extensions?

lib/plugin/screenshotOnFail.js
jpmschuler: Get your point, but I wonder if the list of supported extensions might be platform dependant, I couldn't find one at all to tell the truth, I just entered some by chance on my docker images and OSX locally.
As resemble.js allows to even save an image without extension at all (forgot that once on a filename) there doesn't seem to be much checks in the underlying tools.

docs/plugins.md
jpmschuler: see above

lib/plugin/screenshotOnFail.js
PeterNgTr: So what if I provide something like ffxkf as image extension? 😅

jpmschuler:
seems all dependent on Helper and OS context, can't supply working cross-helper code

PR #3051, opened 2021-09-21
Enhancement to optionally link feature files to helper files

abhimanyupandian:
When there are multiple helper files, there is no way to link feature files to the helper files and thus all hooks are executed. Which might be unnecessary in some cases.
This enhancement helps link a specific feature file to the helper file which ensures that the before/after hooks for tests and suites are executed only for that particular feature file.

Example:
The "feature" variable is the name of the feature file to which the helper is linked to.
If this variable is not set, the usual execution flow happens.

```
class MyHelper extends Helper {

    feature = "feature_one.feature"
        
    _after() {
    }
    _beforeStep(step) {
    }
    _beforeSuite(suite){
    }
}
module.exports = MyHelper;
```

Applicable helpers:

- [ ] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [ ] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/listener/helpers.js
+++ b/lib/listener/helpers.js
@@ -22,9 +22,6 @@ module.exports = function () {
   const runAsyncHelpersHook = (hook, param, force) => {
     if (store.dryRun) return;
     Object.keys(helpers).forEach((key) => {
-      if (helpers[key].feature && param.file) {
-        if (helpers[key].feature !== path.basename(param.file)) { return; }
-      }
       if (!helpers[key][hook]) return;
       recorder.add(`hook ${key}.${hook}()`, () => helpers[key][hook](param), force, false);
     });
--- a/lib/listener/helpers.js
+++ b/lib/listener/helpers.js
@@ -3,6 +3,7 @@ const container = require('../container');
 const recorder = require('../recorder');
 const store = require('../store');
 const { error } = require('../output');
+const path = require('path');
 /**
  * Enable Helpers to listen to test events
  */
@@ -21,6 +22,9 @@ module.exports = function () {
   const runAsyncHelpersHook = (hook, param, force) => {
     if (store.dryRun) return;
     Object.keys(helpers).forEach((key) => {
+      if (helpers[key]['feature'] && param.file) {
+          if (helpers[key]['feature'] != path.basename(param.file)) { return };
+      }
       if (!helpers[key][hook]) return;
       recorder.add(`hook ${key}.${hook}()`, () => helpers[key][hook](param), force, false);
     });
--- a/lib/listener/helpers.js
+++ b/lib/listener/helpers.js
@@ -1,9 +1,9 @@
+const path = require('path');
 const event = require('../event');
 const container = require('../container');
 const recorder = require('../recorder');
 const store = require('../store');
 const { error } = require('../output');
-const path = require('path');
 /**
  * Enable Helpers to listen to test events
  */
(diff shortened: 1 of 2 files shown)


abhimanyupandian:
@DavertMik @Arhell could you please review and merge this enhancements?

Arhell (commented):
cc @DavertMik

DavertMik (changes requested):
This is a good idea but a very partial implementation that might cover only your specific case.
To my understanding inside `_before` you can obtain a `test` object from params, from there you can get a feature file name and then decide on hook execution

```js
_before(test) {
  if (test.file == 'xx') {

  }
}
```

Please, could you look into this direction for a solution?
If we can reuse current interface to solve the problem it would be much better than adding a new approach

abhimanyupandian:
> This is a good idea but a very partial implementation that might cover only your specific case.
> To my understanding inside `_before` you can obtain a `test` object from params, from there you can get a feature file name and then decide on hook execution
> 
> ```js
> _before(test) {
>   if (test.file == 'xx') {
> 
>   }
> }
> ```
> 
> Please, could you look into this direction for a solution?
> If we can reuse current interface to solve the problem it would be much better than adding a new approach

Hi @DavertMik, a couple of things:
1. "a very partial implementation that might cover only your specific case." - I understand that this covers only a specific use case. But could you please elaborate on why you think this is a partial implementation?
2. "from there you can get a feature file name" - As far as I see, currently we do not set the "file" value of suites, steps or tests anywhere. This is probably a bug? This PR however has the changes for that.
3. "reuse current interface to solve the problem" - Sorry, but could you please elaborate on this too?

Just wanted to reiterate that the current PR changes will only **optionally** allow users to link the feature file to the helper. With this change we remove an overhead for the users, who will otherwise have to write the same "conditions" within all the helpers.

DavertMik:
>  currently we do not set the "file" value of suites, steps or tests anywhere

yes, this part can be fixed as PR. So the proposed solution for this problem should work like this:

```
_before(test) {
  if (test.file == 'xx') {

  }
}
```

lib/listener/helpers.js
```diff
@@ -21,6 +22,9 @@ module.exports = function () {
   const runAsyncHelpersHook = (hook, param, force) => {
     if (store.dryRun) return;
     Object.keys(helpers).forEach((key) => {
+      if (helpers[key].feature && param.file) {
```
DavertMik, line 25: Please remove this part, then I'm merging

abhimanyupandian: Sure @DavertMik! Will make the change.
However, I still believe that having this part would really add value, if it is provided out of the box by CodeceptJS.

DavertMik: I understand your point. But there must be some logic in the application's behavior. If you don't know of this feature how would you guess it exists? 

if the same result can be achieved with `_before(test) ...` better to keep only one option here, so user won't be confused by some additional magical logic...

@abhimanyupandian  Also, is possible, could you update BDD chapter in docs to describe this use case there? 
Thanks!

abhimanyupandian:
@DavertMik requested changes made. Please merge this PR.
Will add the suggested documentations too probably over the weekend.

lib/listener/helpers.js
```diff
@@ -1,3 +1,4 @@
+const path = require('path');
```
DavertMik, line 1:
```suggestion
```

DavertMik (approved):

PR #3054, opened 2021-09-27
We should always close the browser

mattonem:
Remote browser or not, we should always close the browser. 
Closing the browser will actually tell the remote browser provider that the session is to be ended (logs to be consolidated and session terminated), whereas only disconnecting might keep the session hanging for no reason.

When executing Codeceptjs/Puppeteer test suite using the browserstack grid, the session hang for 15min after the test suite finished executing and logs are not properly uploaded to the plateform. 
I don't have an issue ID as I identified the bug at the same time, I fixed it.

Applicable helpers:

- [ ] WebDriver
- [X] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [X] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added
No test added as it would require credentials for a remote browsers provider. 
`npm test` -> 2 tests fail but they fail anyway on branch `3.x` on my system

### DIFF
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -554,10 +554,9 @@ class Puppeteer extends Helper {
     this.context = null;
     popupStore.clear();
     this.isAuthenticated = false;
+    await this.browser.close();
     if (this.isRemoteBrowser) {
       await this.browser.disconnect();
-    } else {
-      await this.browser.close();
     }
   }
 
--- a/test/helper/Puppeteer_test.js
+++ b/test/helper/Puppeteer_test.js
@@ -934,15 +934,9 @@ describe('Puppeteer', function () {
 
 let remoteBrowser;
 async function createRemoteBrowser() {
-  if (remoteBrowser) {
-    await remoteBrowser.close();
-  }
   remoteBrowser = await puppeteer.launch({
     args: ['--no-sandbox', '--disable-setuid-sandbox'],
     headless: true,
-  });
-  remoteBrowser.on('disconnected', () => {
-    remoteBrowser = null;
   });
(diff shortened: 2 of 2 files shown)


Arhell (commented):
cc @DavertMik

mattonem:
Actually, I'm actually arguing with myself whether `disconnect` should be handled by the remote browser provider as a proper termination just like `close`... 
[The documentation](https://github.com/puppeteer/puppeteer/blob/main/docs/api.md#browserdisconnect) is not very clear on that point.

DavertMik:
I don't know. I would try to use it with servers like Selenoid or Moon to check what is the right command here. 
They suggest `close` even they previously used `connect`

https://github.com/aerokube/moon-cloud-puppeteer-example/blob/master/index.js#L20

DavertMik (changes requested):
@mattonem Sorry, I can't merge this. Puppeteer tests are failing. I restarted them but it didn't help. Please take a look there

DavertMik:
Thanks! Merging!

PR #2583, opened 2020-09-12
feat(timeout): Added timeouts for steps

Vorobeyko:
* feat(timeout): Added timeouts for steps

* ignore wait methods

* update playwright + add consts command for runner tests&

* draft commit

## Motivation/Description of the PR
- Description of this PR, which problem it solves
- Resolves #issueId (if applicable).

Applicable helpers:

- [ ] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] puppeteerCoverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] wdio

## Type of change

- [ ] :fire: Breaking changes
- [ ] :rocket: New functionality
- [ ] :bug: Bug fix
- [ ] :clipboard: Documentation changes/updates
- [ ] :hotsprings: Hot fix
- [ ] :hammer: Markdown files fix - not related to source code
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added
- [ ] Documentation has been added (Run `npm run docs`)
- [ ] Lint checking (Run `npm run lint`)
- [ ] Local tests are passed (Run `npm test`)

PR #3059, opened 2021-10-02
Add timeout on actor actions

nikocanvacom:
- In some situations CodeceptJS actor action hangs forever on very long or hanging helper methods. Mocha timeouts do not stop hanging action. This PR adds step level timeouts, configurable global and available runtime once off.

Applicable helpers:

- [ ] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [x] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [x] Tests have been added

### DIFF
--- a/lib/actor.js
+++ b/lib/actor.js
@@ -38,18 +38,17 @@ class Actor {
   }
 
   /**
-   * set the maximum execution time for next step,
-   * if function is called more than once for the same step force is required for timeout to be updated
+   * set the maximum execution time for the next step
    * @function
    * @param {number} timeout - step timeout in seconds
-   * @param {boolean} force - whether to update timeout when it is already set for the step, default is false
+   * @param {boolean} force - whether to update timeout when it is already set for the step, default is true
    * @return {this}
    * @inner
    */
-  limitTime(seconds, force = false) {
+  limitTime(timeout, force = true) {
     // give priority to timeout set from test code
     if (this.nextStepTimeout === undefined || force) {
-      this.nextStepTimeout = seconds * 1000;
+      this.nextStepTimeout = timeout * 1000;
     }
     return this;
   }
--- a/lib/actor.js
+++ b/lib/actor.js
@@ -45,7 +45,7 @@ class Actor {
    * @inner
    */
   limitTime(seconds) {
-    this.nextStepTimeout = seconds * 1000;
+    this.nextStepTimeout = Math.max(this.nextStepTimeout, seconds * 1000);
     return this;
   }
 
--- a/lib/actor.js
+++ b/lib/actor.js
@@ -6,7 +6,6 @@ const recorder = require('./recorder');
 const event = require('./event');
 const store = require('./store');
 const output = require('./output');
-const Config = require('./config');
 
 /**
  * @interface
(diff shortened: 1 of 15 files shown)


Arhell (approved):

nikocanvacom:
This PR happens to have some duplication with the work being done in https://github.com/codeceptjs/CodeceptJS/pull/2583 .
Also #2583 makes exception for `waitXXX` functions which is the right thing to do if you treat this change as feature rather than as a bugfix (which it is for my team).

Please read comments here and on the other PR and make your suggestions on what the minimal change is.

DavertMik:
Thanks, @nikocanvacom for the PR
This one looks leally good. #2583 seems to be abandoned so I left it opened just for inspiration. I think it worked.
I like your solution, however as it touches core I'm looking at it with caution. However, let's proceed with this. 

Please add documentation on how to use it. Then let's figure out how to add unit tests.

lib/container.js
```diff
@@ -42,7 +42,7 @@ class Container {
     this.createMocha();
     container.helpers = createHelpers(config.helpers || {});
     container.translation = loadTranslation(config.translation || null);
-    container.support = createSupportObjects(config.include || {});
+    container.support = createSupportObjects(config.include || {}, config.globalActionTimeout);
```
DavertMik, line 45: I don't think we need to change how container works here. We need this param only for `actor` so let's not pass this var here

lib/container.js
```diff
@@ -198,6 +198,7 @@ function createSupportObjects(config) {
 
   if (!config.I) {
     objects.I = require('./actor')();
+    objects.I.setTimeout(globalActionTimeout);
```
DavertMik, line 201: I think we can get this value from `config`...

lib/recorder.js
```diff
@@ -161,7 +161,7 @@ module.exports = {
    * @return {Promise<*> | undefined}
    * @inner
    */
-  add(taskName, fn = undefined, force = false, retry = undefined) {
+  add(taskName, fn = undefined, force = false, retry = undefined, timeout = 0) {
```
DavertMik, line 164: let's make `timeout` undefined as default so we won't accidentlly run it on 0 second

DavertMik (changes requested):
I took a detailed look into this PR. At first, I like the idea, I think we really need it.

However, I don't like API here:

```js
I.timeout(30).click('Hello')
```
reads really bad... Maybe we can find a proper verb here? As `I.retry().click()` reads quite fine.

Just as an idea - maybe an implementation can be made similary to tryTo Plugin https://github.com/codeceptjs/CodeceptJS/blob/codeceptjs-v3.0/lib/plugin/tryTo.js ?
If so, it would be more clear to have it as a plugin.

I also added my comments to the code. And I would love to see this PR going and merged into CodeceptJS

lib/actor.js
```diff
@@ -97,7 +125,7 @@ module.exports = function (obj = {}) {
   return actor;
 };
 
-function recordStep(step, args) {
+function recordStep(step, args, timeout) {
```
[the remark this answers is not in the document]
nikocanvacom: sure, I'll do this update

DavertMik (approved):
Thanks. Let's have it merged, I will write the documentation about this feature before the release

PR #3060, opened 2021-10-05
remove protractor support

PeterNgTr:
- https://github.com/angular/protractor/issues/5502

## Checklist:


Arhell (commented):
cc @DavertMik

DavertMik:
I absolutely agree on this but please keep `Protractor.js` file ther. We can't break compatibility until next major release.
Let's remove Protractor from docs, from tests and from other mentions but it should still be working but with deprecation message

PR #3065, opened 2021-10-06
Add type declarations for @codeceptjs/helper package

wadamek65:
This PR adds type declarations for the `@codeceptjs/helper`. I know there is a separate repo for this package but IMO it is more straightforward to keep all the types here as the other package probably doesn't make sense standalone.

This reduces the amount of errors returned by `dtslib` but there are still a few remaining that I will annotate in the comments.

I am also debating if it isn't better to also change this to an `interface` so that it could be extended in custom types, but will not allow for using visibility modifiers on the fields.

If everything looks alright, I'll change the PR status to ready for review.

Applicable helpers:

- [ ] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [ ] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [x] :nail_care: Polish code

## Checklist:

- [x] Tests have been added

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -97,28 +97,6 @@ declare namespace CodeceptJS {
   interface Globals {
     codeceptjs: typeof codeceptjs;
   }
-
-  class Helper {
-    constructor(config: unknown)
-    static _config(): unknown
-    _validateConfig(config: unknown): unknown
-    _setConfig(opts: unknown): unknown
-    _init(): unknown
-    _before(): unknown
-    _after(): unknown
-    _test(test: Mocha.Test): unknown
-    _passed(test: Mocha.Test): unknown
-    _failed(test: Mocha.Test): unknown
-    _beforeStep(step: CodeceptJS.Step): unknown
-    _afterStep(step: CodeceptJS.Step): unknown
-    _beforeSuite(suite: Mocha.Suite): unknown
-    _afterSuite(suite: Mocha.Suite): unknown
-    _finishTest(suite: Mocha.Suite): unknown
-    _useTo(description?: string, fn?: Function): void
-    get helpers(): any
-    debug(msg: string): void
-    debugSection(section: string, msg: string): void
-  }
 }
 
 // Globals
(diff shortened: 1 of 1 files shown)


typings/index.d.ts
```diff
@@ -97,6 +97,28 @@ declare namespace CodeceptJS {
...
+    _after(): unknown
+    _test(test: Mocha.Test): unknown
+    _passed(test: Mocha.Test): unknown
+    _failed(test: Mocha.Test): unknown
+    _beforeStep(step: CodeceptJS.Step): unknown
+    _afterStep(step: CodeceptJS.Step): unknown
+    _beforeSuite(suite: Mocha.Suite): unknown
+    _afterSuite(suite: Mocha.Suite): unknown
+    _finishTest(suite: Mocha.Suite): unknown
+    _useTo(description?: string, fn?: Function): void
```
wadamek65, line 117: This should probably be typed as either `ReturnType<fn>` or `unknown` since the initial implementation in `@codeceptjs/helper` package returns the returned value of this function call.

typings/index.d.ts
```diff
@@ -97,6 +97,28 @@ declare namespace CodeceptJS {
...
   }
+
+  class Helper {
+    constructor(config: unknown)
+    static _config(): unknown
+    _validateConfig(config: unknown): unknown
+    _setConfig(opts: unknown): unknown
+    _init(): unknown
+    _before(): unknown
+    _after(): unknown
```
wadamek65, line 108: These three in the test are expect to accept an argument but the base class implementation doesn't signify it.

typings/index.d.ts
DavertMik: Agreed

DavertMik:
Thanks, I like it. Please proceed

DavertMik:
Thanks, let's have it merged!

PR #3113, opened 2021-11-02
Timeouts

DavertMik:
Re-implements tests timeouts based on #3059

```js
// timeout for tests in feature to 30s
Feature('', { timeout: 30 })

// timeout a test to 20s
Scenario('', { timeout: 20 })

// timeout a step to 10s
I.limitTime(10).click();
```


nikocanvacom:
Created PR where I fixed `force` option for stepTimeout plugin and added tests for that option https://github.com/codeceptjs/CodeceptJS/pull/3117.

`force` option test is failing now and is not easy to fix because event based approach for step timeout changed the order of timeout updates. Now plugin goes first and then actor but I don't think implementation could rely on order of event listeners execution. One approach which could resolve this is to return control over selection of the right timeout to step by introducing a setter like `step.setTimeout(value, priority)`

test/helper/Puppeteer_test.js
```diff
@@ -996,7 +996,7 @@ describe('Puppeteer (remote browser)', function () {
       }
     });
 
-    it('should clear any prior existing pages on remote browser', async () => {
+    xit('should clear any prior existing pages on remote browser', async () => {
```
nikocanvacom, line 999: why `xit`?

Arhell (approved):

test/helper/Puppeteer_test.js
DavertMik: I don't know, this test constantly fails and I don't know why :no_good:

PR #3119, opened 2021-11-05
Disable step bystep screenshot after fail

ioannisChalkias:
- The stepByStep plugin takes screenshots on every step of the scenario even if the scenario failed it will append the last screenshot to every step after the failing point. So i added a new config option which turn off the saving and in effect the appeding after the failing point.

Applicable helpers:

- [ ] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [X] stepByStepReport
- [ ] stepTimeout
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [ ] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added


lib/plugin/stepByStepReport.js
```diff
@@ -20,6 +20,7 @@ const defaultConfig = {
   fullPageScreenshots: false,
   output: global.output_dir,
   screenshotsForAllureReport: false,
+  disableScreenshotOnFail: false,
```
DavertMik, line 23: I would say this must set to true as there probably not a real need to have screenshots after failure

DavertMik (changes requested):
Thanks, makes sense. Should we change default?

Arhell (commented):
and please resolve conflicts

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -25,14 +25,14 @@ Applicable plugins:
 - [ ] retryFailedStep
 - [ ] screenshotOnFail
 - [ ] selenoid
-- [ ] stepByStepReport
+- [x] stepByStepReport
```
nikocanvacom, line 28: Checklist in PR template shouldn't have anything checked. Those changes are done for each individual PR after they are created on github.

lib/plugin/stepByStepReport.js
```diff
@@ -135,10 +138,14 @@ module.exports = function (config) {
     if (isStepIgnored(step)) return;
     if (savedStep === step) return; // already saved
     // Ignore steps from BeforeSuite function
+    if (scenarioFailed && config.disableScreenshotOnFail) return;
```
nikocanvacom, line 141: What would be use case for this?

It should only save screenshot for the first failure. ? if so what's the point in execution of other tests, you could just use fail fast mode.

Mocha test runner could do that with following config:
```
  mocha: {
    bail: false,
  },

```

nikocanvacom (changes requested):

lib/plugin/stepByStepReport.js
ioannisChalkias: I wanted to keep the functionality as it is, but i think also it would make sense to change it to true as default.

lib/plugin/stepByStepReport.js
ioannisChalkias: Was the intention to only save screenshot for the first failure? Yes it was. I never used mocha I didn't knew that, but i dont think its hurting to having this option too.

nikocanvacom (approved):

DavertMik:
Thanks!

PR #3127, opened 2021-11-13
Timeout operation priority according to the source

nikocanvacom:
- Step timeout is set form multiple sources and need to be selected using following logic:
  - between step timeout set on action from code with `I.limitTime` and timeout set with configuration of stepTimeout plugin one from the code has priority unless stepTimeout.config.force===true
  - test/suite timeout counter update stepTimeout if it is smaller than stepTimeout set from other sources. 

Applicable helpers:

- [ ] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [x] stepTimeout
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [x] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/plugin/stepTimeout.js
+++ b/lib/plugin/stepTimeout.js
@@ -2,7 +2,7 @@ const event = require('../event');
 
 const defaultConfig = {
   timeout: 150,
-  force: false,
+  overrideTimeLimitsInCode: false,
   noTimeoutSteps: [
     'amOnPage',
     'wait*',
@@ -33,7 +33,7 @@ const defaultConfig = {
  * #### Configuration:
  *
  * * `timeout` - global step timeout, default 150 seconds
- * * `force` - whether to use timeouts set in plugin config to override step timeouts set in code with I.limitTime(x).action(...), default false
+ * * `overrideTimeLimitsInCode` - whether to use timeouts set in plugin config to override step timeouts set in code with I.limitTime(x).action(...), default false
  * * `noTimeoutSteps` - an array of steps with no timeout. Default:
  *     * `amOnPage`
  *     * `wait*`
@@ -49,7 +49,7 @@ const defaultConfig = {
  * plugins: {
  *     stepTimeout: {
  *         enabled: true,
- *         force: true,
+ *         overrideTimeLimitsInCode: true,
  *         noTimeoutSteps: [
  *           'scroll*', // ignore all scroll steps
  *           /Cookie/, // ignore all steps with a Cookie in it (by regexp)
@@ -85,6 +85,6 @@ module.exports = (config) => {
       }
     }
     stepTimeout = stepTimeout === undefined ? config.timeout : stepTimeout;
-    step.setTotalTimeout(stepTimeout * 1000, config.force ? 1 : 3);
+    step.setTotalTimeout(stepTimeout * 1000, config.overrideTimeLimitsInCode ? 1 : 3);
   });
 };
--- a/lib/plugin/stepTimeout.js
+++ b/lib/plugin/stepTimeout.js
@@ -2,7 +2,7 @@ const event = require('../event');
 
 const defaultConfig = {
   timeout: 150,
-  overrideTimeLimitsInCode: false,
+  overrideStepLimits: false,
   noTimeoutSteps: [
     'amOnPage',
     'wait*',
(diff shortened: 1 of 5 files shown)


DavertMik (approved):

lib/step.js
```diff
@@ -38,8 +38,34 @@ class Step {
...
+        }
+      });
+      return totalTimeout;
+    };
+    /**
+     * @method
+     * @param {number} timeout - timeout in milliseconds or 0 if no timeout
+     * @param {number} source - 0 - test/suite, 1 - config force=true, 2 - code, 3 - config force=false
+     */
+    this.setTotalTimeout = function (timeout, source = 3) {
```
DavertMik, line 66: `source` is not something that is really clear here.
should we call it `priority`?

lib/plugin/stepTimeout.js
```diff
@@ -85,6 +85,6 @@ module.exports = (config) => {
       }
     }
     stepTimeout = stepTimeout === undefined ? config.timeout : stepTimeout;
-    step.totalTimeout = stepTimeout * 1000;
+    step.setTotalTimeout(stepTimeout * 1000, config.force ? 1 : 3);
```
DavertMik, line 88: what does `config.force` do here? 
I thought it will make global step timeouts more prioritized over local step timeouts. Does it?

DavertMik (commented):
Just a question, so the timeout with the higher `source` is taken, right?
should we call it `priority` then?

lib/plugin/stepTimeout.js
nikocanvacom: `force` makes config values to override values set from code `I.limitTime(t)`

lib/step.js
nikocanvacom: yep, we could call it priority. lower priority value gives higher priority to timeout value. I thought of calling it priority but I ended up calling it source as the logic here has very specific expectations for levels/priorities for timeout values from different sources and I tried to think of scenario where some other source of timeout try to meddle with this and couldn't.

DavertMik: other plugins could also use this API to add their timeouts... So the priority is cleaner. But if lower value makes things more prioritized then let's call it `order`. In this case, `0` should be understandable to be the first value to pull out.

lib/plugin/stepTimeout.js
DavertMik: Ok, got it! But the name is very unclear... How about a more verbose name?

```
ignoreStepLimits: true
```

Arhell (approved):

lib/step.js
```diff
@@ -38,8 +38,39 @@ class Step {
...
+     * @method
+     * @returns {number|undefined}
+     */
+    this.getTimeout = function () {
+      let totalTimeout;
+      // iterate over all timeouts starting from highest values of order
+      new Map([...timeouts.entries()].sort().reverse()).forEach((timeout, order) => {
+        if (timeout !== undefined && (
+          // when orders >= 10 - timeout value overrides those set with higher order elements
+          order >= 10
```
DavertMik, line 53: this breaks my mind... Magical numbers :mage: 🔮
what the point of `10` number here?

Why not just the lower number is is more priority it gets?

lib/step.js
```diff
@@ -38,8 +38,39 @@ class Step {
...
+        }
+      });
+      return totalTimeout;
+    };
+    /**
+     * @method
+     * @param {number} timeout - timeout in milliseconds or 0 if no timeout
+     * @param {number} order - <10 - test/suite, 10-19 - config when set to override timeouts from code, 20-29 - code, 30-39 - config
+     */
+    this.setTimeout = function (timeout, order = 35) {
```
DavertMik, line 71: let's set 50 at the beginning 
so it would be the lowest priority for now
35 is also magical number

DavertMik (changes requested):

lib/step.js
nikocanvacom: Could you have a look at the update in the last commit. It resolves the problem of really magical constant 10. Now the timeout override logic changes for orders below zero. Also  as discussed introduced constants and jsdoc for them.

DavertMik (approved):
Thank you very much!

PR #3191, opened 2022-01-09
chore: pin `faker` to `5.5.3`

MichaelDeBoey:
https://snyk.io/blog/open-source-maintainer-pulls-the-plug-on-npm-packages-colors-and-faker-now-what

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,7 @@
     "eslint-plugin-mocha": "^6.3.0",
     "expect": "^26.6.2",
     "express": "^4.17.1",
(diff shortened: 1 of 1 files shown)


DavertMik:
Thanks! I just discussed that story in [my Twitter](https://twitter.com/davert) and forgot that I might have dependency here as well 😃

The good part that it is in dev dependencies only and I'm quite unsure where we actually use it...

MichaelDeBoey:
Oh sh*t this should have been a `fix` instead @DavertMik 😓

PR #3195, opened 2022-01-11
fix: config name consistency

JiLiZART:
If we look at the image below
image
All popular node and js tools use `.config` postfix for their config files.
So I think this wonderful tool need to follow this naming convention for consistency

## Type of change

- [ ] :fire: Breaking changes
- [ ] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added


Arhell (approved):

DavertMik:
The question is: should we create a new default or just add another config name option?

If just adding a new config file name as an option, this change is ok
If the idea is to bring new defaults, it should also affect `codecept init` script, documentation changes and `create-codecpet` script as well.

Quite a lot of work in this case

JiLiZART:
I think this step is first of whole migration to config consistency. I see it like this:
1. Support `conf` and `config` postfixes (we are here)
2. Create `config` by default when use `codecept init` or other tools and use it as default config file name, rename in all docs
3. Create warnings or migration tools that renames all `conf` files to `config`
4. Remove support for `conf`

JiLiZART:
Good. Well, it make sense.

JiLiZART:
@DavertMik Is there any things I need to do in this PR?
I can soon send second PR with all docs and defaults changes to new config name.

DavertMik:
So far this looks nice, 
@JiLiZART please send next PR with docs updates. But the defaults should not be changed yet to not confuse users of current setup

PR #3198, opened 2022-01-14
attach video to allure report if allure report is enabled

PeterNgTr:
- Currently we don't attach failed videos to allure report, so would be nice to see the failed videos are attached to allure report.

Applicable helpers:

- [x] Playwright

## Type of change

### DIFF
--- a/lib/helper/Playwright.js
+++ b/lib/helper/Playwright.js
@@ -1890,6 +1890,12 @@ class Playwright extends Helper {
 
     if (this.options.recordVideo && this.page.video()) {
       test.artifacts.video = await this.page.video().path();
+
+      const Container = require('../container');
+      const allureReporter = Container.plugins('allure');
+      if (allureReporter) {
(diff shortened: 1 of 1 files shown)


Arhell (approved):

Arhell (approved):

jancorvus:
@PeterNgTr , @Arhell ,
could we also cancel [this change](https://github.com/codeceptjs/CodeceptJS/blob/3.x/typings/index.d.ts#L200) because it is wrong and even this PR proves that.
I can't upgrade because of that line since I use type Mocha.test in my TSed helpers.

Thanks.

DavertMik (changes requested):
This should be part of `allure.js` plugin, not Playwright
Playwright helper should know nothing about plugins currently enabled.

That's why we have `artifacts` option.

You can attach videos from alure plugin by going through all tests and getting video artifacts.
Also, please note, that videos are processed in the very end

PR #3235, opened 2022-02-06
[Invalid PR] Added missing support to listen to step events within workers

abhimanyupandian:
Currently, only events for suite and test can be listened to when tests are run in parallel using workers.
This PR fixes the missing implementations for step events.

Applicable helpers:

- [ ] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] stepTimeout
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [x] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added


abhimanyupandian:
@DavertMik could you please review and merge this bug fix?
The UTs are failing due to someother reason and not due to the changes in this fix.

lib/command/workers/runTests.js
```diff
@@ -134,6 +167,12 @@ function initializeListeners() {
   // calculate duration
   event.dispatcher.on(event.test.started, test => test.start = new Date());
 
+  // tests
```
DavertMik: steps

DavertMik (commented):
Sorry, but what problem does this solve?

If the purpose is to send these events to print out steps - it won't work. Because if we will print all steps from all workers it will produce a total mess, so you won't be able to read the scenario, anyway/

For this case I don't see a way it can be implemented.

Maybe there are cases I'm currently missing?

abhimanyupandian:
> Sorry, but what problem does this solve?
> 
> If the purpose is to send these events to print out steps - it won't work. Because if we will print all steps from all workers it will produce a total mess, so you won't be able to read the scenario, anyway/
> 
> For this case I don't see a way it can be implemented.
> 
> Maybe there are cases I'm currently missing?

@DavertMik 

Although it is anyways not possible to print them out as we have the stdout disabled for parallel tests, the intent is not to print the steps from all workers. 

Currently, when we run tests in parallel, we are unable to capture the step events (finished/started/passed/failed). However, the events for suite and test work just fine. 

Consider the following case with the 'commentStep' plugin enabled:
```typescript
const changeABC = () => step`Save after changing Profile to ABC`;
function save() {
      __changeABC;
      I.selectOption('enable');
      I.click('Save');
      I.waitForText('Saved successfully');
}
```

**We have a plugin that keeps track of the steps that have failed** (something like {stepTitle: count}).  For non-parallel execution, this works fine. It will be useful if we can get the details of the step (here the title 'Apply Changes after Market Change' is useful) for parallel tests also, within a custom plugin just like we have for non-parallel tests.

That is what this PR intends to do.

It would be expected/nice to have the same behavior for parallel tests as we are unable to run tests in parallel due to this.

abhimanyupandian:
Hey @DavertMik, could you please help us with this PR. We are unable to fully support parallel tests due to the missing event listeners for steps in parallel tests

abhimanyupandian:
@DavertMik screwed something up as part of the commits :(
Please confirm if this PR can be deleted, will raise a support ticket for the same.

PR #3290, opened 2022-04-11
fix: missing autoLogin of supportObject

### LINKED ISSUE #3105: How to use the autoLogin() plugin with TypeScript  testing
I'm trying to use the `autoLogin` plugin within a TypeScript test suite. I copied the examples [from the documentation](https://codecept.io/plugins/#autologin), but it doesn't seem to be working. Perhaps some TS incompatibility?

#### What do you get instead?
The testing fails immediately, because function `login()` is undefined.

```bash
TSError: ⨯ Unable to compile TypeScript:oad:flatten Completed in 3ms
../../../../../../test-project/tests/e2e/tests/configuration/configuration_test.ts:6:2 - error TS2349: This expression is not callable.  
  Type 'SupportObject' has no call signatures.

6  login('user'); // login using user session
   ~~~~~

    at createTSError (test-project\tests\e2e\node_modules\ts-node\src\index.ts:750:12)
    at reportTSError (test-project\tests\e2e\node_modules\ts-node\src\index.ts:754:19)
    at getOutput (test-project\tests\e2e\node_modules\ts-node\src\index.ts:941:36)
    at Object.compile (test-project\tests\e2e\node_modules\ts-node\src\index.ts:1243:30)
    at Module.m._compile (test-project\tests\e2e\node_modules\ts-node\src\index.ts:1370:30)
    at Module._extensions..js (internal/modules/cjs/loader.js:1178:10)
    at Object.require.extensions.<computed> [as .ts] (test-project\tests\e2e\node_modules\ts-node\src\index.ts:1374:12)
    at Module.load (internal/modules/cjs/loader.js:1002:32)
    at Function.Module._load (internal/modules/cjs/loader.js:901:14)
    at Module.require (internal/modules/cjs/loader.js:1044:19)
```

```js
// Copied from https://codecept.io/plugins/#autologin
Before(login => {
	login('admin'); // login using user session
});
```

### Details

* CodeceptJS version: 3.1.3
* NodeJS Version: 7.19.1
* Operating System: Windows 10 or Ubuntu Linux
* Configuration file:

```js
require('ts-node/register');
const { setHeadlessWhen } = require('@codeceptjs/configure');
const { bootstrap } = require('./presettings.ts');

// HEADLESS=true npx codecept run

exports.config = {
	name: 'E2E Testing',
	tests: [
		'./tests/configuration/**_test.ts',
	],
	output: './output',
	helpers: {
		Playwright: {
			url: 'https://test-site.local/',
			show: true,
			browser: 'chromium',
			// Keep cookies for all tests
			keepCookies: true,
		},
		ChaiWrapper : {
			"require": "codeceptjs-chai"
		}
	},
	bootstrap: null,
	mocha: {},
	plugins: {
		allure: {

		},
		autoLogin: {
			enabled: true,
			inject: 'login',
			users: {
				admin: {
					login: (I) => {
						I.amOnPage('/login.php');
						I.fillField('log', 'test_admin_user');
						I.fillField('pwd', 'test_admin_password');
						I.click('Submit');
					},
					check: (I) => {
						I.amOnPage('/dashboard');
						I.see('Dashboard');
					},
					fetch: () => {
						// Empty function
					},
					restore: () => {
						// Empty function
					},
				}
			}
		},
		pauseOnFail: {

		},
		retryFailedStep: {
			enabled: true
		},
		tryTo: {
			enabled: true
		},
		screenshotOnFail: {
			enabled: true
		}
	},
}
```

kobenguyent:
- Resolves #3105 

## Type of change
- [x] :bug: Bug fix

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/command/definitions.js
+++ b/lib/command/definitions.js
@@ -146,9 +146,18 @@ module.exports = function (genPath, options) {
     }
   }
 
+  let autoLogin;
+  if (config.plugins.autoLogin) {
+    autoLogin = config.plugins.autoLogin.inject;
+  }
+
   const supportObject = new Map();
   supportObject.set('I', 'I');
   supportObject.set('current', 'any');
+
+  if (autoLogin) {
+    supportObject.set(autoLogin, 'any');
+  }
 
   if (customHelpers.length > 0) {
     hasCustomHelper = true;


Arhell (approved):

DavertMik (approved):

callistino:
I know is already merged but wouldn't it be better if we looped through plugins the same way it does for helpers and includes in order to include all defined plugins and not just autologin? What about if I use a custom login plugin?

DavertMik:
@callistino good idea. Do you know how to implement this in the right way? 
Pull request is very welcome!

callistino:
> @callistino good idea. Do you know how to implement this in the right way? Pull request is very welcome!

I will take a stab at it sure.

callistino:
@DavertMik I finally had time to work on this solution and it's not something we can do on the fly for plugins unfortunately. Some plugins have a `registerGlobal` config property and some don't but we have no way to know that inside the definitions generator. Others like autoLogin, inject themselves into the `SupportObject` at runtime. I think the best solution would be to provide typings for each plugin instead of trying to generate. The solution on this PR is good enough but it still doesn't provide typings for global plugins like `tryTo` for example.

PR #3316, opened 2022-05-20
improve: jsonresponse validation

kobenguyent:
- Sometimes the returned response is an array like [{ user: 1}, { user: 2}], then we cannot use some actions like https://codecept.io/helpers/JSONResponse/#seeresponsecontainsjson to verify the response.

Applicable helpers:
- [x] JSONResponse

## Type of change
- [x] :bug: Bug fix

## Checklist:

- [x] Tests have been added


Arhell (approved):

lib/helper/JSONResponse.js
```diff
@@ -245,7 +257,11 @@ class JSONResponse extends Helper {
    */
   seeResponseEquals(resp) {
     this._checkResponseReady();
-    expect(this.response.data).to.deep.equal(resp);
+    if (this.response.data.length > 0) {
+      this.response.data.forEach(data => expect(data).to.deep.equal(resp));
```
DavertMik: Sorry, can't agree on this.

This checks that data is identical, so in case of array it should be:

```js
I.seeResponseEquals([ a: 'b' ])
```

Even it doesn't work with array, the proposed solution, to check that every element of array equals to provided one won't work on arrays of size > 1. If there are 2 different items it will fail. 

So please remove this one

DavertMik (changes requested):
At first I thought maybe this PR is not so useful enough but then I looked into tests and figured out that these use cases are quite important to cover.

Please update documentation for updated methods to mention this use case with arrays.

Also remove change to `seeResponseEquals` method

lib/helper/JSONResponse.js
kobenguyent: It's done, @DavertMik

lib/helper/JSONResponse.js
```diff
@@ -206,11 +224,21 @@ class JSONResponse extends Helper {
...
+   *
+   * I.seeResponseContainsKeys(['user']);
+   * ```
+   *
    * @param {array} keys
    */
   seeResponseContainsKeys(keys = []) {
     this._checkResponseReady();
-    expect(this.response.data).to.include.keys(keys);
+    if (this.response.data.length > 0) {
```
DavertMik: better to use 
```suggestion
    if (Array.isArray(this.response.data)) {
```

lib/helper/JSONResponse.js
```diff
@@ -189,12 +198,21 @@ class JSONResponse extends Helper {
...
+   *
+   * I.dontSeeResponseContainsJson({ user: 2 });
+   * ```
    *
    * @param {object} json
    */
   dontSeeResponseContainsJson(json = {}) {
     this._checkResponseReady();
-    expect(this.response.data).not.to.deep.match(json);
+    if (this.response.data.length > 0) {
```
DavertMik: better to use 
```suggestion
    if (Array.isArray(this.response.data)) {
```

DavertMik (changes requested):
Also, it would be nice if we mention explicitly in the documentation what we do with arrays. Not only as a code examples but with explanation.

DavertMik (approved):

PR #3360, opened 2022-07-13
fix: TS complains secret args

kobenguyent:
- Resolves https://github.com/codeceptjs/CodeceptJS/pull/3322: TS2554: Expected 1 arguments, but got 2.

## Type of change
- [x] :bug: Types fixing

### DIFF
--- a/lib/secret.js
+++ b/lib/secret.js
@@ -21,7 +21,7 @@ class Secret {
    * @param {*} [fieldToHide]
    * @returns {Secret}
    */
-  static secret(secret, fieldToHide) {
+  static secret(secret, fieldToHide = null) {
     if (typeof secret === 'object') {
       const fields = Array.from(arguments);
       fields.shift();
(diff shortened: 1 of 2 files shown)


lib/secret.js
```diff
@@ -18,9 +18,10 @@ class Secret {
 
   /**
    * @param {*} secret
+   * @param {*} [fieldToHide]
    * @returns {Secret}
    */
-  static secret(secret) {
+  static secret(secret, fieldToHide) {
```
DavertMik:
```suggestion
  static secret(secret, fieldToHide = null) {
```

DavertMik (approved):

lib/secret.js
DavertMik: @PeterNgTr it is possible to use variadic arguments in this function
Do you know how to set this via type definition?

Arhell (approved):

lib/secret.js
kobenguyent: @DavertMik https://github.com/codeceptjs/CodeceptJS/pull/3360/commits/2d0140a911f478e0284ecfd578b56386e0dd3e32

PR #3368, opened 2022-07-20
fix(3367): Android Click Broken When Platform Definition Capitalized

### LINKED ISSUE #3367: Android Click Broken When Platform Definition Capitalized
I recently attempted to set up codceptjs for Android automation using the desired capabilities defined on this page (https://codecept.io/mobile/#configuring). The example capabilities are defined like this:

`helpers: {
  Appium: {
    app: "my_app.apk",
    platform: "Android",
    desiredCapabilities: {}
  }
}`

Using these capabilities, every call to I.click() would fail with a 404 error coming back from Appium. After some investigation, I found this line in [WebDriver.js](https://github.com/codeceptjs/CodeceptJS/blob/e956cb8580764f07facf7c07a315775192eee03e/lib/helper/WebDriver.js)

` const clickMethod = this.browser.isMobile && this.browser.capabilities.platformName !== 'android' ? 'touchClick' : 'elementClick';`

I validated that when I pass in a platform of "**Android**", the touch API is called and seems to be incompatible with Android automation. However, if I lower case the platform to "**android**", then the element click method is used and the automation executes properly.

Given this, the best solution would be to make this check case insensitive. If that is not possible, then the documentation (and possibly code generation) needs to be updated to reflect the case sensitive nature of this code.

kobenguyent:
- Resolves #3367 

Applicable helpers:
- [x] WebDriver
- [x] Appium

## Type of change
- [x] :bug: Bug fix

## Checklist:

### DIFF
--- a/lib/helper/WebDriver.js
+++ b/lib/helper/WebDriver.js
@@ -568,7 +568,7 @@ class WebDriver extends Helper {
     }
 
     if (this.browser.capabilities && this.browser.capabilities.platformName) {
-      this.browser.capabilities.platformName = this.browser.capabilities.platformName.lowercase();
+      this.browser.capabilities.platformName = this.browser.capabilities.platformName.toLowerCase();
     }
     return this.browser;
   }
--- a/lib/helper/WebDriver.js
+++ b/lib/helper/WebDriver.js
@@ -567,7 +567,7 @@ class WebDriver extends Helper {
       });
     }
 
-    if (this.browser.capabilities) {
+    if (this.browser.capabilities && this.browser.capabilities.platformName) {
       this.browser.capabilities.platformName = this.browser.capabilities.platformName.lowercase();
     }
     return this.browser;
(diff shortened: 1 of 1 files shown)


Arhell (approved):

lib/helper/WebDriver.js
```diff
@@ -567,6 +567,9 @@ class WebDriver extends Helper {
       });
     }
 
+    if (this.browser.capabilities && this.browser.capabilities.platformName) {
+      this.browser.capabilities.platformName = this.browser.capabilities.platformName.lowercase();
```
DavertMik:
```suggestion
      this.browser.capabilities.platformName = this.browser.capabilities.platformName.lowercase();
```
maybe it should be https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLowerCase ?

DavertMik (changes requested):

lib/helper/WebDriver.js
kobenguyent: Sorry, my bad! Updated.

DavertMik:
Thanks!

PR #3372, opened 2022-07-26
fix(3341): handleDownloads function is broken with Puppeteer 14.4 and onwards

### LINKED ISSUE #3341: codeceptjs/puppeteer@14.4.1 fails on handleDownloads()
Use handleDownloads() to download file in test with codeceptjs/puppeteer

#### What do you get instead?
On I.handleDownloads() there is the following error:
```
     this.page._client.send is not a function
      at Puppeteer.handleDownloads (node_modules/codeceptjs/lib/helper/Puppeteer.js:1072:30)
      at Step.run (node_modules/codeceptjs/lib/step.js:122:47)
      at /content/n/d3/node_modules/codeceptjs/lib/actor.js:134:23

```
### Details

* CodeceptJS version: 3.3.3
* NodeJS Version: 16.15.1
* Operating System: Linux (WSL2)
* puppeteer: 14.4.1
* Configuration file:
```js
{
  tests: ['./test/*.js', './test/testdir/*.js'],
  output: './test/output',
  helpers: {
    Puppeteer: {
      url: 'http://localhost:8080',
      show: true,
      windowSize: '1200x900',
      chrome: {
        args: ['--no-sandbox', '--window-size=1280,960'],
        defaultViewport: {
          width: 1280,
          height: 960
        }
      }
    },
    FileSystem: {},
    clickIfVisible: {
      require: './test/helpers/click-if-visible.js',
    },
  },
  include: {
    I: './steps_file.js'
  },
  bootstrap: null,
  mocha: {},
  name: 'LAP',
  plugins: {
    pauseOnFail: {},
    retryFailedStep: {
      enabled: true
    },
    screenshotOnFail: {
      enabled: true
    }
  }
}
```

My test worked fine and now it fails on 'I.handleDownloads() with the error shown above.
All I changed was the puppeteer version.

Everything is fine with puppeteer@14.3.0 (and previous versions).
Since puppeteer@14.4.0 (including current 14.4.1) the test always fails.

kobenguyent:
- Resolves #3341 

Applicable helpers:
- [x] Puppeteer

## Type of change
- [x] :bug: Bug fix

## Checklist:

### DIFF
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -38,7 +38,6 @@ let puppeteer;
 let perfTiming;
 const popupStore = new Popup();
 const consoleLogStore = new Console();
-let version;
 
 /**
  * Uses [Google Chrome's Puppeteer](https://github.com/GoogleChrome/puppeteer) library to run tests inside headless Chrome.
@@ -193,8 +192,6 @@ class Puppeteer extends Helper {
     super(config);
 
     puppeteer = requireWithFallback('puppeteer', 'puppeteer-core');
-    version = require('puppeteer/package.json').version;
-
     // set defaults
     this.isRemoteBrowser = false;
     this.isRunning = false;
@@ -1071,13 +1068,12 @@ class Puppeteer extends Helper {
       fs.mkdirSync(downloadPath, '0777');
     }
     fsExtra.emptyDirSync(downloadPath);
-    // Backwards compatibility with Puppeteer < 14.4.0
-    if (parseInt(version.split('.')[0], 10) >= 14) {
-      if (parseInt(version.split('.')[1], 10) >= 4) {
-        return this.page._client().send('Page.setDownloadBehavior', { behavior: 'allow', downloadPath });
-      }
+
+    try {
+      return this.page._client.send('Page.setDownloadBehavior', { behavior: 'allow', downloadPath });
+    } catch (e) {
+      return this.page._client().send('Page.setDownloadBehavior', { behavior: 'allow', downloadPath });
     }
-    return this.page._client.send('Page.setDownloadBehavior', { behavior: 'allow', downloadPath });
   }
 
   /**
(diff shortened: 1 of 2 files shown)


Arhell (approved):

package.json
```diff
@@ -84,13 +84,14 @@
     "ms": "^2.1.3",
     "parse-function": "^5.6.4",
     "promise-retry": "^1.1.1",
+    "read-package-json": "^5.0.1",
```
DavertMik: why do we need this package?

lib/helper/Puppeteer.js
```diff
@@ -1069,6 +1071,12 @@ class Puppeteer extends Helper {
       fs.mkdirSync(downloadPath, '0777');
     }
     fsExtra.emptyDirSync(downloadPath);
+    // Backwards compatibility with Puppeteer < 14.4.0
+    if (parseInt(version.split('.')[0], 10) >= 14) {
```
DavertMik: can we just check for >= 15 ?

also I propose a better syntax to handle version:
```js
const [majorVersion] = version.split('.');
if (majorVersion >= 15)  ...
```

DavertMik (changes requested):

package.json
kobenguyent: well, this one to read the version of any package easier

lib/helper/Puppeteer.js
kobenguyent: I'm just wanting to do the backwards compatibility, like 14.3 and backwards should be not broken. So do you think we only care the major version, don't you?

DavertMik: yes, 15.x is current one, we should not bother of some minor like 14.x

package.json
DavertMik: no, I'm against adding extra packages without any real need. 
please consider a workaround

DavertMik (changes requested):

kobenguyent:
updated @DavertMik!

DavertMik:
Thanks!

PR #3379, opened 2022-07-31
Allow Typescript Support Objects To Use Default Exports

### LINKED ISSUE #3378: Allow Typescript Support Objects To Use Default Exports

It should be possible to use `export default class` in (page) support objects when using Typescript.

#### What do you get instead?

> SomePage.methodName is not a function

### Details

_(This feature request is not environment-specific)_

I recently converted a project over to Typescript and noticed that the recommended way of setting up page objects is a bit error prone. So, I want to try improving CodeceptJS's integration with Typescript a bit.

In particular, the following example will compile fine b/c `this` is implicitly typed as `any`.

```ts
// SomePage.ts
export = {
    function clickThing() {
        I.click('Button')
        I.waitForVisible(this.thingThatNoLongerExists) // Becomes a source of future flakes
    }
}
```

In trying to convert it into a class [per the docs](https://codecept.io/pageobjects/#pageobject) and example repo...

```ts
class SomePage {
    function clickThing() {
        I.click('Button')
        I.waitForVisible(this.thingThatNoLongerExists) // Now a compiler error
    }
}

export = new SomePage();
```

I learned that a limitation of `export =` is that you can't export anything else - like other interfaces or the class itself (which is useful if you want to extend the class). While it's totally possible to split up the exports into `SomePageExports.ts` and have `SomePage.ts` import `SomePageExports.ts`, or to avoid `inject` all-together via direct imports, I want to suggest a possible improvement.

With this change, it should be possible to create support objects like so...

```ts
export default class SomePage {
    function clickThing() {
        I.click('Button')
        I.waitForVisible(this.thingThatNoLongerExists) // Now a compiler error
    }
}
```

Here's a PR with the actual proposed changes... #3379

Watercycle:
This PR resolves #3378 by making it possible to use `export default class` in support objects instead of only `export =` and `module.exports`.

**!!! Breaking Change**: The current recommended solution would require that the function property `default` be removed from any existing page objects.

---

Applicable helpers:

_(This is a core change that could impacts all helpers, but shouldn't)_

- [ ] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

_(This is a core change that could impacts all plugins, but shouldn't)_

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] stepTimeout
- [ ] wdio
- [ ] subtitles

## Type of change

- [x] :fire: Breaking changes
- [ ] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [x] Tests have been added
- [x] Documentation has been added (Run `npm run docs`) _(I'm not sure where/if it's worth adding anything)_

### DIFF
--- a/lib/container.js
+++ b/lib/container.js
@@ -1,7 +1,7 @@
 const glob = require('glob');
 const path = require('path');
 const { MetaStep } = require('./step');
-const { fileExists, isFunction, isAsyncFunction } = require('./utils');
+const { fileExists, isFunction, isClass, isAsyncFunction } = require('./utils');
 const Translation = require('./translation');
 const MochaFactory = require('./mochaFactory');
 const recorder = require('./recorder');
@@ -335,7 +335,18 @@ function loadSupportObject(modulePath, supportObjectName) {
     modulePath = path.join(global.codecept_dir, modulePath);
   }
   try {
-    const obj = require(modulePath);
+    let obj = require(modulePath);
+
+    // Allows projects to use ES6 `default export` so that multiple exports
+    // may be used (which is common in Typescript).
+    if (typeof obj === 'object' && obj.hasOwnProperty('default')) {
+      obj = obj.default;
+    }
+
+    // Automatically instantiate classes.
+    if (isClass(obj)) {
+      obj = new obj();
+    }
 
     if (typeof obj === 'function') {
       const fobj = obj();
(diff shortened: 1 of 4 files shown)


Watercycle:
I still need to QA this a little closer, run tests, update the docs before marking it for review.

Watercycle:
(See issue)

PR #3317, opened 2022-05-20
use W3C standards when interfacing with appium.

mikk150:
It may still be broken when interfacing Ios, because I have no idea what
IOS platform name is

## Motivation/Description of the PR
- Description of this PR, which problem it solves
- Resolves #2904 (if applicable).

Applicable helpers:

- [x] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] stepTimeout
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [ ] :rocket: New functionality
- [x] :bug: Bug fix
- [ ] :clipboard: Documentation changes/updates
- [ ] :hotsprings: Hot fix
- [ ] :hammer: Markdown files fix - not related to source code
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added
- [ ] Documentation has been added (Run `npm run docs`)
- [ ] Lint checking (Run `npm run lint`)
- [ ] Local tests are passed (Run `npm test`)

PR #3414, opened 2022-09-19
check if browser is W3C instead of Android

mikk150:
My first implementation to this was stupid, this is way better one
- Resolves #3317

Applicable helpers:

- [x] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] stepTimeout
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [ ] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [x] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/helper/WebDriver.js
+++ b/lib/helper/WebDriver.js
@@ -916,7 +916,7 @@ class WebDriver extends Helper {
    * {{ react }}
    */
   async click(locator, context = null) {
-    const clickMethod = this.browser.isMobile && this.browser.capabilities.platformName !== 'android' ? 'touchClick' : 'elementClick';
+    const clickMethod = this.browser.isMobile && !this.browser.isW3C ? 'touchClick' : 'elementClick';
     const locateFn = prepareLocateFn.call(this, context);
 
     const res = await findClickable.call(this, locator, locateFn);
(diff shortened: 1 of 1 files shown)


kobenguyent:
Hi @mikk150 so this check would be sufficient for both android and iOS right? Just out of curiosity.

mikk150:
> Hi @mikk150 so this check would be sufficient for both android and iOS right? Just out of curiosity.

if webdriverio reports iOS as W3C browser(Which it should, because it is really Appium), then I do not see why not

Arhell (approved):

DavertMik:
Thanks!

mjalav:
I did check this change with codeceptjs v3.6.7 and seems it's not working. I have native android app and when I open the app and check  "this.browser.isW3C" it's false and "this.browser.isMobile" is true. base on new condition it will do touchClick which gave me error:
ERROR webdriver: WebDriverError: The requested resource could not be found, or a request was received using an HTTP method that is not supported by the mapped resource when running "touch/click" with method "POST" and args "{"element":"00000000-0000-010f-ffff-ffff00000031"}"

I reverted your code changes locally and I see it perform element click action which pass my test. 

My question is about how "this.browser.isW3C" should drive a decision for touchClick or elementClick. seems like unrelated and make error.
Would you please reconsider this change?
(Fail) New changes:     const clickMethod = this.browser.isMobile && !this.browser.isW3C ? 'touchClick' : 'elementClick';
(pass) Old changes:   const clickMethod = this.browser.isMobile && this.browser.capabilities.platformName !== 'android' ? 'touchClick' : 'elementClick';

PR #3428, opened 2022-09-28
feat: adding GA to init script

kobenguyent:
Screen Shot 2022-09-28 at 11 20 48 AM


Arhell (approved):

DavertMik (changes requested):
1. this is possible only for Playwright / Puppeteer and maybe TestCafe as webdriver will require more sophisticated setup  
2. if a Playwright option is selected a command to install Playwright should be added  

```
    - name: Install deps
      run: npx playwright install-deps
```

if a user selects WebDriver or Appium - do not propose to create this file. As it won't be working

lib/command/init.js
```diff
@@ -245,6 +254,33 @@ module.exports = function (initPath) {
...
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js 14.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14
+      - run: npx playwright install-deps
+      - run: npm i
+      - run: npm run test
```
DavertMik: I would prefer to use `npx codeceptjs run` here to ensure that CodeceptJS is executed

lib/command/init.js
```diff
@@ -245,6 +254,33 @@ module.exports = function (initPath) {
         }
       }
 
+      if (installGitHubActions === true) {
+        const ghaYml = `name: Run Tests
```
DavertMik: can we move this to the header of the file?

lib/command/init.js
```diff
@@ -107,6 +108,13 @@ module.exports = function (initPath) {
       choices: helpers,
       message: 'What helpers do you want to use?',
     },
+    {
+      type: 'confirm',
+      name: 'installGitHubActions',
+      message: 'Add a GitHub Actions workflow?',
+      default: false,
+      when: (answers) => ['Playwright', 'Puppeteer'].includes(answers.helper) === true,
```
DavertMik, line 117: I don't think this will work for Puppeteer

DavertMik (changes requested):

lib/command/init.js
kobenguyent: https://github.com/codeceptjs/CodeceptJS/blob/3.x/.github/workflows/puppeteer.yml

I guess this shall work cause checking this file not any additional libs should be installed for puppeteer.

lib/command/init.js
kobenguyent: are you mentioning that moving this step?

lib/command/init.js
DavertMik: But we have 

```
npx playwright install-deps
```

DavertMik:
Updating documentation on initialization got me feeling that init script is already overfilled with lots of questions and it would be hard to answer all of them from start

Maybe we could move CI setup into other script? For instance in `generate.js`
So please @PeterNgTr if you want to keep it, move it to this file

Actually, you don't need CI setup until you have a set of tests written
So this probably should not be part of initial setup

PR #3434, opened 2022-09-30
fix(REST): a secret payload instance is converted to string

### LINKED ISSUE #3394: Rest Helper converts the string data payload to a JSON object when using the Secret.
I'm trying to hide sensitive data in the HTTP payload when using the Rest Helper's SendPostRequest method. The data payload is a string (URL encoded format). 

It creates a new JSON object from the data payload string's each character. 
For example;
data = "user_name=test&password=1234"
converted data payload by Rest Helper: {"0":"u","1":"s","2":"e","3":"r"...}

_test.js
```js
let response = await I.sendPostRequest('http://example.com/api/auth',secret("user_name=test&password=pass123"), {"Content-Type": "application/x-www-form-urlencoded"});
console.log(response);
```

output
```console
...
config: {
   method: 'post',
   data: '{"0":"u","1":"s","2":"e","3":"r"...}'
}
...
```

I debugged the code and I might found something. The executeRequest method is checking the request.data object's instance type. (request.data is instanceof Secret). Also there is one more check on the following line.

```js
if (request.data instanceof Secret) {
      _debugRequest.data = '*****';
      request.data = typeof request.data === 'object' ? { ...request.data.toString() } : request.data.toString();
    }
```

If I'm not wrong, to handle a JSON object and its fields, the method looks checking the request.data. But it always going to be an object because the secret is an object instance as well. In my case, it's a secret object, not a JSON data payload. I think this assumption is the root cause of the problem.

So, I modified the code on my local as you can see below. I made the change on the request.data type check and It works well both secret wrapped and non-secret, JSON and String data payloads. It's not an ideal solution but I'm sharing it just for explanation. 

```js
request.data = (typeof request.data === 'object' && Object.keys(request.data)[0]!='_secret') ? { ...request.data.toString() } : request.data.toString();
```

### Details

* CodeceptJS version: 3.3.4
* NodeJS Version: 18.3.4
* Operating System: macOS 12.4
* Playwright
* Configuration file:

```js
exports.config = {
  tests: './*_test.js',
  timeout: 30000,
  output: '/output',
  helpers: {
  REST{
    endpoint: 'https://'
  }, 
  Playwright: {
      url: "http://localhost",
      show: false,
      browser: 'chromium'
    }
  },
  include: {
   I: './steps_file.js'
  },
  bootstrap: null,
  mocha: {},
  name: "codecept-rest-test"
}```

kobenguyent:
- Resolves #3394 

Applicable helpers:
- [x] REST

## Type of change
- [x] :bug: Bug fix

## Checklist:
- [x] Tests have been added

### DIFF
--- a/lib/helper/REST.js
+++ b/lib/helper/REST.js
@@ -138,7 +138,7 @@ class REST extends Helper {
 
     if (request.data instanceof Secret) {
       _debugRequest.data = '*****';
-      request.data = typeof request.data === 'object' ? { ...request.data.toString() } : request.data.toString();
+      request.data = (typeof request.data === 'object' && request.data instanceof Secret !== true) ? { ...request.data.toString() } : request.data.toString();
     }
 
     if ((typeof request.data) === 'string') {
--- a/test/rest/REST_test.js
+++ b/test/rest/REST_test.js
@@ -77,6 +77,14 @@ describe('REST', () => {
       secretData.toString().should.include('"password":"****"');
     });
 
+    it('should send POST requests with secret form encoded is not converted to string', async () => {
+      const secretData = secret('name=john&password=123456');
+      const response = await I.sendPostRequest('/user', secretData);
+      response.data.name.should.eql('john');
+      response.data.password.should.eql('123456');
+      secretData.getMasked().should.eql('*****');
+    });
+
     it('should send PUT requests: payload format = json', async () => {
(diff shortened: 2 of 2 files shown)


Arhell (approved):

DavertMik:
I merged it directly. Thanks

PR #3442, opened 2022-10-10
Typo fix for waitForTimeout in Webdriver config

himanshuapril1:
- Currently there is a typo in regards to Webdriver configuration where `waitforTimeout` is quoted as `waitForTimeout` with caps `F`. However as per [Webdriver documentation](https://webdriver.io/docs/timeouts/#waitfor-timeout) it should be small `f`.
> WebdriverIO provides multiple commands to wait on elements to reach a certain state (e.g. enabled, visible, existing). These commands take a selector argument and a timeout number, which determines how long the instance should wait for that element to reach the state. The `waitforTimeout` option allows you to set the global timeout for all `waitFor*` commands, so you don't need to set the same timeout over and over again. (**Note the lowercase f!**)

Applicable helpers:

- [x] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] stepTimeout
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [x] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [x] Tests have been added

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -87,7 +87,7 @@ exports.config = {
       key: process.env.CLOUDSERVICE_KEY,
 
       coloredLogs: true,
-      waitForTimeout: 10000
+      waitforTimeout: 10000
     }
   },
 
--- a/docs/helpers/WebDriver.md
+++ b/docs/helpers/WebDriver.md
@@ -41,7 +41,7 @@ Type: [object][16]
 -   `keepBrowserState` **[boolean][29]?** keep browser state between tests when `restart` is set to false.
 -   `keepCookies` **[boolean][29]?** keep cookies between tests when `restart` set to false.
 -   `windowSize` **[string][17]?** default window size. Set to `maximize` or a dimension in the format `640x480`.
--   `waitForTimeout` **[number][20]?** sets default wait time in _ms_ for all `wait*` functions.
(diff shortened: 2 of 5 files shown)


Arhell (commented):
@DavertMik

kobenguyent (approved):

DavertMik (changes requested):
This can be a breaking change.

Can we just add alias so `waitForTimeout` to be the same as `waitforTimeout` ?

Personally I would prefer to use waitForTimeout as it seems more native to me and less confusing

himanshuapril1:
> This can be a breaking change.
> 
> Can we just add alias so `waitForTimeout` to be the same as `waitforTimeout` ?
> 
> Personally I would prefer to use waitForTimeout as it seems more native to me and less confusing

I agree on the same, however if adding alias makes sense then I'd go for that.

PR #3443, opened 2022-10-10
add default value to profile env variable

### LINKED ISSUE #3339: `process.env.profile` is the string "undefined" when no --profile is specified

I wanted to control what browser is run using `--profile` and fallback to chrome if `--profile` was not provided.

#### What do you get instead?

I got errors like "Error: Cannot find the browser. "undefined:headless" is neither a known browser alias, nor a path to an executable file.".

I tried debugging by putting `console.log(process.env.profile, typeof process.env.profile);` in my codecept.conf.js and got "undefined string".

### Details

* CodeceptJS version: 3.3.3
* NodeJS Version: v16.15.1
* Operating System:
* puppeteer || webdriverio || testcafe version (if related): testcafe@1.19.0
* Configuration file:

```js
require('ts-node/register');
const {
  setHeadlessWhen,
  setBrowser,
  setTestHost,
} = require('@codeceptjs/configure');

console.log(process.env.profile, typeof process.env.profile);
setHeadlessWhen(process.env.CI);
setBrowser(process.env.profile || 'chrome');
setTestHost(process.env.TEST_HOST);

exports.config = {
  tests: './tests/e2e/codecept/features/*.ts',
  output: './tests/e2e/codecept/output',
  helpers: {
    TestCafe: {
      url: process.env.BASE_URL || 'http://localhost:3000',
    },
  },
  include: {
    // I: './tests/e2e/codecept/custom-steps.ts',
  },
  bootstrap: null,
  mocha: {},
  name: 'seymour-portal',
};

```

### Investigation

I did a bit of debugging and it looks like the issue is the way that the run command (and other commands) set the profile.

https://github.com/codeceptjs/CodeceptJS/blob/58bf5d907a20e30075ddeaaac5c8df94f986cd9f/lib/command/run.js#L8-L11

It appears that when you assign to `process.env.profile` Node stringifies the value. So if `options.profile` is undefined then it gets stringified to "undefined" instead of resulting in the env not being set.

dwentland24:
Assignment of `undefined` profile to env variable with `process.env.profile` resulted in a string of `"undefined"`.
This results in mysterious behaviors with code like ` const profile = process.env.profile || 'chrome'`. The env variable cannot evaluate to `false` as it holds the string `undefined`. 

This PR sets as empty string as default value, so not setting any profile in the command results in an empty string of the profile env variable.

Resolves #3339

Applicable helpers:

- [ ] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] stepTimeout
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [ ] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/command/run.js
+++ b/lib/command/run.js
@@ -8,7 +8,11 @@ module.exports = async function (test, options) {
   // registering options globally to use in config
   // Backward compatibility for --profile
   process.profile = options.profile;
-  process.env.profile = options.profile || '';
+
+  if (options.profile) {
+    process.env.profile = options.profile;
+  }
+
   const configFile = options.config;
 
   let config = getConfig(configFile);
--- a/lib/command/run.js
+++ b/lib/command/run.js
@@ -8,7 +8,7 @@ module.exports = async function (test, options) {
   // registering options globally to use in config
   // Backward compatibility for --profile
   process.profile = options.profile;
-  process.env.profile = options.profile;
+  process.env.profile = options.profile || '';
   const configFile = options.config;
 
   let config = getConfig(configFile);


kobenguyent (approved):

Arhell (approved):

kobenguyent (approved):

mirao:
Wouldn't be better to return value `undefined` (of type `undefined`) instead of `''` (an empty string)?
I know that `const profile = process.env.profile || 'chrome'` works the same way for `undefined` or `''`.
But e.g. if you use `if (process.env.profile === undefined)` it won't work with your solution.

dwentland24:
@mirao Yes, you are right. 

I would propose to set the env variable only if `options.profile` is not empty. So we have an unset env variable if `options.profile` is empty.

DavertMik (approved):
Thank you!

PR #3459, opened 2022-10-15
fix(init): step_file is not created and resolves custom steps are not showing

kobenguyent:
- A false condition is used to create step_file, hence this file is not created when TS is chosen.
- Fixing the custom steps are not showing.

## Type of change
- [x] :bug: Bug fix

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/command/init.js
+++ b/lib/command/init.js
@@ -168,12 +168,11 @@ module.exports = function (initPath) {
 
     const finish = async () => {
       // create steps file by default
-      if (!isTypeScript) { // no extra step file for typescript (as it doesn't match TS conventions)
-        const stepFile = `./steps_file.${extension}`;
-        fs.writeFileSync(path.join(testsPath, stepFile), defaultActor);
-        config.include.I = stepFile;
-        print(`Steps file created at ${stepFile}`);
-      }
+      // no extra step file for typescript (as it doesn't match TS conventions)
+      const stepFile = `./steps_file.${extension}`;
+      fs.writeFileSync(path.join(testsPath, stepFile), defaultActor);
+      config.include.I = isTypeScript === true ? './steps_file' : stepFile;
+      print(`Steps file created at ${stepFile}`);
 
       let configSource;
       const hasConfigure = isLocal && !initPath;


lib/command/init.js
```diff
@@ -168,12 +168,11 @@ module.exports = function (initPath) {
 
     const finish = async () => {
       // create steps file by default
-      if (!isTypeScript) { // no extra step file for typescript (as it doesn't match TS conventions)
-        const stepFile = `./steps_file.${extension}`;
-        fs.writeFileSync(path.join(testsPath, stepFile), defaultActor);
-        config.include.I = stepFile;
-        print(`Steps file created at ${stepFile}`);
-      }
+      // no extra step file for typescript (as it doesn't match TS conventions)
```
ghost, line 171: Is this comment still accurate, or should it be removed?

kobenguyent: I'm not sure why it's there in the past. Perhaps not relevant anymore.

ghost: Hmm, yeah, the PR/commit that introduced it doesn't seem enlightening enough to make any more sense of it

https://github.com/codeceptjs/CodeceptJS/commit/9d2ead7e7947518de9b044b3996811f1559cb14f#diff-ee8f2604bc47be2c76c3723717364e9e20858f37cf5ea2296087438abdb48dffR174

Feel free to resolve this thread as you you see fit! 😄

Arhell (approved):

lib/command/init.js
DavertMik: it is still valid as TS doesn't like how we add extra functions into actor

DavertMik:
At this moment the format of passing custom actions into I is not really TS-friendly, so I don't recommend creating this file for TS setup:

```js
module.exports = function() {
  return actor({

    // Define custom steps here, use 'this' to access default methods of I.
    // It is recommended to place a general 'login' function here.

  });
}

```

this is not something that TS encourages to do

it would be better for TS to have a format where each step definition would be exported using `export function()` and so on

kobenguyent:
link

hey @DavertMik may you elaborate more, cause so far current step_file.ts with module.exports is still good.

DavertMik:
@PeterNgTr ok, so TS is smarter than me 😁
If so I'm ok to merge it

PR #3464, opened 2022-10-19
Fixed issue of after all event getting called after complete execution

jain-neeeraj:
- With the recent fixes in run-rerun plugin events `event.all.*` started getting triggered after the completion of `minSuccess` or `maxReruns` . This makes plugins listening to `event.all.*` behave weirdly. With this PR, an attempt is to restore the original behavior to emit these events after the execution of each test run. 
-

Applicable helpers:

- [ ] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] stepTimeout
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [X] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/rerun.js
+++ b/lib/rerun.js
@@ -20,14 +20,19 @@ class CodeceptRerunner extends BaseCodecept {
         }
         mocha.files = mocha.files.filter(t => fsPath.basename(t, '.js') === test || t === test);
       }
+      const done = (failures) => {
+        event.emit(event.all.result, this);
+        event.emit(event.all.after, this);
+        if (failures === 0) {
+          // @ts-ignore
+          resolve();
+        } else {
+          reject(new Error(`${failures} tests fail`));
+        }
+      };
       try {
-        mocha.run((failures) => {
-          if (failures === 0) {
-            resolve();
-          } else {
-            reject(new Error(`${failures} tests fail`));
-          }
-        });
+        event.emit(event.all.before, this);
+        mocha.run((failures) => done(failures));
       } catch (e) {
         reject(e);
       }
@@ -62,17 +67,12 @@ class CodeceptRerunner extends BaseCodecept {
   }
 
   run(test) {
-    const done = () => {
-      event.emit(event.all.result, this);
-      event.emit(event.all.after, this);
-    };
-    event.emit(event.all.before, this);
     return this.runTests(test)
       .then(() => {
-        this.teardown(done);
+        this.teardown();
       })
       .catch((e) => {
-        this.teardown(done);
+        this.teardown();
         throw e;
       });
   }
--- a/lib/rerun.js
+++ b/lib/rerun.js
@@ -20,19 +20,14 @@ class CodeceptRerunner extends BaseCodecept {
         }
         mocha.files = mocha.files.filter(t => fsPath.basename(t, '.js') === test || t === test);
       }
-      const done = (failures) => {
-        event.emit(event.all.result, this);
-        event.emit(event.all.after, this);
-        if (failures === 0) {
-          // @ts-ignore
-          resolve();
-        } else {
-          reject(new Error(`${failures} tests fail`));
-        }
-      };
       try {
-        event.emit(event.all.before, this);
-        mocha.run((failures) => done(failures));
+        mocha.run((failures) => {
+          if (failures === 0) {
+            resolve();
+          } else {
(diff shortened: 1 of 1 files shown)


kobenguyent:
hey @jain-neeeraj may you find some times to add tests for it? https://github.com/codeceptjs/CodeceptJS/blob/3.x/test/runner/run_rerun_test.js

DavertMik:
Maybe there should be events triggered for all runs then?
To provide an aggregated result? 

Also, I'm not sure how reporters will work with this change....
Probably this may break them as instead of one report with overridden data they will create multiple reports which probably is not the outcome users expects.

So I'm not yet convinced here. 

@dwentland24 what do you think on it?

jain-neeeraj:
@DavertMik I am sharing my understanding, I could be completely wrong as well and in that case please feel free to correct me with the problem in my understanding.

If we create a plugin listening to `event.all.after` event like below
```
  event.dispatcher.on(event.all.after, async () => {
    console.log('After got called');
  });
```
And run tests using `run-rerun` command then it does not print anything on the console and it means `event.after.all` event was not issued in time to be picked up by the plugin. But if we use the approach outlined in the PR then the text `After got called` gets printed `minSuccess` or `maxReruns` times, whichever occurs first.

One more reason to assume that the current approach fails is the below message which appears on every `rerun`  of tests that means codecept / mocha object gets destroyed with each test run and even if we somehow manage to run `event.all.after` then it might be capturing data only from last `rerun` attempt. 

image

As I said, I could be completely wrong as well but this problem was the driving factor of this PR.

DavertMik:
@jain-neeeraj I am also not sure of the best approach here, as rerun command was proposed and maintained by other contributors and I think they have their own use cases.

Still my guess is to use `all.before` and `all.after` events only once.

If they are used to submit report, start/stop server, or even exit process, they should happen once.

jain-neeeraj:
Your point of view also seems valid but in the current implementation `all.after` was not getting captured even once in the plugin event listener.

dwentland24:
Hi, 
I also would expect the `all.before` and `all.after` once after all runs are finished. 
I had a closer look to the run rerun methods. The reason why the `all.results` and `all.after` hook not called, is that the `done()` function never gets executed. I think the old teardown method accepted a callback function, which is not the case in the current version. 

As I am currently sick I will just put the code I changed this morning in this comment. Code is not tested. 
@jain-neeeraj Maybe you can give it a try. This should trigger the events again once, after all runs are done.

``` javascript

async run(test) {
    event.emit(event.all.before, this);
    try {
      await this.runTests(test);
    } catch (e) {
      output.error(e.stack);
    } finally {
      event.emit(event.all.result, this);
      event.emit(event.all.after, this);
    }
  }


==============================

async runOnce(test) {
    container.createMocha();
    const mocha = container.mocha();
    this.testFiles.forEach((file) => {
      delete require.cache[file];
    });
    mocha.files = this.testFiles;
    if (test) {
      if (!fsPath.isAbsolute(test)) {
        test = fsPath.join(global.codecept_dir, test);
      }
      mocha.files = mocha.files.filter(t => fsPath.basename(t, '.js') === test || t === test);
    }
    mocha.run();
  }


```

jain-neeeraj:
@dwentland24 this code is giving even more weird results but I got the idea of what you are trying to do. I will try to make it work on those lines

jain-neeeraj:
@dwentland24 made changes as per your suggestions. During my testing, I could not find any problem with this code. Since I am new to the community, please see if it covers all use cases

DavertMik:
Thank you @jain-neeeraj and @dwentland24 to taking part in it
I think this PR looks good and simplifies legacy code we had from previous release
This change seems to be non-breaking so I'm ok to merge it

@jain-neeeraj welcome to the community! :wave:

PR #3465, opened 2022-10-20
Promise based helper methods

nlespiaucq:
- Provide TypeScript definitions for all helper methods with promise,
- Add new configuration `fullPromiseBased` to refer the promise-based definition.
- Fix runner-related tests and add typing tests

Applicable helpers:

- [X] WebDriver
- [X] Puppeteer
- [X] Nightmare
- [X] REST
- [X] FileHelper
- [X] Appium
- [X] Protractor
- [X] TestCafe
- [X] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] stepTimeout
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [ ] :bug: Bug fix
- [X] :clipboard: Documentation changes/updates
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [X] Tests have been added

### DIFF
--- a/runok.js
+++ b/runok.js
@@ -33,6 +33,8 @@ module.exports = {
   async defTypings() {
     console.log('Generate TypeScript definition');
     await npx('jsdoc -c typings/jsdoc.conf.js');
+    await npx('jsdoc -c typings/jsdocPromiseBased.conf.js');
+    fs.renameSync('types.d.ts', 'typings/promiseBasedTypes.d.ts');
   },
 
   async docsPlugins() {
--- a/runok.js
+++ b/runok.js
@@ -32,9 +32,11 @@ module.exports = {
 
   async defTypings() {
     console.log('Generate TypeScript definition');
-    await npx('jsdoc -c typings/jsdoc.conf.js');
+    // Generate definitions for promised-based helper methods
     await npx('jsdoc -c typings/jsdocPromiseBased.conf.js');
-    fs.renameSync('types.d.ts', 'typings/promiseBasedTypes.d.ts');
+    fs.renameSync('typings/types.d.ts', 'typings/promiseBasedTypes.d.ts');
+    // Generate all other regular definitions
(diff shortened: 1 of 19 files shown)


runok.js
```diff
@@ -33,6 +33,8 @@ module.exports = {
   async defTypings() {
     console.log('Generate TypeScript definition');
     await npx('jsdoc -c typings/jsdoc.conf.js');
+    await npx('jsdoc -c typings/jsdocPromiseBased.conf.js');
+    fs.renameSync('types.d.ts', 'typings/promiseBasedTypes.d.ts');
```
DavertMik: Maybe we should execute here:

```
await npx('jsdoc -c typings/jsdoc.conf.js');
```

so types.d.ts to be actually generated for classical setup?
because according to the code, you generate types twice, overwrites them on second run and move them, but the result of `'jsdoc -c typings/jsdoc.conf.js'` is overwritten

DavertMik (commented):
@nlespiaucq awesome! Just one question here:

runok.js
nlespiaucq: Yes, you're right - I will fix it

nlespiaucq: Well, it's confusing, but it was done on purpose.
In fact, the 2nd command generates the file types.d.ts on another location, on the root of the folder (to avoid to overwritten the first one). It's why I rename and move this file.

I will have some comments and make it more consistent.

DavertMik (approved):
This was huge!
Thanks, @nlespiaucq for making it happen!

PR #3479, opened 2022-10-27
fix(docs): missing hyperlink

kobenguyent:
Screen Shot 2022-10-27 at 2 44 07 PM

## Type of change

### DIFF
--- a/runok.js
+++ b/runok.js
@@ -218,6 +218,7 @@ Our community prepared some valuable recipes for setting up CI systems with Code
         if (!text) return;
         cfg.replace('<!-- configuration -->', text[1]);
         cfg.replace(/## config((.|\n)*)\[1\]/m, '');
+        cfg.replace(/^: /m, '[1]:');
       });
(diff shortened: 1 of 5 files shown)


DavertMik:
I don't understand how these empty links actually occurred
I would prefer to try to find the cause instead of adding a patch

runok.js
```diff
@@ -218,6 +218,7 @@ Our community prepared some valuable recipes for setting up CI systems with Code
         if (!text) return;
         cfg.replace('<!-- configuration -->', text[1]);
         cfg.replace(/## config((.|\n)*)\[1\]/m, '');
```
kobenguyent: This regex was the issue @DavertMik it is somehow replacing from ## config to [1]. That's why there is only : left and causes the issue.

Arhell (approved):

DavertMik:
Thank you for investigating and improving runok! 
Looks good :+1:

PR #3514, opened 2022-12-02
fix: undefined global codeceptjs dir

kobenguyent:

Applicable helpers:

- [ ] WebDriver
- [ ] Puppeteer
- [ ] Nightmare
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] Protractor
- [ ] TestCafe
- [ ] Playwright

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] stepTimeout
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [ ] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/container.js
+++ b/lib/container.js
@@ -9,6 +9,8 @@ const event = require('./event');
 const WorkerStorage = require('./workerStorage');
 const store = require('./store');
 
+global.codecept_dir = global.codecept_dir || process.cwd();
+
 let container = {
   helpers: {},
   support: {},
--- a/lib/container.js
+++ b/lib/container.js
@@ -182,8 +182,7 @@ function createHelpers(config) {
           throw new Error(`Required modules are not installed.\n\nRUN: ${install}`);
         }
       }
-      let helpers = { helperName };
-      
+
       if (typeof HelperClass === 'object') {
         Object.values(HelperClass).forEach(helper => {
           helpers[helperName] = new helper(config[helperName])
@@ -200,6 +199,7 @@ function createHelpers(config) {
   for (const name in helpers) {
     if (helpers[name]._init) helpers[name]._init();
   }
+  
   return helpers;
 }
 
--- a/lib/container.js
+++ b/lib/container.js
@@ -8,8 +8,9 @@ const recorder = require('./recorder');
 const event = require('./event');
 const WorkerStorage = require('./workerStorage');
 const store = require('./store');
-
-global.codecept_dir = global.codecept_dir || process.cwd();
+const { getTestRoot } = require('../lib/command/utils');
+global.codecept_dir = global.codecept_dir || getTestRoot();
+global.Helper = global.Helper || require('../lib/helper');
 
 let container = {
   helpers: {},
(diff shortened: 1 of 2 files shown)


Arhell (approved):

kobenguyent:
````
----------------------
? Feature which is being tested (ex: account, login, etc) with custom helper
? Filename of a test with_custom helper_test.ts
[class Playwright extends Helper]
{ CustomerHelper: [class CustomerHelper extends Helper] }

Test for with_custom helper_test.ts was created in C:\Users\nguyenth\Desktop\helloworld\with_custom helper_test.ts
PS C:\Users\nguyenth\Desktop\helloworld> npx codeceptjs gt
Creating a new test...
----------------------
? Feature which is being tested (ex: account, login, etc) no custom helper
? Filename of a test no_custom helper_test.ts
[class Playwright extends Helper]

Test for no_custom helper_test.ts was created in C:\Users\nguyenth\Desktop\helloworld\no_custom helper_test.ts
````

Custom Helper:
````
export class CustomerHelper extends Helper {

}
````

File: lib/container.js [hunk 1]
```diff
@@ -8,6 +8,9 @@ const recorder = require('./recorder');
 const event = require('./event');
 const WorkerStorage = require('./workerStorage');
 const store = require('./store');
+const { getTestRoot } = require('../lib/command/utils');
```
DavertMik, line 11: I'm against including this into a container
container should not know anything about filesystem

DavertMik (changes requested):
Sorry, I don't like this fix
also I didn't understand when the bug happens

kobenguyent:
Hey @DavertMik i guess the issue is from here https://github.com/codeceptjs/CodeceptJS/issues/3512 but not enough knowledge where to put fix. So leave it to you then.

lib/container.js
mirao: Just solved it to myself by putting of `global.codecept_dir = testsPath;` after https://github.com/codeceptjs/CodeceptJS/blob/ad5f4afab6e1f5e12065398a7a4e7e94fead33fb/lib/command/generate.js#L26

But still there is an issue that all custom helpers need to have included `const Helper = require('@codeceptjs/helper');`, otherwise you get the issue https://github.com/codeceptjs/CodeceptJS/issues/3512#issuecomment-1335554742

mirao: > But still there is an issue that all custom helpers need to have included const Helper = require('@codeceptjs/helper');, otherwise you get the issue https://github.com/codeceptjs/CodeceptJS/issues/3512#issuecomment-1335554742

On the other hand

- Newly created Helpers (by `codeceptjs gh`) have the variable `Helper` set
- I can fix my existing custom helpers easily
- As for other custom helpers that I use, only `codeceptjs-chai` doesn't have `Helper` set and I must patch it. Others (`codeceptjs-cmdhelper`, `codeceptjs-dbhelper` and `codeceptjs-pixelmatchhelper`) are already set correctly.

So it's not a big issue.

zhigulin-n: @mirao, how did you patch `codeceptjs-chai` helper? I have TS config and this solve doesn't work for me... :( I talk about your #3611

mirao: @zhigulin-n 

`codeceptjs-chai@2.3.2.patch`:
```patch
diff --git a/index.js b/index.js
index bd248fce8c1581fe07a63abae3a469159e7891ca..7cd6dcca0e8ba1e47195e8053f07187ca08782ca 100644
--- a/index.js
+++ b/index.js
@@ -1,3 +1,5 @@
+const Helper = require('@codeceptjs/helper');
+
 /* eslint-disable global-require */
 const chai = require( 'chai' );
 
```

kobenguyent: https://github.com/SitamJana/codeceptjs-chai/pull/39

Perhaps this patch would help in general

mirao: @kobenguyent Thanks. Your PR really helps to solve the issue and therefore my patch is not needed anymore.
I tested that then `codeceptjs run` and `codeceptjs gt` work well in a project using `codeceptjs-chai`.

PR #3539, opened 2022-12-16
feature(allure): Add framework and language labels to tests

vovsemenv:

This PR add `language` and `framework` fields for allure plugin to easily filter testResults by language or frameworks

Applicable plugins:

- [x] allure

## Type of change

- [x] :nail_care: Polish code

## Checklist:

- [x] Tests have been added

### DIFF
--- a/lib/plugin/allure.js
+++ b/lib/plugin/allure.js
@@ -99,6 +99,9 @@ module.exports = (config) => {
   reporter.pendingCase = function (testName, timestamp, opts = {}) {
     reporter.startCase(testName, timestamp);
 
+    plugin.addLabel('language', 'javascript');
+    plugin.addLabel('framework', 'codeceptjs');
+
     if (opts.description) plugin.setDescription(opts.description);
     if (opts.severity) plugin.severity(opts.severity);
     if (opts.severity) plugin.addLabel('tag', opts.severity);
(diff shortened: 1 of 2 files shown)


kobenguyent:
hey @vovsemenv I guess we could expose these as public API to let the one who wants to add them as their choice. https://codecept.io/plugins/#public-api

vovsemenv:
In my opinion it's better to add option to `opt-out` from the codecept metadata labels because by itself this labels didn't  break anything. Sounds fine if i add option to disable it?

kobenguyent:
IMHO, just expose these as public APIs like we currently have https://github.com/codeceptjs/CodeceptJS/blob/408e4d3dc21288722cc3db194b807b2a23164077/lib/plugin/allure.js#L156 for example. Who wants to use them just uses as their choice.

vovsemenv:
But in this case user needs to add

```
allure.language('javascript');
allure.language('codeceptjs');
```

in every test file 😢  but this fields actually the same across all tests with codeceptjs framework

kobenguyent:
> But in this case user needs to add
> 
> ```
> allure.language('javascript');
> allure.language('codeceptjs');
> ```
> 
> in every test file 😢 but this fields actually the same across all tests with codeceptjs framework

I think that is feasible by implementing the custom plugin to subscribe to a certain event to call those APIs when they need.
https://codecept.io/hooks/#api

DavertMik (approved):
Thanks

vovsemenv:
merge?)

DavertMik:
Yes!

PR #3562, opened 2023-01-05
feat(run-workers) : Add support of multiple for run-workers

karanshah-browserstack:
Add Support of using multiple with run-workers to run multiple browsers out of the box with workers instead of triggering it by manually creating a custom runner. 

## Type of change

## Checklist:

- [x] Tests have been added

### DIFF
--- a/bin/codecept.js
+++ b/bin/codecept.js
@@ -130,7 +130,7 @@ program.command('run [test]')
   .option('--child <string>', 'option for child processes')
   .action(errorHandler(require('../lib/command/run')));
 
-program.command('run-workers <workers>')
+program.command('run-workers <workers> [selectedRuns...]')
   .description('Executes tests in workers')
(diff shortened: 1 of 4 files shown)


karanshah-browserstack:
@PeterNgTr @Arhell @DavertMik Can someone review this?

karanshah-browserstack:
@PeterNgTr @Arhell @DavertMik bumping for review.

DavertMik:
I'm sorry, I prefer not to break current functionality and API of run-workers as it was stable for quite a while
Also this feature is not documented so I didn't get an idea of how it should be used
So I think custom runner is still a preferred choice

karanshah-browserstack:
@DavertMik Will add documentation for this, should that be sufficient to reopen this? High level, In this PR we are just adding support for running multiple browsers with run-workers which is currently not supported.

PR #3578, opened 2023-02-07
fix: cannot load wdio helper

kobenguyent:
- fix: cannot load wdio helper

Applicable helpers:

- [x] WebDriver

## Type of change
- [x] :bug: Bug fix

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/helper/WebDriver.js
+++ b/lib/helper/WebDriver.js
@@ -2081,20 +2081,7 @@ class WebDriver extends Helper {
     const client = this.browser;
     const aSec = sec || this.options.waitForTimeoutInSeconds;
     let currUrl = '';
-    if (isWebDriver5()) {
-      return client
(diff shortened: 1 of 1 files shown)


Arhell (approved):

DavertMik (approved):

DavertMik (changes requested):
I think we use `version ` only to check if it is webdriverio v5
If the actual version is 8 I don't think we have to support v5 in any way

So I would remove all the code assumed of v5 usage and version detection as well

kobenguyent:
> I think we use `version ` only to check if it is webdriverio v5 If the actual version is 8 I don't think we have to support v5 in any way
> 
> So I would remove all the code assumed of v5 usage and version detection as well

updated.

DavertMik (approved):

PR #3622, opened 2023-03-21
feat: appium v2 support

kobenguyent:
- appium v2 support
- migration guide: https://github.com/codeceptjs/CodeceptJS/wiki/Migration-to-Appium-v2---CodeceptJS

Applicable helpers:
- [x] Appium

## Type of change
- [x] :fire: Breaking changes

## Checklist:
- [ ] Tests have been added


Arhell (commented):
@DavertMik

ChetanChohan7:
@kobenguyent I was taking a look at this as it's of interest to my current work. I noticed in the appium helper file there are several instances where "browser" is misspelt as, "broswer". See lines 192 and 197 in /helper/appium.js

xuanmai-agilityio:
@kobenguyent I also found another issue with the `_startBrowser` function, when Appium runs the 2nd test case, it will run this function and will convert the existing options to add one more `appium:` prefix in the config.
So I think we need to check and only convert it one time when the Appium starts.
So I've updated it as below and it runs smoothly. Hope this help!

Screen Shot 2023-05-09 at 10 49 59

kobenguyent:
> @kobenguyent I also found another issue with the `_startBrowser` function, when Appium runs the 2nd test case, it will run this function and will convert the existing options to add one more `appium:` prefix in the config. So I think we need to check and only convert it one time when the Appium starts. So I've updated it as below and it runs smoothly. Hope this help!
> 
> Screen Shot 2023-05-09 at 10 49 59

Thanks @xuanmai-agilityio, issue has been addressed!

DavertMik:
Thank you, @kobenguyent
I think we are ready to merge this one.
Sorry for the delay in reviewing it

DavertMik (approved):

PR #3665, opened 2023-05-15
feat(playwright)/new api: `.clear()`, `.blur()`, `.focus()`

EgorBodnar:
### The reason for the PR is the Integration of the new Playwright API into CodeceptJS and the bumping Playwright version to `1.32.3`.

 **A list of the new Playwright API, which is important to the CodeceptJS framework, but has not yet been included**:
* Added in [v1.28 of Playwright](https://playwright.dev/docs/release-notes#version-128)
``` js
.blur()
.clear()
```
* Added in [v1.14 of Playwright](https://playwright.dev/docs/release-notes#version-114)
``` js
.focus()
```

 ### As part of this PR added implementations of these methods in Codeceptjs.
- Bringing the name of options to a general form: [opts vs options](https://github.com/EgorBodnar/CodeceptJS/commit/698701c71679c92a8240b4e6d79bea791a0ad485)

 - Added tests, HTML forms for tests and documentation:
    * `I.focus()`
    * `I.blur()`
    * `I.clearField()`modified  | old `I.clearField()` In fact, didn't clean the field, but replaced its value with an empty string. It doesn't always work. So it's better to use the new Playwright API (`.clear`) for this.  Backward compatibility is preserved for those who use the old version of the Playwright. 

- Created feature request #3677 because all the new Playwright functionality is tied to the [Locator](https://playwright.dev/docs/api/class-locator) and it is impossible to use the new Playwright API for elements found through the [ElementHandle](https://playwright.dev/docs/api/class-elementhandle#element-handle-query-selector-all) only if it does not match with `devtools-protocol` API.

- Added temporary solution function [getXPathForElement(elementHandle)](https://github.com/codeceptjs/CodeceptJS/pull/3665/commits/c4efd65f6dfb281d30650ee1b27b332f7f22c85b) to use like a bridge between old [ElementHandle](https://playwright.dev/docs/api/class-elementhandle#element-handle-query-selector-all) and new [Locator](https://playwright.dev/docs/api/class-locator)

Applicable helpers:

- [X] Playwright

## Type of change

- [X] :nail_care: Polish code

## Checklist:

- [X] Tests have been added

### DIFF
--- a/lib/helper/Playwright.js
+++ b/lib/helper/Playwright.js
@@ -906,8 +906,8 @@ class Playwright extends Helper {
   }
 
   /**
-   * {{> focus }}
    * Calls [focus](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus) on the matching element.
+   * @param {CodeceptJS.LocatorOrString} locator field located by label|name|CSS|XPath|strict locator.
    * @param {any} [options] [Additional options](https://playwright.dev/docs/api/class-locator#locator-focus) for available options object as 2nd argument.
    *
    * Examples:
@@ -923,6 +923,7 @@ class Playwright extends Helper {
     const els = await this._locate(locator);
     assertElementExists(els, locator, 'Element to focus');
     const el = els[0];
+
     await el.focus(options);
     return this._waitForAction();
   }
--- a/lib/helper/Playwright.js
+++ b/lib/helper/Playwright.js
@@ -929,9 +929,9 @@ class Playwright extends Helper {
   }
 
   /**
-   * {{> blur }}
    * Remove focus from a text input, button, etc
    * Calls [blur](https://playwright.dev/docs/api/class-locator#locator-blur) on the element.
+   * @param {CodeceptJS.LocatorOrString} locator field located by label|name|CSS|XPath|strict locator.
    * @param {any} [options] [Additional options](https://playwright.dev/docs/api/class-locator#locator-blur) for available options object as 2nd argument.
    *
    * Examples:
@@ -950,8 +950,9 @@ class Playwright extends Helper {
   async blur(locator, options = {}) {
     const els = await this._locate(locator);
     assertElementExists(els, locator, 'Element to blur');
-    const el = els[0];
-    await el.blur(options);
+
+    // TODO: locator change required after #3677 implementation
+    await this.page.locator(buildLocatorString(new Locator(locator))).blur(options);
     return this._waitForAction();
   }
 
--- a/lib/helper/Playwright.js
+++ b/lib/helper/Playwright.js
@@ -1530,9 +1530,34 @@ class Playwright extends Helper {
 
   /**
    * {{> clearField }}
+   * (deprecated) use `I.clear() instead`
    */
   async clearField(field) {
+    console.log('I.clearField() is DEPRECATED: use `I.clear()` to clear field or text area');
     return this.fillField(field, '');
+  }
+
+  /**
+   * {{> clear }}
+   * Clear the <input>, <textarea> or [contenteditable] .
+   * @param {any} [options] [Additional options](https://playwright.dev/docs/api/class-locator#locator-clear) for available options object as 2nd argument.
+   *
+   * Examples:
+   *
+   * ```js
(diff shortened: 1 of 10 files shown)


Arhell (approved):

kobenguyent (approved):

test/helper/Playwright_test.js
```diff
@@ -184,6 +184,33 @@ describe('Playwright', function () {
...
+      .then(() => I.blur('#button'))
+      .then(() => I.see('Button not focused', '#buttonMessage'))
+      .then(() => I.focus('#field'))
+      .then(() => I.see('Input field is focused', '#fieldMessage'))
+      .then(() => I.blur('#field'))
+      .then(() => I.see('Input field not focused', '#fieldMessage'))
+      .then(() => I.focus('#textarea'))
+      .then(() => I.see('Textarea is focused', '#textareaMessage'))
+      .then(() => I.focus('#textarea'))
+      .then(() => I.see('Textarea not focused', '#textareaMessage')));
```
kobenguyent: Just curious if we could use await/async here?

EgorBodnar: Agree with you. Fixed

lib/helper/Playwright.js
```diff
@@ -902,6 +902,55 @@ class Playwright extends Helper {
...
+   * I.see('#add-to-cart-btn');
+   * I.blur('#product-tile')
+   * I.dontSee('#add-to-cart-btn');
+   * ```
+   *
+   */
+  async blur(locator, options = {}) {
+    const els = await this._locate(locator);
+    assertElementExists(els, locator, 'Element to blur');
+    const el = els[0];
```
EgorBodnar: @DavertMik what is the proper way to locate the element here? 
use `._locate()` or `.findClickable()`? 
Because `.findClickable()` in fact, not just looking for an available element to click, but just well and steadily locating an element

lib/helper/Playwright.js
```diff
@@ -1468,11 +1517,36 @@ class Playwright extends Helper {
...
+   *
+   * ```js
+   * I.clear('.text-area')
+   * ```
+   * ```js
+   * I.clear('#submit', { force: true }) // force to bypass the [actionability](https://playwright.dev/docs/actionability) checks.
+   * ```
+   *
+   */
+  async clear(locator, options = {}) {
```
peterngtr: What do you think if we name it clearFieldValue()

EgorBodnar: In a good way, we need to completely replace the `.clearField()`. 
But the`.clearField()` has already been used many times by users. 

#### Here are the possible options, as I see it:
1.  we need to either replace it slowly in the next Codeceptjs **major** version,
2. leave it as it is and remove deprecated in the next versions
3. completely replace the `.clearField()` here

So we need a final decision from @DavertMik

kobenguyent: I'm up with replacing the implementation of clearField() instead of introducing different names. As far as I know the current clearField still brings some issues.

EgorBodnar: I've the same opinion as you.
As this is a major change, need approval from @DavertMik

DavertMik: > I'm up with replacing the implementation of clearField() instead of introducing different names. As far as I know the current clearField still brings some issues.

@EgorBodnar yes, I agree to replace implementation

It's not a major change as the behavior should be the same as current clearField
Sure, implementation is different and may bring some edge cases issue but I agree, no need to introduce new method

lib/helper/Playwright.js
DavertMik: I think `_locate` is better here

DavertMik (changes requested):
Here is the thing. Playwright introduced `clear` only in recent releases, while we had clearField (not with the best implementation) from the very beginning :smiley: 

But now please replace `clearField` implementation with a native one and I'm merging

lib/helper/Playwright.js
EgorBodnar: replaced

DavertMik (approved):

DavertMik:
Thank you so much!

PR #3691, opened 2023-06-07
fix: handle importing custom helper nicely

kobenguyent:
- This PR aims to handle the importing helpers nicely.

````
// new export syntax

class CustomHelper {}

export default CustomHelper;


// old export syntax

class CustomHelper {}

export = CustomHelper;

// js class
class TestHelper extends Helper {

}

module.exports = TestHelper;

````

## Type of change
- [x] :bug: Bug fix

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/container.js
+++ b/lib/container.js
@@ -165,7 +165,11 @@ function createHelpers(config) {
       } else {
         moduleName = `./helper/${helperName}`; // built-in helper
       }
-      const HelperClass = require(moduleName);
+
+      // @ts-ignore
+      // this handles the loading of custom helper. Either old syntax `export = HelperName` or new syntax `export default HelperName`
+      const HelperClass = moduleName.startsWith('./helper/') ? require(moduleName) : require(moduleName).default || require(moduleName);
+
       if (HelperClass._checkRequirements) {
         const requirements = HelperClass._checkRequirements();
         if (requirements) {
@@ -351,8 +355,8 @@ function loadSupportObject(modulePath, supportObjectName) {
       }
     }
     if (typeof obj !== 'function'
-        && Object.getPrototypeOf(obj) !== Object.prototype
-        && !Array.isArray(obj)
+      && Object.getPrototypeOf(obj) !== Object.prototype
+      && !Array.isArray(obj)
     ) {
       const methods = getObjectMethods(obj);
       Object.keys(methods)
--- a/lib/container.js
+++ b/lib/container.js
@@ -167,8 +167,13 @@ function createHelpers(config) {
       }
 
       // @ts-ignore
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lib/container.js [hunk 1]
```diff
@@ -165,7 +165,11 @@ function createHelpers(config) {
       } else {
         moduleName = `./helper/${helperName}`; // built-in helper
       }
-      const HelperClass = require(moduleName);
+
+      // @ts-ignore
```
EgorBodnar, line 169: Could we avoid of the `// @ts-ignore` by something like this: 

```js
const getModule = (moduleName) => {
  const module = require(moduleName);
  return module.default ? module : { default: module };
};

const isHelperModule = moduleName.startsWith('./helper/');

const HelperClass = isHelperModule
  ? require(moduleName)
  : getModule(moduleName).default;

```

?

kobenguyent: Screenshot 2023-06-07 at 16 14 41
it doesn't resolves the issue unfortunately.

EgorBodnar: Oh... I see. Hard situation.

A good time to think about switching to TypeScript and solve the typing issue the correct way :slightly_smiling_face:

EgorBodnar (approved):

DavertMik:
Sorry, I can't see purpose of this

`export default` looks nicer, indeed. But as we still use it through require() I don't see any point of it
It might confuse users as they may think that they can use `import` inside CodeceptJS which is not true

Moving away from CommonJS would be a nice idea but I see that it requires a lot of time, especially as we have dynamic imports everywhere.

kobenguyent:
Hey @DavertMik, I wanted to discuss a change regarding how we handle the loading of custom helpers. Nowadays, most users are using TypeScript for their custom helpers, and the "export =" syntax, which is deprecated, is being replaced with "export default" to export classes. However, this change causes errors when users try to do the same with CodeceptJS. This pull request aims to address this issue by ensuring that any syntax used to export classes in custom helpers will not cause any problems. By doing so, we won't limit users in their custom helper choices. Of course, it's important to note that their custom helpers will remain their responsibility, but our core functionality will continue to work flawlessly. This change could be a way to attract more users.

lib/container.js
```diff
@@ -165,7 +165,11 @@ function createHelpers(config) {
...
+      // this handles the loading of custom helper. Either old syntax `export = HelperName` or new syntax `export default HelperName`
+      const HelperClass = moduleName.startsWith('./helper/') ? require(moduleName) : require(moduleName).default || require(moduleName);
```
DavertMik: Could we split this into multiple operators as this becomes unreadable and it's hard to understand which operator is used in which context

DavertMik (changes requested):
OK, I'm convinced to accept this
But please do a minor change to improve the code readability

PR #3692, opened 2023-06-08
fix(pw): drag n drop issue

kobenguyent:
- Drag and drop doesn't work with Angular and React webapp

Applicable helpers:
- [ ] Playwright

## Type of change
- [ ] :bug: Bug fix

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/helper/Playwright.js
+++ b/lib/helper/Playwright.js
@@ -960,20 +960,27 @@ class Playwright extends Helper {
   /**
    * {{> dragAndDrop }}
    *
-   * @param {any} [options] [Additional options](https://playwright.dev/docs/api/class-page#page-drag-and-drop) can be passed as 3rd argument.
-   *
    * ```js
    * // specify coordinates for source position
    * I.dragAndDrop('img.src', 'img.dst', { sourcePosition: {x: 10, y: 10} })
    * ```
    *
-   * > By default option `force: true` is set
    */
-  async dragAndDrop(srcElement, destElement, options = { force: true }) {
-    const src = new Locator(srcElement, 'css');
-    const dst = new Locator(destElement, 'css');
+  async dragAndDrop(srcElement, destElement) {
+    const _smallWaitInMs = 600;
+    const src = new Locator(srcElement);
+    const dst = new Locator(destElement);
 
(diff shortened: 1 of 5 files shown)


EgorBodnar:
It's a Playwright issue in the feature implementation: https://github.com/microsoft/playwright/issues/1094

We need to study the problem in more detail. Please don't merge it yet :pray:

lib/helper/Playwright.js
```diff
@@ -960,20 +960,27 @@ class Playwright extends Helper {
...
    * // specify coordinates for source position
    * I.dragAndDrop('img.src', 'img.dst', { sourcePosition: {x: 10, y: 10} })
    * ```
    *
-   * > By default option `force: true` is set
    */
-  async dragAndDrop(srcElement, destElement, options = { force: true }) {
-    const src = new Locator(srcElement, 'css');
-    const dst = new Locator(destElement, 'css');
+  async dragAndDrop(srcElement, destElement) {
```
mirao: If I understand well, it would break my use case because you removed the options parameter:
```ts
    // Move the object from default to initial position
    I.dragAndDrop(konvajsFragment.gameArea, konvajsFragment.gameArea, {
        sourcePosition: OBJECT_DEFAULT_POSITION,
        targetPosition: OBJECT_INITIAL_POSITION,
    });
```

kobenguyent: with the new implementation, I don't feel like we need the options parameter.

mirao: @kobenguyent 
My use case is "Move an object from one position to another one inside an area (canvas)". Therefore I need to pass source and target coordinates to `I.dragAndDrop()`.

kobenguyent: hey @mirao @EgorBodnar fixing for backward compatibility.

EgorBodnar: > hey @mirao @EgorBodnar fixing for backward compatibility.

great! Looks good

An additional flag would be more intuitive for users, but your solution is more elegant.

EgorBodnar (approved):

Arhell (approved):

PR #3697, opened 2023-06-09
fix: file name is not unique enough when running tests in parallel

kobenguyent:
- The fileName is not unique when running tests in parallel and it leads to the issue that attachment couldn't attach to allure report.

Applicable helpers:
- [ ] Playwright

## Type of change
- [ ] :bug: Bug fix

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/helper/Playwright.js
+++ b/lib/helper/Playwright.js
@@ -3111,7 +3111,7 @@ async function refreshContextSession() {
 
 async function saveVideoForPage(page, name) {
   if (!page.video()) return null;
-  const fileName = `${`${global.output_dir}${pathSeparator}videos${pathSeparator}${Date.now()}_${clearString(name)}`.slice(0, 245)}.webm`;
+  const fileName = `${`${global.output_dir}${pathSeparator}videos${pathSeparator}${makeId(20)}_${clearString(name)}`.slice(0, 245)}.webm`;
   page.video().saveAs(fileName).then(() => {
     if (!page) return;
     page.video().delete().catch(e => {});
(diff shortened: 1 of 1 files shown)


EgorBodnar:
a good fix, as a result of our discussion :raised_hands:

The custom attachment `fileName` should be unique for reporters or it would be rewritten

lib/helper/Playwright.js
```diff
@@ -3122,7 +3122,19 @@ async function saveVideoForPage(page, name) {
 async function saveTraceForContext(context, name) {
   if (!context) return;
   if (!context.tracing) return;
-  const fileName = `${`${global.output_dir}${pathSeparator}trace${pathSeparator}${Date.now()}_${clearString(name)}`.slice(0, 245)}.zip`;
+  const fileName = `${`${global.output_dir}${pathSeparator}trace${pathSeparator}${makeId(20)}_${clearString(name)}`.slice(0, 245)}.zip`;
```
EgorBodnar: Should the `fileName` be unique for one test or it probably should be unique within one test step too?

* If it should be unique per test, use `mocha current test id` or `nodejs worker.threadId`

* If it should be unique within one test use already installed CodeceptJS dependency - `uuid` to generate really unique id. 

So better to delete the new `function makeId(length) {}`

EgorBodnar (changes requested):
Should the `fileName` be unique for one test or it probably should be unique within one test step too?

* If it should be unique per test, use `mocha current test id` or `nodejs worker.threadId`

* If it should be unique within one test use already installed CodeceptJS dependency - `uuid` to generate really unique id. 

So better to delete the new `function makeId(length) {}`

kobenguyent:
> Should the `fileName` be unique for one test or it probably should be unique within one test step too?
> 
> * If it should be unique per test, use `mocha current test id` or `nodejs worker.threadId`
> * If it should be unique within one test use already installed CodeceptJS dependency - `uuid` to generate really unique id.
> 
> So better to delete the new `function makeId(length) {}`

updated per CR. Thanks for pointing it out!

EgorBodnar (approved):

Arhell (approved):

PR #3713, opened 2023-07-01
AI features for CodeceptJS

DavertMik:
- [x] Integrated OpenAI library 
- [x] added heal plugin to automatically fix broken steps in test
- [x] added support for OpenAI requests inside interactive pause
- [x] added general OpenAI helper for whatever


lib/helper/OpenAI.js
```diff
@@ -0,0 +1,122 @@
...
+  async askGptGeneralPrompt(prompt) {
+    const messages = [
+      { role: 'user', content: prompt },
+    ];
+
+    const completion = await this.aiAssistant.createCompletion(messages);
+
+    const response = completion?.data?.choices[0]?.message?.content;
+
+    console.log(response);
```
kobenguyent, line 118: Shall we remove this log?

lib/helper/OpenAI.js
```diff
@@ -0,0 +1,122 @@
...
+    const html = await this.helper.grabHTMLFrom(locator);
+
+    const messages = [
+      { role: 'user', content: prompt },
+      { role: 'user', content: `Within this HTML: ${minifyHtml(html)}` },
+    ];
+
+    const response = await this.aiAssistant.createCompletion(messages);
+
+    console.log(response);
```
kobenguyent, line 99: Shall we remove this?

lib/helper/OpenAI.js
DavertMik: Not really, we would like to see the response from OpenAI here
This is why I explicitly print as console.log... But ,maybe I should use output print, that also makes a deal

PR #3722, opened 2023-07-02
Highlight fixes

DavertMik:
- auto start highlight in debug mode
- replaced border with box shadow as it doesn't change the layout
- refactored code to simplify


kobenguyent (approved):

lib/helper/scripts/highlightElement.js
```diff
@@ -1,15 +1,12 @@
...
   try {
-    context.evaluate(el => el.style.border = '2px solid red', element);
+    context.evaluate(el => {
+      const style = '0px 0px 4px 3px rgba(255, 0, 0, 0.7)';
+      const prevStyle = el.style.boxShadow;
+      el.style.boxShadow = style;
+      setTimeout(() => el.style.boxShadow = prevStyle, 2000);
+    }, element);
   } catch (e) {
-    context.execute(el => el.style.border = '2px solid red', element);
```
kobenguyent, line 5: @DavertMik not sure if the eval works for webdriver that's why I used execute here

DavertMik: thanks, got it

PR #3730, opened 2023-07-04
Fix backtick inside Playwright.md

Horsty80:
That lead to truncate doc on website

image

- Playwright helper docs page was truncate due to wrong number of backtick

## Type of change

- [x] :hammer: Markdown files fix - not related to source code


kobenguyent (approved):

DavertMik:
Thanks for quick fix, but actual problem was in tags `<input>`, `<textarea>` which broke HTML

Horsty80:
> Thanks for quick fix, but actual problem was in tags `<input>`, `<textarea>` which broke HTML

New PR with good fix

PR #3738, opened 2023-07-06
feat: move to Locator playwright

### LINKED ISSUE #3677: [FEATURE REQUEST](playwright) migrate locators from `ElementHandle` to `Locator`.

In the CodeceptJS `Playwright` helper all locators are tied to [ElementHandle](https://playwright.dev/docs/api/class-elementhandle#element-handle-query-selector-all), but it is officially discouraged/deprecated. 

Need to move to the [Locator](https://playwright.dev/docs/api/class-locator)

The whole `Playwright` helper should be changed due to migration to the [Locator](https://playwright.dev/docs/api/class-locator)

kobenguyent:
- Resolves #3677 

Applicable helpers:
- [x] Playwright

## Type of change
- [x] :bug: Bug fix

## Checklist:


lib/helper/Playwright.js
```diff
@@ -1556,21 +1551,16 @@ class Playwright extends Helper {
...
-      const elXpath = await getXPathForElement(els[0]);
+    highlightActiveElement.call(this, el, this.page);
 
-      await this.page.locator(elXpath).clear(options);
-      result = await this._waitForAction();
-    } else {
-      result = await this.fillField(locator, '');
-    }
-    return result;
+    await el.fill('');
```
DavertMik: why we don't use `clear` in this case?
`fill('')` could produce different result

DavertMik (changes requested):
Good to me
But can we use new `clear` method?

lib/helper/Playwright.js
```diff
@@ -1526,13 +1526,8 @@ class Playwright extends Helper {
...
     const el = els[0];
-    const tag = await el.getProperty('tagName').then(el => el.jsonValue());
-    const editable = await el.getProperty('contenteditable').then(el => el.jsonValue());
-    if (tag === 'INPUT' || tag === 'TEXTAREA') {
-      await this._evaluateHandeInContext(el => el.value = '', el);
-    } else if (editable) {
-      await this._evaluateHandeInContext(el => el.innerHTML = '', el);
-    }
+
+    await el.clear();
```
hatufacci, line 1550: @kobenguyent 
why we need clear() ? in fillField() method ? we have separated method for this.
Do we need clear field before fill something in all cases?

hatufacci (changes requested):

lib/helper/Playwright.js
kobenguyent: I don't know why, it was like that to make sure the field is all clear before we fill something out.

hatufacci: I'm ok with that but i prefer leave such things at the user's discretion :)

kobenguyent: Understood but that behavior was like that before so just don't wanna break existing tests of users. 😊

Arhell (approved):

hatufacci (approved):

test/helper/webapi.js
```diff
@@ -1121,10 +1121,10 @@ module.exports.tests = function () {
...
       await I.amOnPage('/form/example4');
       await I.seeElement('#navbar-collapse-menu');
-      await I._withinBegin('#register');
-      await I.see('E-Mail');
-      await I.dontSee('Toggle navigation');
-      await I.dontSeeElement('#navbar-collapse-menu');
+      I._withinBegin('#register')
+        .then(() => I.see('E-Mail'))
+        .then(() => I.dontSee('Toggle navigation'))
+        .then(() => I.dontSeeElement('#navbar-collapse-menu'));
```
EgorBodnar, line 1110: Could we switch here to async/await?

kobenguyent: I'm not sure why, but async/await there failed the tests. That's why I updated to .then()

lib/helper/Playwright.js
```diff
@@ -2699,7 +2679,8 @@ async function getXPathForElement(elementHandle) {
 async function findElements(matcher, locator) {
   if (locator.react) return findReact(matcher, locator);
   locator = new Locator(locator, 'css');
-  return matcher.$$(buildLocatorString(locator));
+
+  return matcher.locator(buildLocatorString(locator)).all();
```
EgorBodnar: It could be a huge issue for projects with dynamic rendering. 
[$$](https://playwright.dev/docs/api/class-elementhandle#element-handle-query-selector-all) from ElementHandle  awaits for all matching elements, but [.all()](https://playwright.dev/docs/api/class-locator#locator-all) from Locator not, it immediately returns whatever is present in the page. 

So as the function `findElements()` use as entry point for locating, it's dangerous to stay as is

DavertMik: @EgorBodnar I think it is dangerous only if we would not call this methods in sequence

If we would have it like this:

```js
const els = await findElements()
await I.doSomeOtherStuff()
els[0].click()
```
the behavior of `$$` and `locator().all()` would be different
but if we found elements, picked the first one, clicked it, it should be the same

maybe there is slight difference for internal Playwright retries but I think they should stabilize execution, not cause failures

EgorBodnar: I think we really have no way out. Users will have to pay more attention to waiting for the elements to load.

I suggest leaving the implementation of the `findElements()` function as @kobenguyent  proposed, but also adding a `findElement()` function that will not use the `.all()` method. And then in the Playwright helper, we should use the new `findElement()` function as much as possible, where work with a single element is required. This way we can increase stability.

@DavertMik @kobenguyent what do you think?

kobenguyent: @EgorBodnar correct me if I'm wrong, when using `findElements()`, we just grab the first element from the list to interact with, so may you share what the differences when we use the approach to use `findElement()` where it is possible?

EgorBodnar: [locator.all()](https://playwright.dev/docs/api/class-locator#locator-all) does not wait for elements to match the locator, but without `.all()` waits

So to make selectors more stable better to use it as often as we can without  the`.all()` method

DavertMik: I'm not sure I understand all details here but here is the plan:

@kobenguyent try to follow suggestions from @EgorBodnar 

when you think the fix is ready I will **publish a beta release from this branch** and we will ask community to test it. If everything is ok, I will merge it and make a proper release

kobenguyent: hey @DavertMik, updated as suggestions from @EgorBodnar

EgorBodnar (approved):

DavertMik:
@kobenguyent @EgorBodnar 

Published this PR as `codeceptjs@3.5.4-beta.1`
Please check this new version

kobenguyent:
> @kobenguyent @EgorBodnar 
> 
> 
> 
> Published this PR as `codeceptjs@3.5.4-beta.1`
> 
> Please check this new version
> 
> 

Thank you @DavertMik 😊

ngraf:
I ran my biggest test suite with a lot of different locators and techniques against 3.5.4-beta.1 with Playwright.
I discovered one issue:

🐞 **Issue**: `I.switchTo(locator)` step does not fail when locator was not found.
*Expected*
`switchTo` does fail when locator does not match any element.
*Actual*
`switchTo` does not fail when locator does not match any element.

Proof: 
This scenario fails as expected with latest stable version, but is unexpectedly green in tested beta version:
```
Scenario('Playwright locator Bug with iframe not found', async ({I}) => {
    I.amOnPage('https://www.w3schools.com/html/html_iframe.asp')
    // I.switchTo('iframe[ title = "W3Schools HTML Tutorial']) // This would be the correct locator
    I.switchTo('iframe[ title = "W3Schools HTML Tutorial    -> I ADDED THIS TEXT WITH THE GOAL TO NOT FIND IFRAME. TEST SHOULD FAIL BUT IT DOES NOT WITH CodeceptJS version 3.5.4-beta.1 <-"]')

})
```

kobenguyent:
> I.amOnPage('https://www.w3schools.com/html/html_iframe.asp')
>     // I.switchTo('iframe[ title = "W3Schools HTML Tutorial']) // This would be the correct locator
>     I.switchTo('iframe[ title = "W3Schools HTML Tutorial    -> I ADDED THIS TEXT WITH THE GOAL TO NOT FIND IFRAME. TEST SHOULD FAIL BUT IT DOES NOT WITH CodeceptJS version 3.5.4-beta.1 <-"]')

Hi @ngraf I tried your scenario against 3.x and nothing fails, so I think it's not the issue from this Locator changes.

```
    it.only('should switch reference to iframe content', () => {
      I.amOnPage('https://www.w3schools.com/html/html_iframe.asp');
      // I.switchTo('iframe[ title = "W3Schools HTML Tutorial']) // This would be the correct locator
      I.switchTo('iframe[ title = "W3Schools HTML Tutorial    -> I ADDED THIS TEXT WITH THE GOAL TO NOT FIND IFRAME. TEST SHOULD FAIL BUT IT DOES NOT WITH CodeceptJS version 3.5.4-beta.1 <-"]')

    });
```

```
➜  CodeceptJS git:(3.x) ✗ npm run test:unit:webbapi:playwright

> codeceptjs@3.5.3 test:unit:webbapi:playwright
> mocha test/helper/Playwright_test.js



  Playwright
    #switchTo
      ✔ should switch reference to iframe content


  1 passing (825ms)

➜  CodeceptJS git:(3.x) ✗ 

```

ngraf:
Ok, I guess you are right and the issue is not caused by this feature branch then.

PR #3748, opened 2023-07-10
feat(playwright): checking network stuff

kobenguyent:
- feat(playwright): checking network stuff
- implemention based on https://gist.github.com/ngraf/554b8ea51a6bcda47cd8975c76a2343f from @ngraf 

Applicable helpers:
- [ ] Playwright

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/helper/Playwright.js
+++ b/lib/helper/Playwright.js
@@ -2683,10 +2683,17 @@ class Playwright extends Helper {
       this.debugSection(`REQUEST: ${information}`);
 
       this.requests.push(information);
+      return this._waitForAction();
     });
   }
 
-  grabRecordingTraffic() {
+  /**
+  *  Grab the recording network traffics
+  *
+  * @return { Promise<Array<any>> }
+  *
+  */
+  async grabRecordingTraffic() {
     if (!this.recordedAtLeastOnce) {
       throw new Error('Failure in test automation. You use "I.grabRecordingTraffic", but "I.startRecordingTraffic" was never called before.');
     }
@@ -2698,13 +2705,14 @@ class Playwright extends Helper {
    *
    * @param url URL to block . URL can contain * for wildcards. Example: https://www.example.com* to block all traffic for that domain.
    */
-  blockTraffic(url) {
+  async blockTraffic(url) {
     this.page.route(url, (route) => {
       route
         .abort()
         // Sometimes it happens that browser has been closed in the meantime. It is ok to ignore error then.
         .catch((e) => {});
     });
+    return this._waitForAction();
   }
 
   /**
@@ -2715,7 +2723,7 @@ class Playwright extends Helper {
    * @param responseString string The string to return in fake response.
    * @param contentType Content type of response. If not specified default value 'application/json' is used.
    */
-  mockTraffic(urls, responseString, contentType = 'application/json') {
+  async mockTraffic(urls, responseString, contentType = 'application/json') {
     // Required to mock cross-domain requests
     const headers = { 'access-control-allow-origin': '*' };
 
@@ -2737,6 +2745,7 @@ class Playwright extends Helper {
         });
       });
     });
+    return this._waitForAction();
   }
 
   /**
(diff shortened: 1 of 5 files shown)


ngraf:
I am happy to see my unpolished custom helper being picked up and potentially becoming part of official CodeceptJS library.
The code as it is in the Gist is already battle proven. I used this helper a lot over many months of productive execution. 

There is one feature missing that I would consider important before it is mature enough for community:
Support to check POST parameters with `I.seeTraffic`. Currently only GET parameters are supported. I am sure it is doable to check POST parameters as well. I was just lagging the need and time to implement this as well.

DavertMik:
@ngraf could you improve this PR by adding usage examples into docblocks? Please, please!

@kobenguyent is it possible for you to implement 

> Support to check POST parameters with I.seeTraffic

if not, this limitation should be mentioned in documentation

kobenguyent:
> @ngraf could you improve this PR by adding usage examples into docblocks? Please, please!
> 
> @kobenguyent is it possible for you to implement
> 
> > Support to check POST parameters with I.seeTraffic
> 
> if not, this limitation should be mentioned in documentation

updated the docs along with POST verifications

ngraf:
> @ngraf could you improve this PR by adding usage examples into docblocks? Please, please!

Ok, I will try to contribute usage examples in doclblocks this week @DavertMik .

lib/helper/Playwright.js
```diff
@@ -2655,6 +2661,265 @@ class Playwright extends Helper {
...
+    });
+  }
+
+  /**
+  *  Grab the recording network traffics
+  *
+  * @return { Promise<Array<any>> }
+  *
+  */
+  async grabRecordingTraffic() {
```
ngraf: I would rather call it `I.grabRecordedTraffic`. 

The reality is that this method just returns the traffic requests recorded so far. Nothing more, nothing less.
When I see name `I.grabRecordingTraffic` I am bit unsure if receive some kind of stream object, because of the active meaning of the word "recording" in english language.

lib/helper/Playwright.js
```diff
@@ -2655,6 +2661,265 @@ class Playwright extends Helper {
...
+        });
+      });
+    });
+    return this._waitForAction();
+  }
+
+  /**
+   * Resets all recorded network requests.
+   */
+  _flushTraffic() {
```
ngraf: I personally would make `I._flushTraffic` even officially public and remove the underscore. `_`.

I can imagine in some test scenarios user wants to be able to flush recorded traffic, for example if you pass a lot of pages during a long user journey, you might want to flush the traffic in between. 
Given you want to assert traffic via "I.seeTraffic" on a 2nd follow-up page,
then you don't want to get fooled by some similar old traffic from 1st page which is still in array of recorded traffic.

ngraf:
> @ngraf could you improve this PR by adding usage examples into docblocks? Please, please!


As promised I created a pull-request that contains some examples for the docblocks:
https://github.com/codeceptjs/CodeceptJS/pull/3758

DavertMik (approved):

Arhell (approved):

lib/helper/Playwright.js
```diff
@@ -2655,6 +2661,266 @@ class Playwright extends Helper {
...
+   * @param {Object} opts - options when checking the traffic network.
+   * @param {string} opts.trafficName A name of that request. Can be any value. Only relevant to have a more meaningful error message in case of fail.
+   * @param {string} opts.url Expected URL of request in network traffic
+   * @param {Object} [opts.parameters] Expected parameters of that request in network traffic
+   * @param {Object} [opts.requestPostData] Expected that request contains post data in network traffic
+   * @param {number} [opts.timeout] Timeout to wait for request in seconds. Default is 10 seconds.
+   * @return { Promise<*> }
+   */
+  async seeTraffic({
+    trafficName, url, parameters, requestPostData, timeout = 10,
```
ngraf: I suggest to rename the key `trafficName` to just `name`. 

The context is already clear. I see no reason why to add prefix "traffic". Only clutters up readability. In the examples by the way "name" is already used ;)

lib/helper/Playwright.js
```diff
@@ -2655,6 +2661,266 @@ class Playwright extends Helper {
...
+  }) {
+    if (!this.recording || !this.recordedAtLeastOnce) {
+      throw new Error('Failure in test automation. You use "I.seeInTraffic", but "I.startRecordingTraffic" was never called before.');
+    }
+
```
ngraf, line 2845: When you accidentically run `I.seeTraffic('https://www.example.com')` with just a url string as parameter instead of required object with required keys, the test will pass, which is not really a good behavior.
I stumbled in this pitfall and it took me 10 minutes to figure out what I did wrong and why the test was passing that I expected to fail.
Therefor I suggest to add something like this in here:

```
        if (!trafficName) {
            throw new Error('Missing required key "trafficName" in object given to "I.seeInTraffic".');
        }
        if (!url) {
            throw new Error('Missing required key "url" in object given to "I.seeInTraffic".');
        }
```

lib/helper/Playwright.js
```diff
@@ -2655,6 +2661,266 @@ class Playwright extends Helper {
...
+    assert.fail(`Method "getTrafficUrl" failed: No request found in traffic that matches ${urlMatch}`);
+  }
+
+  /**
+   * Verifies that a certain request is not part of network traffic.
+   * @param {string} name A name of that request. Can be any value. Only relevant to have a more meaningful error message in case of fail.
+   * @param {string} url Expected URL of request in network traffic
+   */
+  async dontSeeTraffic(name, url) {
```
ngraf: We whould make the make the paramters for "I.dontSeeTraffic" also an object to be more consistent to how "I.seeTraffic" consumes paramters.

lib/helper/Playwright.js
kobenguyent: thanks for your review, that was a complain back to the original implementation when using `name` there, that's why I changed to `trafficName`. However, changing this param to an object won't cause the issue anymore.

PR #3758, opened 2023-07-12
added examples in docblocks of new network test steps in Playwright helper

ngraf:
- I created examples in docblock of the new steps in Playwright helper for network testing.
- Minor Functional adjustments/improvements:
   - I removed `async` from  `dontSeeTraffic`, because there is no need for it. The evaluating code inside runs synchronously.
    - I make `getTrafficUrl` fail hard when recording of traffic has not yet started. I think this is helpful for the test automation engineer. I cannot imagine there is usecase where this new behavior is unwanted.
- Resolves request for support from @DavertMik at https://github.com/codeceptjs/CodeceptJS/pull/3748#issuecomment-1629711741

Applicable helpers:

- [ x ] Protractor

## Type of change

- [ ] :fire: Breaking changes
- [ ] :bug: Bug fix
- [ x ] :clipboard: Documentation changes/updates
- [ ] :hotsprings: Hot fix
- [ x ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/helper/Playwright.js
+++ b/lib/helper/Playwright.js
@@ -2667,7 +2667,7 @@ class Playwright extends Helper {
    * This also resets recorded network requests.
    *
    * ```js
-   * await I.startRecordingTraffic();
+   * I.startRecordingTraffic();
    * ```
    *
    * @return {Promise<void>}
@@ -2712,10 +2712,10 @@ class Playwright extends Helper {
    * Examples:
    *
    * ```js
-   * await I.blockTraffic('http://example.com/css/style.css');
-   * await I.blockTraffic('http://example.com/css/*.css');
-   * await I.blockTraffic('http://example.com/**');
-   * await I.blockTraffic(/\.css$/);
+   * I.blockTraffic('http://example.com/css/style.css');
+   * I.blockTraffic('http://example.com/css/*.css');
+   * I.blockTraffic('http://example.com/**');
+   * I.blockTraffic(/\.css$/);
    * ```
    *
    * @param url URL to block . URL can contain * for wildcards. Example: https://www.example.com** to block all traffic for that domain. Regexp are also supported.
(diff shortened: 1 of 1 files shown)


DavertMik:
@kobenguyent please merge if you think it is good
Looks overall good to me

lib/helper/Playwright.js
```diff
@@ -2719,11 +2732,19 @@ class Playwright extends Helper {
 
   /**
    * Mocks traffic for URL(s).
-   * This is a powerful feature to manipulate network traffic. Can be used e.g. to stabilize your tests.
+   * This is a powerful feature to manipulate network traffic. Can be used e.g. to stabilize your tests, speed up your tests or as a last resort to make some test scenarios even possible.
+   *
+   * Examples:
+   *
+   * ```js
+   * await I.mockTraffic('/api/users/1', '{ id: 1, name: 'John Doe' }');
```
DavertMik: do we need await here?

lib/helper/Playwright.js
```diff
@@ -2705,7 +2709,16 @@ class Playwright extends Helper {
   /**
    * Blocks traffic for URL.
    *
-   * @param url URL to block . URL can contain * for wildcards. Example: https://www.example.com* to block all traffic for that domain.
+   * Examples:
+   *
+   * ```js
+   * await I.blockTraffic('http://example.com/css/style.css');
```
DavertMik: do we always need await?

lib/helper/Playwright.js
```diff
@@ -2840,10 +2865,22 @@ class Playwright extends Helper {
 
   /**
    * Returns full URL of request matching parameter "urlMatch".
+   *
+   * Examples:
+   *
+   * ```js
+   * await I.getTrafficUrl('https://api.example.com/session');
+   * await I.getTrafficUrl(/session.*start/);
```
DavertMik: I just thought, maybe we should use `grab` prefix? 
juist to keep one standard

DavertMik (changes requested):

lib/helper/Playwright.js
ngraf: I agree with DavertMilk.
I also would prefer "grabTrafficUrl".

lib/helper/Playwright.js
ngraf: No. We don't need actually.
Good catch.
I made hands-on tests and they all worked just fine without `await`.

-> I updated the examples in docblock. I didn't yet change it in the functional code to make the methods sync, that don't need to be async. A
ctually almost all new steps are sync methods except `await I.seeTraffic`
`seeTraffic` is async because we decided to have waiting mechanism in there for up to 10 seconds (default value).

Arhell (approved):

DavertMik:
@kobenguyent please merge it

kobenguyent (approved):

PR #3814, opened 2023-08-15
fix(playwright): userDataDir and locale config options show error this.browser.contexts is not a function

### LINKED ISSUE #2887: userDataDir and locale config options throws error this.browser.contexts is not a function (Playwright)

Change default browser language setting config options **userDataDir** and **locale** in **codecept.conf.js**

#### What do you get instead?
It works and changes the browser language correctly (Chrome and Firefox), but after finishing the execution of the tests (at the moment the browser is closed), a CodeceptJS internal error is showed in the console (although all test pass successfully):

```bash
# [Browser:Error] JSHandle@cannot get token JSHandle@error= JSHandle@error
    [1] Retrying... Attempt #3
  ✔ OK in 30147ms

TypeError: this.browser.contexts is not a function
    at Playwright._after (/repos/goto-app/node_modules/codeceptjs/lib/helper/Playwright.js:337:43)
    at /repos/goto-app/node_modules/codeceptjs/lib/listener/helpers.js:26:69
 › [Session] cleaning cookies and localStorage

```

```js
// import { UserType } from './user-type'

Feature('Login and Logout').retry(3)

Scenario(
  'Login, Verify GoToContent is Visible, Logout and Verify Landing Page is Displayed',
  async ({ I, LoginComponent }) => {
    const user = await I.getUser(UserType.gotoAppUserWithoutAvatar)
    LoginComponent.login(user)
    LoginComponent.acceptTos()
    LoginComponent.verifyGoToContentVisible()
    LoginComponent.logout()
    LoginComponent.verifyLandingPageDisplayed()
  },
)
```

### Details

* CodeceptJS version: 3.0.6
* NodeJS Version: v12.19.0
* Operating System: macOS BigSur 11.2.3
* Playwright version: 1.10.0
* Configuration file: 

```js
# 
const { DEFAULT_HOST, USER_DATA_PATH } = require('./host.config')
const { setHeadlessWhen } = require('@codeceptjs/configure')

setHeadlessWhen(process.env.baseUrl)

require('ts-node/register')
exports.config = {
  tests: './*.e2e-spec.ts',
  output: './output',
  helpers: {
    Playwright: {
      url: process.env.baseUrl || DEFAULT_HOST,
      show: true,
      browser: 'chromium',
      waitForAction: 250,
      waitForNavigation: 'networkidle0',
      chromium: {
        userDataDir: `${USER_DATA_PATH}${process.env.BROWSER}`,
        channel: process.env.BROWSER,
        locale: 'fr-CA',
      },
    },
    LoginHelper: {
      require: './login-helper.ts',
    },
    BrowserHelper: {
      require: './browser-helper.ts',
    },
  },
  include: {
    I: './steps-file.ts',
    LoginComponent: './login-component.ts',
  },
  bootstrap: null,
  mocha: {},
  multiple: {
    parallel: {
      browsers: ['chromium', 'firefox'],
    },
  },
  name: 'e2e',
  plugins: {
    pauseOnFail: {},
    retryFailedStep: {
      enabled: true,
      retries: 6,
    },
    tryTo: {
      enabled: true,
    },
    screenshotOnFail: {
      enabled: false,
    },
  },
}

```

kobenguyent:
- Resolves #2887 

Applicable helpers:
- [ ] Playwright

## Type of change
- [ ] :bug: Bug fix

## Checklist:

- [ ] Tests have been added


Arhell (approved):

lib/helper/Playwright.js
```diff
@@ -520,7 +520,15 @@ class Playwright extends Helper {
 
     // close other sessions
     try {
-      const contexts = await this.browser.contexts();
+      let contexts = [];
+
+      // when using userDataDir, this.browser is now BrowserContext.
```
HaruhikoKubo: Sorry for adding a comment suddenly but I have a question.
At line 543, this method return `this.browser` object, does this also work for BrowserContext for the further test scenarios?

kobenguyent: I guess in this case, this is the `_after` hook therefore I don't think the returned this.browser would be used further. Are you encountering another issue there?

HaruhikoKubo: Thanks for checking the comment @kobenguyent  I faced this error at the beginning of the 2nd scenario execution if there are multiple scenarios. I tried to execute it sequentially.

`    [1] Error | Error: browserContext.newPage: Target page, context or browser has been closed
Error: browserContext.newPage: Target page, context or browser has been closed`

kobenguyent: @HaruhikoKubo did it work with you before this patch? Like you see the error but the tests are still passed?

HaruhikoKubo: Sorry I am wrong but yes this is because,
before the patch, I got this as same as original issue.
`TypeError: this.browser.contexts is not a function
    at Playwright._after (/node_modules/codeceptjs/lib/helper/Playwright.js:522:43)
    at /node_modules/codeceptjs/lib/listener/helpers.js:26:69`
    
Then `await Promise.all(contexts.map(c => c.close()));` at line 528 was not reached.
However having this patch, we can reach that section I guess. 


Note: I use `restart: "context"` option

If we use `restart: "session"` option, we can get the  same `TypeError: this.browser.contexts is not a function` error at line 3436. we also need to have a fix for that option.

kobenguyent: @HaruhikoKubo may you give a try, the issue has been addressed.

HaruhikoKubo: Thank you @kobenguyent , I think this time it works for me without any error.

PR #3953, opened 2023-10-26
fix(workers): event improvements

kobenguyent:
- Emit the `workers.result` event
```
You could get test stats when running with workers

const { event } = require('codeceptjs');

module.exports = function() {

  event.dispatcher.on(event.workers.result, function (result) {

    console.log(result);

  });
}

// in console log
FAIL  | 7 passed, 1 failed, 1 skipped   // 2s
{
    "tests": {
        "passed": [
            {
                "type": "test",
                "title": "Assert @C3",
                "body": "() => { }",
                "async": 0,
                "sync": true,
                "_timeout": 2000,
                "_slow": 75,
                "_retries": -1,
                "timedOut": false,
                "_currentRetry": 0,
                "pending": false,
                "opts": {},
                "tags": [
                    "@C3"
                ],
                "uid": "xe4q1HdqpRrZG5dPe0JG+A",
                "workerIndex": 3,
                "retries": -1,
                "duration": 493,
                "err": null,
                "parent": {
                    "title": "My",
                    "ctx": {},
                    "suites": [],
                    "tests": [],
                    "root": false,
                    "pending": false,
                    "_retries": -1,
                    "_beforeEach": [],
                    "_beforeAll": [],
                    "_afterEach": [],
                    "_afterAll": [],
                    "_timeout": 2000,
                    "_slow": 75,
                    "_bail": false,
                    "_onlyTests": [],
                    "_onlySuites": [],
                    "delayed": false
                },
                "steps": [
                    {
                        "actor": "I",
                        "name": "amOnPage",
                        "status": "success",
                        "agrs": [
                            "https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/POST"
                        ],
                        "startedAt": 1698760652610,
                        "startTime": 1698760652611,
                        "endTime": 1698760653098,
                        "finishedAt": 1698760653098,
                        "duration": 488
                    },
                    {
                        "actor": "I",
                        "name": "grabCurrentUrl",
                        "status": "success",
                        "agrs": [],
                        "startedAt": 1698760653098,
                        "startTime": 1698760653098,
                        "endTime": 1698760653099,
                        "finishedAt": 1698760653099,
                        "duration": 1
                    }
                ]
            }
        ],
        "failed": [],
        "skipped": []
    }
}
```

CodeceptJS also exposes the env var `process.env.RUNS_WITH_WORKERS` when running tests with `run-workers` command so that you could handle the events better in your plugins/helpers

```js
const { event } = require('codeceptjs');

module.exports = function() {
    // this event would trigger the  `_publishResultsToTestrail` when running `run-workers` command
  event.dispatcher.on(event.workers.result, async () => {
    await _publishResultsToTestrail();
  });
  
  // this event would not trigger the  `_publishResultsToTestrail` multiple times when running `run-workers` command
  event.dispatcher.on(event.all.result, async () => {
      // when running `run` command, this env var is undefined
    if (!process.env.RUNS_WITH_WORKERS) await _publishResultsToTestrail();
  });
}
```

## Type of change
- [ ] :bug: Bug fix

## Checklist:

- [ ] Tests have been added


docs/internal-api.md
```diff
@@ -96,6 +96,51 @@ module.exports = function() {
...
+    failures: 1,
+  tests: 9,
+  pending: 1,
+  start: 2023-10-26T13:15:57.685Z,
+  end: 2023-10-26T13:15:59.880Z,
+  duration: 2195
+}
+```
+
+CodeceptJS also expose the env var `process.env.RUN_WITH_WORKERS` when running tests with `run-workers` command so that you could handle the events better in your plugins/helpers
```
DavertMik:
```suggestion
CodeceptJS also exposes the env var `process.env.RUN_WITH_WORKERS` when running tests with `run-workers` command so that you could handle the events better in your plugins/helpers
```

lib/workers.js
```diff
@@ -338,6 +338,8 @@ class Workers extends EventEmitter {
     this.stats.start = new Date();
     recorder.startUnlessRunning();
     event.dispatcher.emit(event.workers.before);
+    // expose this to manage events better when using run-workers command
+    process.env.RUN_WITH_WORKERS = 'true';
```
DavertMik:
```suggestion
    process.env.RUNS_WITH_WORKERS = 'true';
```

DavertMik (changes requested):
Sorry, I don't think we can accept this implementation
This will cause extra side effects from main thread like starting/closing browsers, which is not required there

Arhell (approved):

exotlv:
@DavertMik can you please re-check changes ?

exotlv:
@DavertMik  can you please review? Need this changes/fixes

docs/internal-api.md
```diff
@@ -96,7 +96,117 @@ module.exports = function() {
   });
 }
 ```
+You could get test stats when running with workers
```
DavertMik: I like this but I think it is better to move this part of the documentation into 

https://github.com/codeceptjs/CodeceptJS/blob/3.x/docs/parallel.md

At least this is the place where user will look for it

DavertMik (approved):

PR #3962, opened 2023-10-30
Bugfix/ai html updates

### LINKED ISSUE #3829: html-minifier@4.0.0 Regular Expression Denial of Service vulnerability

Clean up the codebase security / performance alerts

We recently implemented Snyk and are going through a cleanup process to see where we can tidy up things in the system. Our only outlier seems to be this alert that doesn't seem to be high priority but would be nice to clean up as well.

I do see this package has not been updated in a while, is there a more maintained alternative that can be put in place to address this?

#### What do you see

https://cwe.mitre.org/data/definitions/1333.html

image

### Details

* CodeceptJS version: 3.5.3

DavertMik:
- replaced minify library with a modern and more secure fork. Fixes #3829
- AI class is implemented as singleton
- refactored heal.js plugin to work on edge cases
- add configuration params on number of fixes performed by ay heal 
- improved `recorder` class to add more verbose log
- improved `recorder` class to ignore some of errors


DavertMik:
Released as beta 

codeceptjs@3.5.7-beta.1

package.json
```diff
@@ -86,7 +86,7 @@
     "fn-args": "4.0.0",
     "fs-extra": "8.1.0",
     "glob": "6.0.1",
-    "html-minifier": "4.0.0",
+    "html-minifier-terser": "^7.2.0",
```
kobenguyent, line 91: shall we remove ^?

lib/ai.js
```diff
@@ -35,13 +40,17 @@ class AiAssistant {
     this.openai = new OpenAIApi(configuration);
   }
 
-  setHtmlContext(html) {
+  static getInstance() {
+    return aiInstance || new AiAssistant();
```
kobenguyent, line 44: out of curiosity, I saw `const aiInstance = null;` at line 19 and it is not assigned, does this code here alway return the `new AiAssistant(); `

Arhell (approved):

PR #3966, opened 2023-11-01
fix: appium helper docs update

kobenguyent:

The Appium core team does not maintain Appium 1.x anymore since the [1st of January 2022](https://github.com/appium/appium). Recent versions of all officially supported platform drivers are no longer compatible with Appium 1.x, and require Appium 2 to run.

- This Appium 1.x support will be removed in next major release
- Adding iOS Appium Tests

Applicable helpers:
- [x] Appium

## Checklist:

- [x] Tests have been added


mirao:
I migrated from Appium 1.x to 2.x and was a bit confused that the message
```
The Appium core team does not maintain Appium 1.x anymore since the 1st of January 2022. Please migrating to Appium 2.x by adding appiumV2: true to your config.\nThis Appium 1.x support will be removed in next major release
```
is displayed even when I run a test with Appium 2.x (with `appiumV2: true` in a config).

Better would be to display the message only for Appium 1.x configurations.

kobenguyent:
Thanks @mirao that sounds reasonable!

PR #4043, opened 2023-11-30
feat: expose WebElement

### LINKED ISSUE #4042: Introduce CodecetptJS WebElement Class to mirror chosen helpers' element instance
### Discussed in https://github.com/codeceptjs/CodeceptJS/discussions/3001

<div type='discussions-op-text'>

<sup>Originally posted by **m3t4n** August 24, 2021</sup>
While we are using CodeceptJS with test framework helpers, the main entity for a DOM object is not exposed.
These are referred to as different names in frameworks. `ElementHandle` for Playwright, `WebElement` for WebDriverIO, `ElementFinder` for Protractor etc...

CodeceptJS currently is swallowing this exposure and only accepting locator parameters to the wrapped assertion methods such as `I.click('#submit');`. Whereas we should be able to locate element(s) and should be able to perform desired actions on these element(s) accordingly. 

A common use case can be:

1. locate multiple elements matching the selector criteria in the page
2. populate them in an Array<CodeceptElement>
3. perform a check for an attribute value on all these elements using a loop.
4. find the matching one
5. perform a click action on that specific element or any further test.

with current design we can only use a hack like below to perform this action (using Playwright for this example):

1. create a variable in `:any` type
2. find the element using `I.usePlayWrightTo(...., ({page}) => { this.localElm = page.$$(...) });`
3. then loop to find your match, again using` I.usePlayWrightTo(...., ({page}) => { this.localElm.foreach(elm => (elm. getProperty(....)........`

This class should bind according to the chosen helper and shall allow us to be exposed to further methods of DOMElement such as getProperty, getInnerHtml, getBoundingBox etc... further more, it will also allow us to perform additional searches within the element, such as: elm.$ or elm.$$....</div>

kobenguyent:
- Resolves #4042

Now we expose the WebElements that are returned by the WebHelper and you could make the subsequence actions on them.

```
// Playwright helper would return the Locator

I.amOnPage('/form/focus_blur_elements');
const webElements = await I.grabWebElements('#button');
webElements[0].click();
```

Applicable helpers:
- [ ] Playwright
- [ ] Puppeteer
- [ ] WebDriver

## Type of change

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/docs/helpers/Playwright.md
+++ b/docs/helpers/Playwright.md
@@ -1389,6 +1389,36 @@ let inputs = await I.grabValueFromAll('//form/input');
 
 Returns **[Promise][22]&lt;[Array][10]&lt;[string][9]>>** attribute value
 
+### grabWebElement
+
+Grab WebElement for given locator
+Resumes test execution, so **should be used inside an async function with `await`** operator.
+
+```js
+const webElement = await I.grabWebElement('#button');
+```
+
+#### Parameters
+
+-   `locator` **([string][9] | [object][6])** element located by CSS|XPath|strict locator.
+
+Returns **[Promise][22]&lt;any>** WebElement of being used Web helper
+
+### grabWebElements
+
+Grab WebElements for given locator
+Resumes test execution, so **should be used inside an async function with `await`** operator.
+
+```js
+const webElements = await I.grabWebElements('#button');
+```
+
+#### Parameters
+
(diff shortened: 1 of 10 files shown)


Arhell (approved):

DavertMik:
I like this feature, thanks @kobenguyent !

PR #4157, opened 2024-01-26
fix(webApi): selectOption method

dyaroman:

I was trying to select an option by value '1999' from year of birth select and I used Playwright 'selectOption' method for this. But I got error:
image

Then I debugged method 'selectOption' and found that '.textContent()' method return value with spaces: ' 1999 ', that why I got error. I think this is a realistic case, many template engines can produce additional spaces.

Here HTML of select element:
image

I suggest to trim the value returned by the 'textContent' method. If 'textContent' returned 'null' we will get type error, but 'catch' block will catch this error and return 'option' argument.
What do you think?

https://github.com/codeceptjs/CodeceptJS/blob/439364ec0a9e0787de2a112f851599715a2538e5/lib/helper/Playwright.js#L1808

Applicable helpers:

- [x] Playwright
- [ ] Puppeteer
- [ ] WebDriver
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] TestCafe

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] stepTimeout
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [x] :bug: Bug fix
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/helper/Playwright.js
+++ b/lib/helper/Playwright.js
@@ -1805,7 +1805,7 @@ class Playwright extends Helper {
     let optionToSelect = '';
 
     try {
-      optionToSelect = await el.locator('option', { hasText: option }).textContent();
+      optionToSelect = (await el.locator('option', { hasText: option }).textContent()).trim();
     } catch (e) {
       optionToSelect = option;
     }
--- a/test/data/app/view/form/select_additional_spaces.php
+++ b/test/data/app/view/form/select_additional_spaces.php
@@ -1,0 +1,25 @@
+<html>
+<body>
+<form action="/form/complex" method="POST">
+    <label for="age">Select your age</label>
+    <select name="age" id="age">
+        <option value="child">
+            below 13
+        </option>
+        <option value="teenage">
+            13-21
+        </option>
+        <option value="adult">
+            21-60
+        </option>
+        <option value="oldfag" selected="selected">
+            60-100
+        </option>
+        <option value="dead">
+            100-210
+        </option>
+    </select>
+    <input type="submit" value="Submit" />
+</form>
+</body>
+</html>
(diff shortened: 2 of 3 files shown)


dyaroman:
Or we can get value from `value` attribute:
```js
optionToSelect = await el.locator('option', { hasText: option }).getAttribute('value');
```
What do you think?

lib/helper/Playwright.js
```diff
@@ -1805,7 +1805,7 @@ class Playwright extends Helper {
     let optionToSelect = '';
 
     try {
-      optionToSelect = await el.locator('option', { hasText: option }).textContent();
+      optionToSelect = await el.locator('option', { hasText: option }).textContent().then(s => s.trim());
```
kobenguyent:
```suggestion
      optionToSelect = (await el.locator('option', { hasText: option }).textContent()).trim();
```
maybe this? wdyt?

kobenguyent:
I guess the other Web helpers may get the same error, would you mind checking it? and add tests for it here https://github.com/codeceptjs/CodeceptJS/blob/3.x/test/helper/webapi.js#L391

dyaroman:
I will try.

dyaroman:
@kobenguyent please take a look. I have run tests local, all works good. Only `WebDriver_test.js` can't run local, don't have selenium.
Also added new test and markup for it.

kobenguyent:
thanks @dyaroman!

PR #4230, opened 2024-03-04
Draft: Healers introduced, AI refactored

### LINKED ISSUE #3731: Make OpenAI prompts configurable
Allow users to set OpenAI prompts inside configuration under `ai` section

DavertMik:
Resolves #3731 
This PR introduces a new concept of healers, basically taken out of refactored AI self-healing implementation.

:tada:  Please try it in beta version"

```
npm i codeceptjs@3.6.0-beta.1.ai-healers
```

### What is a Healer?

Healer is a declarative way of fixing flaky tests.

Instead of trying to guess all possible errors and reactions instead in the code, you declare a set of recipes that will be executed on a failure. Let's say an error happened and from the error and context you may guess it can be fixed by just reloading browser page and trying once again. This
 allows you to write a healer recipe:

```js
heal.addRecipe('reloadPage', {
  priority: 1,
  steps: [
    'click',
  ],
  fn: async ({ error, step }) => {
    return async ({ I }) => {
      I.reloadPage();
      I.wait(1);
      return step.run();
    };
  },
});
```
This recipe will be fired on every failed "click" step. It can be improved if you analyze more details like checking the error message, page elements, etc before applying it.

Another example of heal recipes is replacing `fillField` with `type`+`click` combination in a non-standard input.

```js
heal.addRecipe('clickAndType', {
  priority: 1,
  steps: [
    'fillField',
    'appendField',
  ],
  suggest: true,
  fn: async ({ step }) => {
    const locator = step.args[0];
    const text = step.args[1];

    return ({ I }) => {
      I.click(locator);
      I.wait(1); // to open modal or something
      I.type(text);
    };
  },
});
```
If a fillField has failed because the field can't be identified as "input" or "contentediable" you can still try to emulate user actions click and fill data from keyboard and check where it goes.

If a recipe fixes a test, it will print you a suggestion on how to replace your test to fix an error.

Healing recipes can also be used for:

* restarting server when it is gone
* sending API to update/create data when it is not found
* re-running part of the scenario to fix rendering issues
* etc...

Basically, they can define how you, would fix the error, if you could just stop execution, perform some actions and continue test from this point.

## AI Healer

AI module was reworked to provide a healer recipe as well. In this case code suggestions can be retrieved from AI provider like OpenAI, Azure OpenAI, Mistral, etc. 

image

Current implementation decouples healing from AI component, and makes prompts as well as AI providers to be configurable. **Now AI is no more limited to OpenAI API**. You need to configure it inside `codecept.conf.js`:

```js
     ai: {
       request: async (messages) => {
         const OpenAI = require('openai');
         const openai = new OpenAI({ apiKey: process.env['OPENAI_API_KEY'] })
         const response = await openai.chat.completions.create({
          model: 'gpt-3.5-turbo-0125',
          messages,
         });
         return response?.data?.choices[0]?.message?.content;
       }
     }
```
AI Changes:

* added `--ai` flag to expliticly enable ai mode
* added `maxTokens` setting to limit usage for AI in one session
* added naive token counter to prevent overuse
* added request config method to configure which AI provider is used
* added response config method that helps transforming AI response into CodeceptJS code
* renamed OpenAI helper to AI

## TODO

- [x] Write documentation
- [x] Try more examples with OpenAI and Claude 
- [ ] Collect feedback


lib/ai.js
```diff
@@ -1,47 +1,112 @@
...
+        I have opened web page and I want to use CodeceptJS to ${input} on this page
+        Provide me valid CodeceptJS code to accomplish it
+        Use only locators from this HTML: \n\n${html}`,
+  },
+  { role: 'user', content: 'Propose only CodeceptJS steps code. Do not include Scenario or Feature into response' },
+  ],
+
+  healStep: (html, { step, error, prevSteps }) => [
+    { role: 'user', content: 'As a test automation engineer I am testing web application using CodeceptJS.' },
+    { role: 'user', content: `I want to heal a test that fails. Here is the list of executed steps: ${prevSteps.map(s => s.toString()).join(', ')}` },
```
kobenguyent:
```suggestion
    { role: 'user', content: `I want to heal this failed test. Here is the list of executed steps: ${prevSteps.map(s => s.toString()).join(', ')}` },
```

lib/ai.js
```diff
@@ -50,98 +115,93 @@ class AiAssistant {
...
-    return parseCodeBlocks(response);
+    return this.config.response(response);
+  }
+
+  calculateTokens(messages) {
+    // we implement naive approach for calculating tokens
+    // this approach was tested via https://platform.openai.com/tokenizer
+    // we need it to display current usage tokens usage so users could analyze effectiveness of AI
+
+    const inputString = messages.map(m => m.content).join(' ');
```
kobenguyent:
```suggestion
    const inputString = messages.map(m => m.content).join(' ').trim();
```

lib/heal.js
```diff
@@ -0,0 +1,166 @@
...
+            snippet: codeSnippet,
+          });
+
+          recorder.add('healed', () => output.print(colors.bold.green(`  Code healed successfully by ${suggestion.name}`), colors.gray('(no errors thrown)')));
+          this.numHealed++;
+          // recorder.session.restore();
+          return;
+        } catch (err) {
+          debug('Failed to execute code', err);
+          recorder.ignoreErr(err); // healing ded not help
```
kobenguyent:
```suggestion
          recorder.ignoreErr(err); // healing did not help
```

lib/heal.js
```diff
@@ -0,0 +1,166 @@
...
+          // recorder.catch(() => output.print(colors.bold.red('  Failed healing code')));
```
kobenguyent: shall we need this?

samsaggace:
Should `openai` dependency be removed from the `package.json` ?
I'm asking because the current version has a vulnerability on axios that is fixed in openai v4.

lib/heal.js
DavertMik: I agree, this part should have been changed

DavertMik:
@kobenguyent @samsaggace 
I have update documentation and removed dependency from OpenAI

Could you check my code and documentation for the healers feature?

docs/ai.md
```diff
@@ -14,47 +14,138 @@ Think of it as your testing co-pilot built into the testing framework
...
+Once AI provider is configured run tests with `--ai` flag to enable AI features
 
 ```
-npx create-codeceptjs .
+npx codeceptjs run --ai
 ```
-[Obtain API token](https://platform.openai.com/account/api-keys) from OpenAI. Please check out [their pricing](https://openai.com/pricing) first, as unlike ChatGPT using OpenAI API requires a paid account.
 
-Add `OPENAI_API_KEY` environment variable with a key when running codeceptjs:
+Below we list sample configuration for popular AI providers
```
kobenguyent, line 71: Shall we provide a command to set up this AI configuration?

lib/heal.js
```diff
@@ -0,0 +1,172 @@
...
+          this.fixes.push({
+            recipe: suggestion.name,
+            test: failureContext?.test,
+            step: failedStep,
+            snippet: codeSnippet,
+          });
+
+          recorder.add('healed', () => output.print(colors.bold.green(`  Code healed successfully by ${suggestion.name}`), colors.gray('(no errors thrown)')));
+          this.numHealed++;
+          // recorder.session.restore();
```
kobenguyent, line 131: Could we remove this?

lib/template/heal.js
```diff
@@ -0,0 +1,39 @@
...
+
+heal.addRecipe('clickAndType', {
+  priority: 1,
+  steps: [
+    'fillField',
+    'appendField',
+  ],
+  fn: async ({ step }) => {
+    const locator = step.args[0];
+    const text = step.args[1];
```
kobenguyent, line 31: I'm not sure if that's possible but what if we could do something like const { locator, text } = step?

docs/ai.md
DavertMik: I think no, as we won't be able to update it accordingly to all possible AI providers.
So I added this as documentation + error message but this should be custom to users

lib/heal.js
DavertMik: I think yes but just in case..

lib/template/heal.js
DavertMik: I think it will be:

```js
const [locator, text] = step.args
```

Which also look nice

docs/ai.md
kobenguyent: Yeah, totally understand. Imho, it's just a scaffolding and of course users need to adjust to fit their needs. But at least it's a good way to encourage users to try out the AI helper.

kobenguyent (approved):

examples/heal_recipes.js
```diff
@@ -0,0 +1,23 @@
+const { heal } = require('../lib/index');
+
+heal.addRecipe('clickAndType', {
+  priority: 1,
+  steps: [
```
kobenguyent, line 5: Out of curiosity, does it work with custom steps?

DavertMik: For steps declared in helpers

samsaggace:
Hello @DavertMik @kobenguyent, 
thanks for the hard work on this !
do you know when a new release including these changes is planned ?

thanks again

kobenguyent:
Hi @samsaggace it is now ready! https://codecept.io/changelog/#_3-6-0

samsaggace:
Awesome thanks!

PR #4263, opened 2024-03-18
feat(puppeteer): network traffics manipulation

kobenguyent:
🛩️ Features

* [Puppeteer] Added commands to check network traffics
  * `startRecordingTraffic`
  * `grabRecordedNetworkTraffics`
  * `flushNetworkTraffics`
  * `stopRecordingTraffic`
  * `seeTraffic`
  * `dontSeeTraffic`

Examples:

```js
// recording traffics and verify the traffic
  I.startRecordingTraffic();
  I.amOnPage('https://codecept.io/');
  I.seeTraffic({ name: 'traffics', url: 'https://codecept.io/img/companies/BC_LogoScreen_C.jpg' });
```

```js
// check the traffic with advanced params
  I.amOnPage('https://openai.com/blog/chatgpt');
  I.startRecordingTraffic();
  I.seeTraffic({
    name: 'sentry event',
    url: 'https://images.openai.com/blob/cf717bdb-0c8c-428a-b82b-3c3add87a600',
    parameters: {
      width: '1919',
      height: '1138',
    },
  });
```

Applicable helpers:

- [ ] Playwright
- [ ] WebDriver
- [ ] Puppeteer

## Type of change

## Checklist:

- [ ] Tests have been added


lib/helper/networkTraffics/shares/actions.js
```diff
@@ -0,0 +1,123 @@
+const assert = require('assert');
+const { isInTraffic, createAdvancedTestResults, getTrafficDump } = require('../utils');
```
DavertMik: why do we need `shares` folder for that?

DavertMik (changes requested):
There is no plural for traffic and share

so better to rename "networkTraffics" => `network`

"shares" => "share"

but maybe we don't need this share? 
it is already shared

lib/helper/networkTraffics/shares/actions.js
kobenguyent: Those are the same pieces of code that appeared in 3 helpers. So I wanna move them to a shared folder which is better than copy/paste the same code in 3 helpers.

PR #4264, opened 2024-03-20
feat(ppt): websocket messages

kobenguyent:
- feat(Puppeteer): recording WS messages

Recording WS messages:
```
      await I.startRecordingWebSocketMessages();
      await I.amOnPage('https://websocketstest.com/');
      I.waitForText('Work for You!');
      const wsMessages = I.grabWebSocketMessages();
      expect(wsMessages.length).to.greaterThan(0);

```

flushing WS messages:
```
      await I.startRecordingWebSocketMessages();
      I.amOnPage('https://websocketstest.com/');
      I.waitForText('Work for You!');
      I.flushWebSocketMessages();
      const wsMessages = I.grabWebSocketMessages();
      expect(wsMessages.length).to.equal(0);

```

Applicable helpers:
- [ ] Puppeteer

## Type of change

## Checklist:
- [ ] Tests have been added

### DIFF
--- a/docs/helpers/Playwright.md
+++ b/docs/helpers/Playwright.md
@@ -2217,6 +2217,8 @@ This also resets recorded websocket messages.
 await I.startRecordingWebSocketMessages();
 ```
 
+Returns **void** automatically synchronized promise through #recorder
+
 ### stopMockingRoute
 
 Stops network mocking created by `mockRoute`.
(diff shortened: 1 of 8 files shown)


mirao:
I've just used the feature and noticed that `await I.grabWebSocketMessages()` returns a string instead of an array :bug: .
It seems to be caused by this code that incorrectly adds a string message instead of pushing it to the array?

https://github.com/codeceptjs/CodeceptJS/blob/457ef796ecb102dc37d27652fad6be1307737b83/lib/helper/Puppeteer.js#L2600-L2602

kobenguyent:
Oh that's originally an issue since it's implemented in playwright. But I guess no much usage since. But thanks! Will fix it. 

https://github.com/codeceptjs/CodeceptJS/pull/3789

PR #4357, opened 2024-05-21
fix(webApi): error message of dontSeeCookie

a-stankevich:
The error message reported for 'dontSeeCookie' was copy-pasted from 'seeCookie', while it should be the opposite.
The only helper with correct message is [TestCafe](https://github.com/codeceptjs/CodeceptJS/blob/0e1b86af238ca648f1fbd0964d42abb3d969dc81/lib/helper/TestCafe.js#L1055), so wording from there is applied to other helpers.

Applicable helpers:

- [ ] Nightmare
- [x] Playwright
- [x] Puppeteer
- [ ] WebDriver

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] stepTimeout
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [x] :bug: Bug fix
- [ ] 🧹 Chore
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/helper/Nightmare.js
+++ b/lib/helper/Nightmare.js
@@ -989,7 +989,7 @@ class Nightmare extends Helper {
    */
   async dontSeeCookie(name) {
     const res = await this.browser.cookies.get(name);
-    truth(`cookie ${name}`, 'not to be set').negate(res);
+    truth(`cookie ${name}`, 'to be set').negate(res);
   }
 
   /**
--- a/lib/helper/Nightmare.js
+++ b/lib/helper/Nightmare.js
@@ -989,7 +989,7 @@ class Nightmare extends Helper {
    */
   async dontSeeCookie(name) {
     const res = await this.browser.cookies.get(name);
-    truth(`cookie ${name}`, 'to be set').negate(res);
+    truth(`cookie ${name}`, 'not to be set').negate(res);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


kobenguyent:
I'm not sure if we should have this change. As the error message is from here already indicated not. https://github.com/codeceptjs/CodeceptJS/blob/0e1b86af238ca648f1fbd0964d42abb3d969dc81/lib/assert.js#L63

a-stankevich:
hm, what would be the best way to have the correct error message then? currently it's like this:
image

a-stankevich:
Oh, I think I understand.
I've removed 'not' where negate is used

kobenguyent:
@a-stankevich I saw you also mentioned WebDriver, but don't see the changes there. May you check?

a-stankevich:
> @a-stankevich I saw you also mentioned WebDriver, but don't see the changes there. May you check?

Those changes were done by mistake and have been reverted. I've unchecked related checkboxes now

Issue #4358, opened 2024-05-21
[Asking for help] Process are blocking when throw error in scenario with Retry enabled


Horsty80:
When i run to all test (~70) in 4 workers, at the end, i get stuck with this error message

```bash
✖ "after all" hook: codeceptjs.afterSuite in "Homepage feature" in 0ms
```

* CodeceptJS version: 3.6.2
* NodeJS Version: 20
* playwright: 1.43.0
* allure-codeceptjs: 2.15.1
* Configuration file:

```js

    Playwright: {
      url: testConfig.url,
      show: false,
      locale: "en-EN",
      browser: 'chromium',
      chromium: {
        args: ['--auth-server-whitelist=""'],
      },
      ignoreHTTPSErrors: true,
      fullPageScreenshots: true,
      windowSize: '1920x1080',
      waitForAction: 250,
      waitForTimeout: 10000, 
      waitForNavigation: 'load',
      restart: 'context',
      keepCookies: false,
      keepBrowserState: false
    },
```

I have allready open here https://github.com/codeceptjs/CodeceptJS/issues/3640 an issue before, but no response and the probleme has gone, but it's here again :'(

Horsty80:
Just downgrade to `3.5.15` and it's working fine.
So i presume, something have been added in `3.6.X` version and cause this kind of issue

Horsty80:
This issue, append in local too but it's not on each run, so it's seems to be random...

I don't know how to fix this

Horsty80:
I've worked on a reproductible repo, you can find here https://github.com/Horsty80/codeceptjs-debug

#### Working test ####

```
pnpm i
pnpm working
```

#### Throwing error in After/Before without try/catch ####

```
 pnpm error-1
```

Lead to error but the CI is not block

```bash
  4) Error 1 @error1
       "before each" hook: Before for "test something 6":
     Error: done() called multiple times in hook <Error 1 @error1 "before each" hook: Before for "test something 6"> of file /Users/cyril/Workspace/codeceptjs-debug/tests/error-1_test.ts
      at /Users/cyril/Workspace/codeceptjs-debug/node_modules/.pnpm/codeceptjs@3.6.2_@babel+core@7.24.5_@babel+preset-env@7.24.5_react@18.2.0_typescript@5.4.5/node_modules/codeceptjs/lib/scenario.js:156:55
      at runNextTicks (node:internal/process/task_queues:60:5)
      at processTimers (node:internal/timers:511:9)

[...]

PR #4367, opened 2024-05-29
fix(cli): stale process

kobenguyent:
## Motivation/Description of the PR
- Resolves #4197
- resolves #4358 
- resolves #4211

```
  mocha:runner _addEventListener(): adding for event uncaughtException; 0 current listeners +8ms
  mocha:runner _addEventListener(): adding for event unhandledRejection; 0 current listeners +0ms
  mocha:runner run(): starting +2ms
  mocha:runner run(): emitting start +0ms

  mocha:runner run(): emitted start +1ms
  mocha:runner runSuite(): running  +0ms
  mocha:runner runSuite(): running Error 1 @error1: +1ms
Error 1 @error1 --
  mocha:runner _addEventListener(): adding for event error; 0 current listeners +1ms
    [1]  Starting recording promises
    Retries: 3
    Timeouts: 
 › [Session] Starting singleton browser session
  test issue
  mocha:runner _addEventListener(): adding for event error; 0 current listeners +1s
  mocha:runner _addEventListener(): adding for event error; 0 current listeners +1s
    I am on page "http://example.org"
    › [Browser:Error] Failed to load resource: the server responded with a status of 404 (Not Found)
    I wait for visible ".nothing", 1
    [1]  Error (Non-Terminated) | Error: element (.nothing) still not visible after 1 sec
locator.waitFor: Timeout 1000ms exceeded.
Call log:
  - waiting for locator('.nothing').first() to be visible
 | (err) => { step.status = 'failed'; step.endTime = ...
    [1] Error | Error: element (.nothing) still not visible after 1 sec
locator.waitFor: Timeout 1000ms exceeded.
Call log:
  - waiting for locator('.nothing').first() to be visible
 (e) => { const err = (recorder.getAsyncErr() === n...
    [1] <teardown>  Stopping recording promises
 › <screenshotOnFail> Test failed, try to save a screenshot
 › Screenshot is saving to /Users/t/Desktop/projects/codeceptjs-debug/output/test_issue.failed.png
  mocha:runner _addEventListener(): adding for event error; 0 current listeners +2s
    [2]  Starting recording promises
  mocha:runner _addEventListener(): adding for event error; 0 current listeners +35ms
  mocha:runner runSuite(): running Error 2 @error2: +2ms
  mocha:runner runSuite(): bailing +0ms
  mocha:runner runSuite(): running Error 3 @error3: +0ms
  mocha:runner runSuite(): bailing +0ms
  mocha:runner runSuite(): running Error 4 @error4: +0ms
[...]

Issue #4197, opened 2024-02-13
retryTo leads to stale process on failure


cjhille:
I'm sorry @kobenguyent, but we must have missed a regression with the last fix here https://github.com/codeceptjs/CodeceptJS/issues/4110

It works fine as long as assertions are a success, but as soon as there is a failure – even outside the retryTo block – the test never terminates.

Simple test code:
```js
Feature('foo');

Scenario('test issue', async ({ I }) => {
    I.amOnPage('http://example.org')
    I.waitForVisible('.nothing', 1); // should fail here but it won't terminate
    await retryTo( (tryNum) => {
        I.see(".doesNotMatter");
    }, 10);
});
```
EDIT: Adding an irregular await before the failing assertion `await I.waitForVisible('.nothing', 1)` will let the promise chain resolve.

<img width="955" alt="image" src="https://github.com/codeceptjs/CodeceptJS/assets/8192226/de370355-0de8-4993-b0a4-64bece67d78a">

cjhille:
The issue still persists (see test case). Unfortunately, `retryTo` can not be used currently. FYI @DavertMik

kobenguyent:
hey @cjhille I reverted those fixes at https://github.com/codeceptjs/CodeceptJS/pull/4117/files#diff-3799cca2cab187279c2a50bbd0c3a7451b24e20da2bdad08d42eb95444d392baR92 but the issue still persists, so I think it's coming from somewhere else.

![Screenshot 2024-05-29 at 07 51 15](https://github.com/codeceptjs/CodeceptJS/assets/7845001/a0f5acb7-ea26-457b-bf84-315b162ded9a)

kobenguyent:
per my initial thinking, I guess the issue is from `waitForVisible` 

![Screenshot 2024-05-29 at 07 57 23](https://github.com/codeceptjs/CodeceptJS/assets/7845001/29bcce41-2209-4adc-88e2-8bf2e621a080)

Horsty80:
@kobenguyent after investigation, we have found that the issue is not about `retryTo` but occurs when a promise reject.

This is enough to stale the process

```ts
Scenario('Rejected promise stale the process', async ({ I }) => {
  await new Promise((resolve, reject) => setTimeout(reject, 500));
})
```

We investigate how to solve this

PR #4377, opened 2024-06-04
fix: handle throw error inside retryTo promise

Horsty80:
- When a throw error occurs inside promise, that stale the process, by rewrite with good reject and done call, we prevent this to happen
- #4367 

- maybe this pr can help this issue #4358

Applicable plugins:

- [x] retryTo

## Type of change

- [x] :bug: Bug fix

## Checklist:

- [x] Tests have been added


Horsty80:
@kobenguyent with a friend, we have work on this, you can find a reproducible unit test that lead to a stale process without our change on `retryTo.js`

It's seems not terminating promises properly can lead to a stale process.
So maybe a clue to resolve issue #4358 

Let me know if this needs some reworks 

Thanks

kobenguyent:
Hey @Horsty80 thanks for the investigation. May you hell check the UTs? Failed tests there. 🤔

Horsty80:
> Hey @Horsty80 thanks for the investigation. May you hell check the UTs? Failed tests there. 🤔

I check the why it's failing, on my laptop the `npm run test` is working and correct that

Horsty80:
@kobenguyent TU are fix now :)

kobenguyent:
Could you do me a favor please?

Trying your fix with scenario listed here. https://github.com/codeceptjs/CodeceptJS/issues/4197

If that the case then I don't think my fix is needed or perhaps an extra layer to be sure.

Horsty80:
> Could you do me a favor please?
> 
> Trying your fix with scenario listed here. #4197
> 
> If that the case then I don't think my fix is needed or perhaps an extra layer to be sure.

I've check the scenario 

```ts
Scenario('test issue', async ({ I }) => {
    I.amOnPage('http://example.org')
    I.waitForVisible('.nothing', 1); // should fail here but it won't terminate
    await retryTo( (tryNum) => {
        I.see(".doesNotMatter");
    }, 10);
});
```

And I think is another issue, not on the retryTo 
I'm working on fix this to so I will fix it in another PR 

I've tested this scenario and the issue (without your fix) is still present.
Sequence I.waitForVisible that failed and after a retryTo lead to stale the process

kobenguyent:
> > Could you do me a favor please?
> > Trying your fix with scenario listed here. #4197
> > If that the case then I don't think my fix is needed or perhaps an extra layer to be sure.
> 
> I've check the scenario
> 
> ```ts
> Scenario('test issue', async ({ I }) => {
>     I.amOnPage('http://example.org')
>     I.waitForVisible('.nothing', 1); // should fail here but it won't terminate
>     await retryTo( (tryNum) => {
>         I.see(".doesNotMatter");
>     }, 10);
> });
> ```
> 
> And I think is another issue, not on the retryTo I'm working on fix this to so I will fix it in another PR
> 
> I've tested this scenario and the issue (without your fix) is still present. Sequence I.waitForVisible that failed and after a retryTo lead to stale the process

Thanks @Horsty80. I guess that aforementioned covered by https://github.com/codeceptjs/CodeceptJS/pull/4367

Horsty80:
> > > Could you do me a favor please?
> > > Trying your fix with scenario listed here. #4197
> > > If that the case then I don't think my fix is needed or perhaps an extra layer to be sure.
> > 
> > 
> > I've check the scenario
> > ```ts
> > Scenario('test issue', async ({ I }) => {
> >     I.amOnPage('http://example.org')
> >     I.waitForVisible('.nothing', 1); // should fail here but it won't terminate
> >     await retryTo( (tryNum) => {
> >         I.see(".doesNotMatter");
> >     }, 10);
> > });
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And I think is another issue, not on the retryTo I'm working on fix this to so I will fix it in another PR
> > I've tested this scenario and the issue (without your fix) is still present. Sequence I.waitForVisible that failed and after a retryTo lead to stale the process
> 
> Thanks @Horsty80. I guess that aforementioned covered by #4367

Indeed maybe your fix cover this case.

kobenguyent (approved):

Arhell (approved):

kobenguyent:
as this also addresses the stale process, so I think we would combine them into a [branch](https://github.com/codeceptjs/CodeceptJS/pull/4367). Then provide a beta version to test it out.

Issue #3640, opened 2023-04-14
Test pipeline stuck in "after all" hooks steps


Horsty80:
I run multiple test scenario in a jenkins pipeline
When all test are finished to run, i get stuck in this step until my configured timeout trigger the end of the pipeline

```bash
-- FAILURES:

  ✖ "after all" hook: codeceptjs.afterSuite in "MyScenario" in 10ms
```

It's occurs randomly and by waves. Always at the end of the codeceptjs script, i presume in the teardown / cleanup step.
It's look that occurs only on jenkins, when i run test locally everything works.

This is the last log i've got from codecetp, after that, it's timeout from jenkins
```bash

... some codeceptjs steps

 --- ENDED "before each" hook: Before for "Test scenario @smoketest" ---

... some codeceptjs steps
... last instructions
  ✔ OK in 1806ms

    --- STARTED "after each" hook: After for "Test scenario @smoketest" ---
    I delete fake user "fake-user.e2e@mail.com"
    I delete fake workspace "fake-user.e2e@mail.com"
    I delete fake user preferences "fake-user.e2e@mail.com"
    --- ENDED "after each" hook: After for "Test scenario @smoketest" ---

-- FAILURES:

  ✖ "after all" hook: codeceptjs.afterSuite in "Test Feature" in 6ms

Cancelling nested steps due to timeout
Sending interrupt signal to process
10:22:30.290166 common.go:40: exit status 1
script returned exit code 1
```

I don't know how to debug this behavior, it's very annoying because all test in the pipeline are down in few minutes, but get stuck 
Can someone help me to debug this strange behavior ?
Thanks

### Details

* CodeceptJS version: 3.3.7 
* NodeJS Version: 18.15.0
* Operating System: Linux & Windows same behavior
* Playwright: 1.32.1

```js
exports.config = {
  tests: ['./tests/**/*_test.ts', './tests/**/test.ts'],
  output: './output',
  helpers: {
    Playwright: {
      url: config.base_url,
      show: true,
      trace: true,
      browser: 'chromium',
      chromium: {
        args: ['--auth-server-whitelist=""'],
      },
      ignoreHTTPSErrors: true,
      fullPageScreenshots: true,
      waitForAction: 250,
      waitForTimeout: 10000,
      waitForNavigation: 'load',
      restart: 'context',
      keepCookies: false,
      keepBrowserState: false,
      basicAuth: {
        username: defaultUser.email,
        password: defaultUser.password,
      },
    },
[...]

PR #4378, opened 2024-06-05
Ensure scenario termination

### LINKED ISSUE #4358: [Asking for help] Process are blocking when throw error in scenario with Retry enabled

Horsty80:
- The CLI was staling on some cases (promise and throw error). We manage to reproduce this behavior. Added some text and fixe this issue inside `scenario.js` code
- Resolves #4358 , #4197 , #3640, #4378

Also, we have updated `done()` to prevent it from being called more than once inside Scenario.
Possibly the root of this kind of error.
`Error: done() called multiple times in `

Applicable helpers:

- [x] Playwright

## Type of change

- [x] :bug: Bug fix
- [x] 🧹 Chore
- [x] :nail_care: Polish code

## Checklist:

- [x] Tests have been added

Capture d’écran 2024-06-05 à 16 44 50


Horsty80:
Hey @kobenguyent I've find and fix that cause the issue #4197  that you talk about here 
 https://github.com/codeceptjs/CodeceptJS/pull/4377#issuecomment-2149360983

I have added the test case here and it's ok

https://github.com/codeceptjs/CodeceptJS/pull/4378/files#diff-1d3c534eda54e8e8ac941eb8fcce76dad6eb240a5def22158aea06cf097a5135R20

lib/plugin/retryTo.js
```diff
@@ -113,6 +113,9 @@ module.exports = function (config) {
 
       recorder.add('retryTo', async () => {
         tryBlock();
+      }).catch(err => {
+        console.error('An error occurred:', err);
+        done(null);
```
kobenguyent, line 119: I'm curious if done() should be enough?

lib/scenario.js
```diff
@@ -1,9 +1,9 @@
...
-const event = require('./event');
-const recorder = require('./recorder');
-const assertThrown = require('./assert/throws');
-const { ucfirst, isAsyncFunction } = require('./utils');
-const parser = require('./parser');
+const promiseRetry = require("promise-retry");
+const event = require("./event");
+const recorder = require("./recorder");
+const assertThrown = require("./assert/throws");
+const { ucfirst, isAsyncFunction } = require("./utils");
```
kobenguyent: Could we revert to single quotes like it used to be?

test/runner/scenario_stale_test.js
```diff
@@ -0,0 +1,23 @@
+const { expect } = require('expect');
+const path = require('path');
+const exec = require('child_process').exec;
```
kobenguyent: Could we also use const { exec } here?

lib/scenario.js
Horsty80: Yes sur ! Just it's could be fine if the linter will be fixed because when i run the lint command, i get an error due to eslint file.
But yes i will push with the good formatting

lib/plugin/retryTo.js
Horsty80: the test case that i added passed with the resolve promise so i guess it's enough ^^

Just `retryTo` will be update with my other pr, so when the first it's merge, i will update this file
I think it's a good practice to add the `catch` block when we add something to the recorder i presume

lib/scenario.js
kobenguyent: perhaps you could wait for this #4379 to run the lint command to fix the lint issue.

Horsty80: I still have issue when commit 

```
> git -c user.useConfigOnly=true commit --quiet --allow-empty-message --file -

> codeceptjs@3.6.2 lint
> eslint bin/ examples/ lib/ test/ translations/ runok.js

✖ 4079 problems (4079 errors, 0 warnings)

husky - pre-commit hook exited with code 1 (error)
```

kobenguyent:
as this also addresses the stale process, so I think we would combine them into a [branch](https://github.com/codeceptjs/CodeceptJS/pull/4367). Then provide a beta version to test it out.

lib/plugin/retryTo.js
kobenguyent: ah I meant `done()` instead of `done(null)`

kobenguyent:
hi @Horsty80 would you mind checking the failed tests? I changed the base branch there are some conflicts, I tried to resolve them but some failed tests.

Horsty80:
> hi @Horsty80 would you mind checking the failed tests? I changed the base branch there are some conflicts, I tried to resolve them but some failed tests.

Yes of course ! I know that will be conflict with my other PR ^^
I will fix that and push changes today

lib/plugin/retryTo.js
Horsty80: > ah I meant `done()` instead of `done(null)`

The done is the `resolve` of the promise and I get this type 

Capture d’écran 2024-06-07 à 10 13 01

and in the file, done function already call with `done(null)` 🤷‍♂️

PR #4476, opened 2024-09-02
feat(cli): print failed hooks

### LINKED ISSUE #4466: Error handling suggestion for `AfterSuite` Hook
!!! This is not a bug, just a suggestion !!!

Suppose I have 2 scenarios and 1 `AfterSuite` hook in a file and while execution both the scenarios passed but the `AfterSuite` hook failed, it shows as `FAIL  | 2 passed, 1 failed`. This creates confusion, it may better show as 2 passed and the `AfterSuite` hook failed.

kobenguyent:
- Resolves #4466 

run command
Screenshot 2024-09-02 at 15 25 20

run workers command
Screenshot 2024-09-02 at 15 24 53

## Type of change
- [ ] :bug: Bug fix

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -145,6 +145,7 @@ class Cli extends Base {
 
   result() {
     const stats = this.stats;
+    stats.failedHooks = 0;
     console.log();
 
     // passes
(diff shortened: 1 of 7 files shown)


kobenguyent:
Successfully published 3.6.6-beta.5

mirao:
@kobenguyent 
I have suspicion that this bug fix caused that if there are failed scenarios and also failed hooks, then the failed scenarios aren't reported in total count of failures :bug: 

E.g.
Here 36 failures should be reported (because 36 scenarios failed), but we are getting this (Mochawesome HTML report)

image

`[1.dev:chromium]   FAIL  | 191 passed, 4 failedHooks, 44 skipped   // 2h`

Tested with CodeceptJS 3.6.7

lib/cli.js
```diff
@@ -216,8 +217,14 @@ class Cli extends Base {
       console.log();
     }
 
+    this.failures.forEach((failure) => {
+      if (failure.constructor.name === 'Hook') {
+        stats.failures -= stats.failures
```
kobenguyent, line 222: hey @mirao perhaps the failure count should not be touched, just adding `failedHooks` is good enough. What do you think?

mirao: Yes, it makes sense.

mirao: `stats.failures -= stats.failures` isn't the same as `stats.failures = 0` ?

PR #4557, opened 2024-11-11
fix(playwright): Different behavior of see* and waitFor* when used in within

### LINKED ISSUE #4547: Different behavior of see* and waitFor* when used in within
I've been trying to use within in some of my test cases and noticed that its behavior isn't consistent across different assertions. I'm testing this by using an element that is present on page but should fail inside the context given to within. Tested on both Puppeteer or Playwright.

#### What do you get instead?
I'm trying to verify that an element is not present inside the given context. `dontSeeElement` correctly succeeds as it only searches inside the given context but `waitForInvisible` fails because it always uses the full page as a context.

I've looked through the source code to try and debug this and it seems like the issue is in `_getContext` method. 

`waitFor*` assertions (at least those that I've looked at) determine the searchable context by calling [_getContext](https://github.com/codeceptjs/CodeceptJS/blob/0275acb9b99a582b785829f362225ed9461e6a96/lib/helper/Puppeteer.js#L2307C60-L2307C68) which only returns `this.context` if it points to a `FrameLocator`(Playwright) or `CdpFrame` (Puppeteer), otherwise it defaults to the whole page. 

`(dont)see*` assertions on the other hand use `_locate` which uses `this.context` that gets set inside `_withinBegin`.

I am aware that documentation suggests [refactoring without using within](https://codecept.io/basics/#within) in case of unexpected results but given those snippets I would expect it to work on regular locators too, not just limited to frames. At the very least if within is expected to only work with frames I think this should be noted in the docs. What's the proper behavior?

```bash
within --
  testing within
    I am on page "https://codecept.io/"
    I wait for visible ".frameworks"
    I wait for visible "[alt="React"]"
    I wait for visible ".mountains"
    Within ".mountains" ""
      I dont see element "[alt="React"]"
      I wait for invisible "[alt="React"]", 2
  ✖ FAILED in 3896ms


-- FAILURES:

  1) within
       testing within:
     Error: element ([alt="React"]) still visible after 2 sec
Waiting for selector `[alt="React"]` failed: Waiting failed: 2000ms exceeded
```

```js
Feature('within');

Scenario('testing within',  ({ I }) => {
    I.amOnPage('https://codecept.io/')
    I.waitForVisible('.frameworks')
    I.waitForVisible('[alt="React"]')
    I.waitForVisible('.mountains')
    within('.mountains', () => {
        I.dontSeeElement('[alt="React"]')
        I.waitForInvisible('[alt="React"]', 2)
    })
});

```

### Details

* CodeceptJS version: 3.6.7
* NodeJS Version: v20.10.0
* Operating System: macOS Sonoma 14.7 
* puppeteer || webdriverio || testcafe version (if related): Puppeteer 21.11.0 and Playwright 1.48.2
* Configuration file: (default config that's generated by `codeceptjs init` command)

```js
const { setHeadlessWhen, setCommonPlugins } = require('@codeceptjs/configure');

setCommonPlugins();

/** @type {CodeceptJS.MainConfig} */
exports.config = {
  tests: './*_test.js',
  output: './output',
  helpers: {
    Playwright: {
      browser: 'chromium',
      url: 'https://codecept.io',
      show: true,
      waitForTimeout: 10000,
    }
  },
  include: {
    I: './steps_file.js'
  },
  name: 'withintest'
}
```

kobenguyent:
- Resolves #4547

Applicable helpers:

- [ ] Playwright
- [ ] Puppeteer
- [ ] WebDriver
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] TestCafe

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] stepTimeout
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [ ] :bug: Bug fix
- [ ] 🧹 Chore
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/helper/Playwright.js
+++ b/lib/helper/Playwright.js
@@ -900,7 +900,7 @@ class Playwright extends Helper {
   }
 
   async _evaluateHandeInContext(...args) {
-    const context = await this._getContext()
+    const context = (await this.context) || (await this._getContext())
     return context.evaluateHandle(...args)
   }
 
@@ -1327,7 +1327,7 @@ class Playwright extends Helper {
    * ```
    */
   async _locateClickable(locator) {
-    const context = await this._getContext()
+    const context = (await this.context) || (await this._getContext())
     return findClickable.call(this, context, locator)
   }
 
@@ -2478,7 +2478,7 @@ class Playwright extends Helper {
     locator = new Locator(locator, 'css')
 
     let waiter
-    const context = await this._getContext()
+    const context = (await this.context) || (await this._getContext())
     if (!locator.isXPath()) {
       const valueFn = function ([locator]) {
         return Array.from(document.querySelectorAll(locator)).filter((el) => !el.disabled).length > 0
@@ -2506,7 +2506,7 @@ class Playwright extends Helper {
     locator = new Locator(locator, 'css')
 
     let waiter
-    const context = await this._getContext()
+    const context = (await this.context) || (await this._getContext())
     if (!locator.isXPath()) {
       const valueFn = function ([locator]) {
         return Array.from(document.querySelectorAll(locator)).filter((el) => el.disabled).length > 0
@@ -2534,7 +2534,7 @@ class Playwright extends Helper {
     const locator = new Locator(field, 'css')
     const matcher = await this.context
     let waiter
-    const context = await this._getContext()
+    const context = (await this.context) || (await this._getContext())
     if (!locator.isXPath()) {
       const valueFn = function ([locator, value]) {
         return (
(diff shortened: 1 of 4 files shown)


lib/helper/Playwright.js
```diff
@@ -900,7 +900,7 @@ class Playwright extends Helper {
   }
 
   async _evaluateHandeInContext(...args) {
-    const context = await this._getContext()
+    const context = (await this.context) || (await this._getContext())
```
DavertMik: can we just upate `this._getContext` for this?

kobenguyent: Updated

PR #4650, opened 2024-12-17
fix: all contributors

kobenguyent:
- remove contributor-faces as it is not updated actively and causing a lot of security issues.

## Type of change
- [ ] :bug: Bug fix


kobenguyent:
@thomashohn here you go

PR #4727, opened 2025-01-08
Check command

DavertMik:
Created `check` command to quickly test config, tests, and that browser can be started and finished

This can be done locally or on CI before running tests

image

### DIFF
--- a/lib/command/check.js
+++ b/lib/command/check.js
@@ -1,0 +1,174 @@
+const { getConfig, getTestRoot } = require('./utils')
+const Codecept = require('../codecept')
+const output = require('../output')
+const standardActingHelpers = require('../plugin/standardActingHelpers')
+const store = require('../store')
+const container = require('../container')
+const figures = require('figures')
+const chalk = require('chalk')
+const { createTest } = require('../mocha/test')
+const { getMachineInfo } = require('./info')
+const definitions = require('./definitions')
+
+module.exports = async function (test, options) {
+  if (options.grep) process.env.grep = options.grep
+  const configFile = options.config
+
+  setTimeout(() => {
+    output.error("Something went wrong. Checks didn't pass and timed out. Please check your config and helpers.")
+    process.exit(1)
+  }, options.timeout || 50000)
+
+  const checks = {
+    config: false,
+    container: false,
+    pageObjects: false,
+    helpers: false,
+    setup: false,
+    tests: false,
+    def: false,
+  }
+
+  const testRoot = getTestRoot(configFile)
+  let config = getConfig(configFile)
+
+  try {
+    config = getConfig(configFile)
+    checks['config'] = true
+  } catch (err) {
+    checks['config'] = err
+  }
+
+  printCheck('config', checks['config'], config.name)
+
+  let codecept
+  try {
+    codecept = new Codecept(config, options)
+    codecept.init(testRoot)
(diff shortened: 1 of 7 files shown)


kobenguyent:
@DavertMik what do you think if running all checks in parallel using Promise.all?
here is my local run elapsed around 1.81s compared to in sequence run around 2.3s 

```
✔ CONFIG acceptance
✔ CONTAINER 
✔ BOOTSTRAP 
✔ TESTS Total: 47 tests
✖ HELPERS helper._init is not a function
✔ PAGE OBJECTS Total: 0 support objects
✖ HELPERS HOOKS helper._beforeSuite is not a function
✔ TYPESCRIPT DEFINITIONS 
Something went wrong. Checks didn't pass. (Total Elapsed: 1.81s)
```

lib/command/check.js
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  printCheck('TypeScript Definitions', checks['def'])
+
+  output.print('')
+
+  if (!Object.values(checks).every(check => check === true)) {
+    output.error("Something went wrong. Checks didn't pass.")
+    output.print()
+    getMachineInfo()
```
kobenguyent: missing await here I think.

package.json
```diff
@@ -93,7 +93,6 @@
     "fn-args": "4.0.0",
     "fs-extra": "11.2.0",
     "glob": "^11.0.0",
-    "fuse.js": "^7.0.0",
```
kobenguyent: got error if this lib removed

DavertMik: ouch, this is bad merge, right

DavertMik:
@kobenguyent good point for using Promise.all
but I think to keep things simple and traceable better to do things one by one
otherwise I think it won't be stable

lib/command/check.js
DavertMik: yes! thank you

DavertMik:
@kobenguyent btw, it's only 1s on our workflow 

image

https://github.com/codeceptjs/CodeceptJS/actions/runs/12682764702/job/35348708849?pr=4727

kobenguyent:
@DavertMik I have two more things:
- shall we add the option to specify the configuration as in the real world they would be different configurations in projects?
- I'm curious if calling the check by default when initializing run or run-workers is bringing value?

kobenguyent:
@DavertMik one more thing:
- regarding the typescript definitions, is it possible that we read the conf file, if it's .ts then we add this check, otherwise, we shall skip this check?

lib/command/check.js
```diff
@@ -0,0 +1,173 @@
...
+  let config = getConfig(configFile)
+
+  try {
+    config = getConfig(configFile)
+    checks['config'] = true
+  } catch (err) {
+    checks['config'] = err
+  }
+
+  printCheck('config', checks['config'], config.name)
```
kobenguyent, line 41: how about `config name` here?

DavertMik:
@kobenguyent 

> shall we add the option to specify the configuration as in the real world they would be different configurations in projects?

yes, I added it as `-c --config`

> I'm curious if calling the check by default when initializing run or run-workers is bringing value?

For `run` - no, as you probably use it for development

For `run-workers` - may be

But anyway, I think `check` to be CI command that you can have as a precondition step

PR #4751, opened 2025-01-17
fix: handle logging step

kobenguyent:
- export step as public API so that it could be used in the custom helper, for instance our helper https://github.com/codeceptjs/expect-helper/blob/main/index.js#L51

```
const  Step = require('codeceptjs')
const step = new Step('I', `expect "${JSON.stringify(actualValue)}" to not equal "${JSON.stringify(expectedValue)}"`);
output.step(step)
```

- handle the case which the step is a string

```
    [1]  Error (Non-Terminated) | TypeError: step.toCliStyled is not a function | err => { step.status = 'failed' step.endTime = Dat...
    [1] Error | TypeError: step.toCliStyled is not a function undefined...
    [1] <teardown>  Stopping recording promises
    [2]  Starting recording promises

```

## Type of change

- [ ] :fire: Breaking changes
- [ ] :bug: Bug fix
- [ ] 🧹 Chore
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/output.js
+++ b/lib/output.js
@@ -1,6 +1,7 @@
 const colors = require('chalk')
 const figures = require('figures')
 const { maskSensitiveData } = require('invisi-data')
+const Step = require('./step')
 
 const styles = {
   error: colors.bgRed.white.bold,
@@ -115,7 +116,7 @@ module.exports = {
       }
     }
 
-    try {
+    if (step instanceof Step) {
       let stepLine = step.toCliStyled()
       if (step.metaStep && outputLevel >= 1) {
         // this.stepShift += 2;
@@ -127,9 +128,11 @@ module.exports = {
 
       const _stepLine = isMaskedData() ? maskSensitiveData(stepLine) : stepLine
       print(' '.repeat(this.stepShift), truncate(_stepLine, this.spaceShift))
-    } catch (e) {
+    } else if (typeof step === 'string') {
       // sometimes the step is just the string, so we basically print it out
       print(step)
+    } else {
+      console.log(`Something went wrong with ${step}`)
     }
   },
 
--- a/lib/output.js
+++ b/lib/output.js
@@ -115,17 +115,22 @@ module.exports = {
       }
     }
 
-    let stepLine = step.toCliStyled()
-    if (step.metaStep && outputLevel >= 1) {
-      // this.stepShift += 2;
-      stepLine = colors.dim(truncate(stepLine, this.spaceShift))
-    }
-    if (step.comment) {
-      stepLine += colors.grey(step.comment.split('\n').join('\n' + ' '.repeat(4)))
-    }
+    try {
+      let stepLine = step.toCliStyled()
+      if (step.metaStep && outputLevel >= 1) {
+        // this.stepShift += 2;
+        stepLine = colors.dim(truncate(stepLine, this.spaceShift))
+      }
+      if (step.comment) {
+        stepLine += colors.grey(step.comment.split('\n').join('\n' + ' '.repeat(4)))
+      }
 
-    const _stepLine = isMaskedData() ? maskSensitiveData(stepLine) : stepLine
-    print(' '.repeat(this.stepShift), truncate(_stepLine, this.spaceShift))
+      const _stepLine = isMaskedData() ? maskSensitiveData(stepLine) : stepLine
+      print(' '.repeat(this.stepShift), truncate(_stepLine, this.spaceShift))
+    } catch (e) {
+      // sometimes the step is just the string, so we basically print it out
+      print(step)
+    }
   },
 
   /** @namespace */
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -33,7 +33,8 @@
     ".": "./lib/index.js",
     "./els": "./lib/els.js",
     "./effects": "./lib/effects.js",
-    "./steps": "./lib/steps.js"
+    "./steps": "./lib/steps.js",
+    "./step": "./lib/step.js"
```
DavertMik: how about adding step export to `index.js`?

```
const { step } = require('codeceptjs')
```

this is how we access all internal API 

`codeceptjs/steps` is userland API which is used inside tests, not custom helpers, or plugins

DavertMik (changes requested):

package.json
kobenguyent: Sounds better @DavertMik

lib/output.js
```diff
@@ -115,17 +115,22 @@ module.exports = {
...
 
-    let stepLine = step.toCliStyled()
-    if (step.metaStep && outputLevel >= 1) {
-      // this.stepShift += 2;
-      stepLine = colors.dim(truncate(stepLine, this.spaceShift))
-    }
-    if (step.comment) {
-      stepLine += colors.grey(step.comment.split('\n').join('\n' + ' '.repeat(4)))
-    }
+    try {
```
DavertMik: maybe better would be:

```js
if (step instanceof Step) {
  // ...
} elseif (typeof step == 'string') {
} else {
   // incorrect step
}
```

try catch is not the best approach here if we know where error can happen and how to avoid it

DavertMik (changes requested):

DavertMik:
Something happened with tests in this :cry:

DavertMik:
Hey @kobenguyent , I have a better idea for this case

How about create a Step that can execute arbitrary function

See my implementation in #4781 

https://github.com/codeceptjs/CodeceptJS/pull/4781/files#diff-175fae64a79f8903d1998c3068b43193025a23c36cecc50031fc7009e19b0460


here is what you can do:

```js
const  FuncStep = require('codeceptjs/lib/steps/func')
const step = new Step(`expect "${JSON.stringify(actualValue)}" to not equal "${JSON.stringify(expectedValue)}"`);
step.setCallable(...) // actual assertion happens here?
output.step(step)
```

you can actually add this step to promise chain

```js
const recordStep = require('codeceptjs/lib/steps/record')
const  FuncStep = require('codeceptjs/lib/steps/func')
const step = new Step(`expect "${JSON.stringify(actualValue)}" to not equal "${JSON.stringify(expectedValue)}"`);
step.setCallable(...) // actual assertion happens here?
recordStep(step)
```

In this case, all steps should are actual Steps
if needed more we can extend from `BaseStep` to create customized steps

so no need to pass strings

kobenguyent:
> Hey @kobenguyent , I have a better idea for this case
> 
> 
> 
> How about create a Step that can execute arbitrary function
> 
> 
> 
> See my implementation in #4781 
> 
> 
> 
> https://github.com/codeceptjs/CodeceptJS/pull/4781/files#diff-175fae64a79f8903d1998c3068b43193025a23c36cecc50031fc7009e19b0460
> 
> 
> 
> 
> 
> here is what you can do:
> 
> 
> 
> ```js
> 
> const  FuncStep = require('codeceptjs/lib/steps/func')
> 
> const step = new Step(`expect "${JSON.stringify(actualValue)}" to not equal "${JSON.stringify(expectedValue)}"`);
> 
> step.setCallable(...) // actual assertion happens here?
> 
> output.step(step)
> 
> ```
> 
> 
> 
> you can actually add this step to promise chain
> 
> 
> 
> ```js
> 
> const recordStep = require('codeceptjs/lib/steps/record')
> 
> const  FuncStep = require('codeceptjs/lib/steps/func')
> 
> const step = new Step(`expect "${JSON.stringify(actualValue)}" to not equal "${JSON.stringify(expectedValue)}"`);
> 
> step.setCallable(...) // actual assertion happens here?
> 
> recordStep(step)
> 
> ```
> 
> 
> 
> In this case, all steps should are actual Steps
> 
> if needed more we can extend from `BaseStep` to create customized steps
> 
> 
> 
> so no need to pass strings

Hi @DavertMik thanks for sharing this. So still do we need this PR just to handle the custom helpers for instance that are not yet adapted with new implementation?

DavertMik:
@kobenguyent if you think it is needed for backwards compatibility with Expect helper I'm ok
but in 4.0 this should be dropped

kobenguyent:
> @kobenguyent if you think it is needed for backwards compatibility with Expect helper I'm ok
> 
> but in 4.0 this should be dropped

Oh, if this shall be dropped in 4.0 then I'd say let's exclude this.

PR #4775, opened 2025-01-22
fix: glob backward compatibility

kobenguyent:

Applicable helpers:

- [ ] Playwright
- [ ] Puppeteer
- [ ] WebDriver
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] TestCafe

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] stepTimeout
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [ ] :bug: Bug fix
- [ ] 🧹 Chore
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added


kobenguyent:
published as codeceptjs@3.7.0-beta.7 for testing.

DavertMik:
I'm not sure if we use it correctly:

image

In 10-11 our `glob = require('glob')` probably won't work

https://github.com/isaacs/node-glob/blob/main/changelog.md#1000

package.json
```diff
@@ -94,7 +94,7 @@
     "figures": "3.2.0",
     "fn-args": "4.0.0",
     "fs-extra": "11.2.0",
-    "glob": "^11.0.1",
+    "glob": ">=9.0.0",
```
DavertMik:
```suggestion
    "glob": ">=9.0.0|<12",
```

we should not install versions which are not tested

DavertMik (changes requested):
I didn't test this notation, but we need to ensure we use compatible glob version

Also I'm concerned that we can actually use glob from v11 because of no named exports

DavertMik:
@kobenguyent please change the notation where glob is used

so instead of 

```js
const glob = require('glob')
```

we would have

```js
const { globSync }  = require('glob')
```

as it is recommended now in docs

I checked it works on 9 and 11 version

PR #4782, opened 2025-01-25
better hooks handling by submitting where test.failed occur

DavertMik:
* Minor debug improvements to Playwright
* Added hookInfo to event if test failed in a hook
* fixed screenshot and stepByStep plugin to work in Before/After hooks

### DIFF
--- a/lib/plugin/screenshotOnFail.js
+++ b/lib/plugin/screenshotOnFail.js
@@ -72,11 +72,12 @@ module.exports = function (config) {
     return
   }
 
-  event.dispatcher.on(event.test.failed, test => {
-    if (test.ctx?._runnable.title.includes('hook: ')) {
-      output.plugin('screenshotOnFail', 'BeforeSuite/AfterSuite do not have any access to the browser, hence it could not take screenshot.')
+  event.dispatcher.on(event.test.failed, (test, _err, hookName) => {
+    if (hookName == 'BeforeSuite' || hookName == 'AfterSuite') {
+      // no browser here
       return
     }
+
     recorder.add(
       'screenshot of failed test',
       async () => {
--- a/lib/plugin/screenshotOnFail.js
+++ b/lib/plugin/screenshotOnFail.js
@@ -64,7 +64,7 @@ module.exports = function (config) {
   }
 
   if (Codeceptjs.container.mocha()) {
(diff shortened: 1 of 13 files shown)


lib/plugin/screenshotOnFail.js
```diff
@@ -72,11 +72,12 @@ module.exports = function (config) {
     return
   }
 
-  event.dispatcher.on(event.test.failed, test => {
-    if (test.ctx?._runnable.title.includes('hook: ')) {
-      output.plugin('screenshotOnFail', 'BeforeSuite/AfterSuite do not have any access to the browser, hence it could not take screenshot.')
+  event.dispatcher.on(event.test.failed, (test, _err, hookName) => {
+    if (hookName == 'BeforeSuite' || hookName == 'AfterSuite') {
```
kobenguyent: I guess it shall be ===, shall not it?

kobenguyent: Shall we also print something to acknowledge users that there is no browser available at this stage

lib/plugin/stepByStepReport.js
```diff
@@ -121,12 +121,13 @@ module.exports = function (config) {
     deleteDir(dir)
   })
 
-  event.dispatcher.on(event.test.failed, (test, err) => {
-    if (test.ctx._runnable.title.includes('hook: ')) {
-      output.plugin('stepByStepReport', 'BeforeSuite/AfterSuite do not have any access to the browser, hence it could not take screenshot.')
+  event.dispatcher.on(event.test.failed, (test, _err, hookName) => {
+    if (hookName == 'BeforeSuite' || hookName == 'AfterSuite') {
```
kobenguyent: Same aforementioned question

lib/plugin/screenshotOnFail.js
DavertMik: it doesn't matter but yes, can be

DavertMik: no, I think this information is not important

it would be valuable if user  does some actions in a browser and wants screenshots to be stored
it would be confusing in this case to see that some screenshots are missing 
(btw it is current behavior)

but as they can't do actions in AfterSuite, they don't expect screenshots to be added

PR #4783, opened 2025-01-27
Feat/auth plugin

DavertMik:
* renamed `autoLogin` plugin to `auth` plugin
* wrapped login steps into a meta step
* added Comment step
* added support to use `loginAs` function to work from `BeforeSuite`

```js
BeforeSuite({ loginAs } => loginAs('user0');
```

In this case cookies can be preloaded into Playwright when creating new browser session for faster access


lib/plugin/auth.js
```diff
@@ -0,0 +1,430 @@
...
+const path = require('path')
+const { fileExists } = require('../utils')
+const CommentStep = require('../step/comment')
+const Section = require('../step/section')
+const container = require('../container')
+const store = require('../store')
+const event = require('../event')
+const recorder = require('../recorder')
+const { debug } = require('../output')
+const isAsyncFunction = require('../utils').isAsyncFunction
```
kobenguyent: const { isAsyncFunction } = require('../utils')

just a code style I guess to make everything look consistant.

lib/plugin/auth.js
```diff
@@ -0,0 +1,430 @@
...
+    }
+
+    const section = new Section(`I am logged in as ${name}`)
+
+    if (config.saveToFile && !store[`${name}_session`]) {
+      loadCookiesFromFile(config)
+    }
+
+    if (isPlaywrightSession() && test?.opts?.cookies) {
+      if (test.opts.user == name) {
```
kobenguyent, line 294: - Double equals (==) will perform a type conversion when comparing two things, and will handle NaN, -0, and +0 specially to conform to IEEE 754 (so NaN != NaN, and -0 == +0);
- Triple equals (===) will do the same comparison as double equals (including the special handling for NaN, -0, and +0) but without type conversion; if the types differ, false is returned.

i'm not sure if we would try to do strict equality

lib/plugin/auth.js
```diff
@@ -0,0 +1,430 @@
...
+    suite.eachTest(test => {
+      // preload from store
+      if (store[`${name}_session`]) {
+        test.opts.cookies = store[`${name}_session`]
+        test.opts.user = name
+        return
+      }
+
+      if (!config.saveToFile) return
+      const cookieFile = path.join(global.output_dir, `${name}_session.json`)
```
kobenguyent, line 380: may you think we shall store this `output_dir` in `Store`?

lib/plugin/auth.js
```diff
@@ -0,0 +1,430 @@
...
+        ...config.users[u],
+      }),
+  )
+
+  if (config.saveToFile) {
+    // loading from file
+    loadCookiesFromFile(config)
+  }
+
+  const loginFunction = async name => {
```
kobenguyent, line 271: I could be wrong but what if user just puts some name that is not defined in config, shall we handle that case?

lib/plugin/auth.js
DavertMik: I don't think this matters for 2 strings....

lib/plugin/auth.js
DavertMik: yes, I was thinking about it... Probably we need to switch to this
but it's too much of refactoring to do

lib/plugin/auth.js
DavertMik: Yes, you are right, this should be handled

PR #4784, opened 2025-01-27
fix: allow gherkin keywords as supported by cucumber

ebo-zig:

Because managing translations of Gherkin keywords happens in this repo, the features cannot be run in every language. When setting the language of the feature to Dutch (nl), CodeceptJS threats it as English.

Therefore, the translations of keywords should be loaded from Cucumber: https://github.com/cucumber/gherkin/blob/main/javascript/src/gherkin-languages.json

## Type of change

- [x] :bug: Bug fix

## Checklist:

- [x] Tests have been added


kobenguyent:
hi @ebo-zig actually, we are not only translating gherkin keywords, but also the codeceptjs actions are translated as well, so that's why there is now separated file for each language, for instance https://github.com/codeceptjs/CodeceptJS/blob/3.x/translations/de-DE.js

translations/nl-NL.js
```diff
@@ -0,0 +1,78 @@
...
+const langCode = 'nl'
+
+module.exports = {
+  I: 'Ik',
+  contexts: {
+    Feature: gherkinTranslations[langCode].feature[0],
+    Scenario: gherkinTranslations[langCode].scenario[0],
+    ScenarioOutline: gherkinTranslations[langCode].scenarioOutline[0],
+  },
+  actions: {
```
kobenguyent, line 9: @ebo-zig may you help review this translation?

ebo-zig:
Hi @kobenguyent, I agree with the fact that not only Gherkin keywords should be translated.

But re-defining the Gherkin translations feels a little bit duplicate to me. So, if that is what we want, then that is allright.

translations/nl-NL.js
ebo-zig: Done! Came to the conclusion that the Dutch language is a little bit strange after reviewing this 😄

kobenguyent: Haha, yeah, I could imagine that as I asked help from chatgpt to translate them. Thanks for the review!

kobenguyent:
> Hi @kobenguyent, I agree with the fact that not only Gherkin keywords should be translated.
> 
> 
> 
> But re-defining the Gherkin translations feels a little bit duplicate to me. So, if that is what we want, then that is allright.

May you happen to have any better approaches for this?

ebo-zig:
I think there are two possible solutions:
1. Fully support all languages, which means that you proposal is the way to go.
2. Use the keywords of Gherkin as managed by Cucumber, as I am proposing.

The con of 1 is that using BDD tests in CodeceptJS doesn't work out of the box. And we are defining what already is defined by Cucumber.
The con of 2 is that you can use a language in BDD but using the same language in acceptance tests isn't always possible.

So there is no better approach IMO, other than chosing between the possible solutions.

kobenguyent:
> I think there are two possible solutions:
> 
> 1. Fully support all languages, which means that you proposal is the way to go.
> 2. Use the keywords of Gherkin as managed by Cucumber, as I am proposing.
> 
> The con of 1 is that using BDD tests in CodeceptJS doesn't work out of the box. And we are defining what already is defined by Cucumber. The con of 2 is that you can use a language in BDD but using the same language in acceptance tests isn't always possible.
> 
> So there is no better approach IMO, other than chosing between the possible solutions.

thanks for your input!

DavertMik:
Thank you! This looks really cool
I will check this tomorrow

DavertMik (approved):

ebo-zig:
@kobenguyent is there any chance this change will be available soon in a new tag?

kobenguyent:
@ebo-zig I could not say for certain when there would be a release, but of course, you could try `codeceptjs@3.7.0-beta.12` with latest changes.

PR #4785, opened 2025-01-28
injecting test & siute to container

DavertMik:
* injected suite and test into DI

```js

Feature('store-test-and-suite suite')

BeforeSuite(({ suite }) => {
})

Before(({ test }) => {
  test.artifacts.screenshot = 'screenshot'
})

Scenario('test store-test-and-suite test', ({ test }) => {
  assert(test.artifacts)
  assert(test.meta)
  test.meta.browser = 'chrome'
})

After(({ test }) => {
})

```

Based on feedback

* changed color for check
* changed standard acting helper location to Container (plugin to be removed in 4.0)
* added AI checker for check
* rolled back codeceptjs object to executed for event.all.before and event.all.after


lib/command/check.js
```diff
@@ -23,6 +22,7 @@ module.exports = async function (options) {
     container: false,
     pageObjects: false,
     plugins: false,
+    ai: true, // we don't need to check AI
```
kobenguyent, line 25: I'm just a bit confused here. Is setting true meaning to skip the check?

DavertMik: it should not fail if off

kobenguyent (approved):

ngraf:
‼️ When I use this branch, my **Typescript compiler fails** when I run

```
$ npx codeceptjs run                                                                                          

TSError: ⨯ Unable to compile TypeScript:
src/config/config.ts:7:10 - error TS2305: Module '"codeceptjs"' has no exported member 'container'.
```

Line 7 in `src/config/config.ts` looks like this:
```
import { container } from 'codeceptjs';
```

.. I have the impression that any kind of import fails. Not just for `container`.
I guess there is something fundamentally broken in this branch.
I hope you can easily reproduce.
If not I can provide more details.

DavertMik:
@ngraf is that only here or on 3.x as well?

also try other things except of Container 

sorry, I don't use TS much so I don't know how better to test it

DavertMik:
I think I fixed this issue. Please re-check TS

ngraf:
I installed the beta version in a wrong way. My fault.
Ignore the reported issue in https://github.com/codeceptjs/CodeceptJS/pull/4785#issuecomment-2619475436

PR #4831, opened 2025-02-06
Fixed running retryFailedStep inside tryTo

DavertMik:
Fixed regression when retryFailedStep was enabled inside tryTo block

* moved switch from env var to store
* added autoRetry store setting
* fixed and adjusted tests
* changed how to disable test for retryFailedStep

Other changes

* improved `check` messages
* fixed subtitles test
* fixed TS definitions for Step
* within introduced as part of effects module (should be moved to effects in 3.8/4.0)


lib/command/check.js
```diff
@@ -131,22 +132,36 @@ module.exports = async function (options) {
...
-    try {
-      for (const helper of Object.values(helpers)) {
+    checks.setup = true
+    for (const helper of Object.values(helpers)) {
+      try {
         if (helper._beforeSuite) await helper._beforeSuite(suite)
         if (helper._before) await helper._before(test)
+      } catch (err) {
+        err.message = `${helper.constructor.name} helper: ${err.message}`
+        if (checks.setup instanceof Error) err.message = `${err.message}\n\n${checks.setup?.message || ''}`.trim()
```
kobenguyent, line 142: out of curiosity, the line 135 defines as a `boolean`  and then line 143 we assign an `Error` to it.
Shall we assign `err.message` to `checks.setup?.message`

lib/command/check.js
```diff
@@ -131,22 +132,36 @@ module.exports = async function (options) {
...
+    for (const helper of Object.values(helpers).reverse()) {
+      try {
         if (helper._passed) await helper._passed(test)
         if (helper._after) await helper._after(test)
         if (helper._finishTest) await helper._finishTest(suite)
         if (helper._afterSuite) await helper._afterSuite(suite)
+      } catch (err) {
+        err.message = `${helper.constructor.name} helper: ${err.message}`
+        if (checks.teardown instanceof Error) err.message = `${err.message}\n\n${checks.teardown?.message || ''}`.trim()
+        checks.teardown = err
```
kobenguyent, line 159: same as above concern

lib/effects.js
```diff
@@ -178,11 +179,14 @@ async function tryTo(callback) {
   const sessionName = 'tryTo'
 
   let result = false
+  let hasAutoRetriesEnabled = store.autoRetries
   return recorder.add(
     sessionName,
     () => {
       recorder.session.start(sessionName)
-      store.tryTo = true
+      hasAutoRetriesEnabled = store.autoRetries
```
kobenguyent: how about `isAutoRetriesEnabled`?

lib/helper/Playwright.js
```diff
@@ -484,7 +484,7 @@ class Playwright extends Helper {
     this.currentRunningTest = test
 
     recorder.retry({
-      retries: process.env.FAILED_STEP_RETRIES || 3,
+      retries: test?.opts?.conditionalRetries || 3,
```
kobenguyent, line 487: as you touched this, what do you think if we store default retries = 3 in `Store`?

lib/step/base.js
```diff
@@ -195,8 +201,8 @@ class Step {
           args.push(arg.name)
         } else if (typeof arg == 'string') {
           args.push(arg)
-        } else {
-          args.push(JSON.stringify(arg).slice(0, 300))
+        } else if (arg) {
+          args.push((JSON.stringify(arg) || '').slice(0, 300))
```
kobenguyent, line 205: is there any chance `JSON.stringify(arg)` is `null` or `undefined` then '' is used here?

lib/command/check.js
DavertMik: good point

DavertMik: yes

the value of setup is `boolean|| Error`

if it is error, we pick a previous message and append to current error
if it is true but we fail, we write error
if it is true and never fail we have true

See line 170

lib/helper/Playwright.js
DavertMik: then we need a better name for it...
this handles a very strange edge cases in Playwright / Puppeteer
I don't know what name in store is applicable for this and in which other contexts can be used

lib/step/base.js
DavertMik: I'd rather not print undefined here...

Not sure about null

But I'd rather not, it's not readable...

kobenguyent (approved):

Arhell (approved):

PR #4832, opened 2025-02-07
add icons to check command

kobenguyent:
screenshot 20


Arhell (approved):

DavertMik:
Idea is cool but I'm not sure how it will work on old-fashioned non-interactive terminals like in CI systems Jenkins, Bamboo, etc

And check is mostly supposed to run on CI

PR #4890, opened 2025-03-06
cli: improve get info

kobenguyent:

```
➜  helloworld npx codeceptjs info
 Environment information: 

codeceptVersion:  "3.7.2"
nodeInfo:  18.19.0
osInfo:  macOS 14.4
cpuInfo:  (8) x64 Apple M1 Pro
osBrowsers:  "chrome: 133.0.6943.143, edge: 133.0.3065.92, firefox: not installed, safari: 17.4"
playwrightBrowsers:  "chromium: 133.0.6943.16, firefox: 134.0, webkit: 18.2"
helpers:  {
 "Playwright": {
  "url": "http://localhost",
...
```

Applicable helpers:

- [ ] Playwright
- [ ] Puppeteer
- [ ] WebDriver
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] TestCafe

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] stepTimeout
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [ ] :bug: Bug fix
- [ ] 🧹 Chore
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/lib/command/info.js
+++ b/lib/command/info.js
@@ -53,7 +53,7 @@ module.exports = async function (path) {
   info.osInfo = await envinfo.helpers.getOSInfo()
   info.cpuInfo = await envinfo.helpers.getCPUInfo()
   info.osBrowsers = await getOsBrowsers()
-  info.playwrightBrowsers = await getInstalledBrowsers()
+  info.playwrightBrowsers = await getPlaywrightBrowsers()
 
   const { helpers, plugins } = config
   info.helpers = helpers || "You don't use any helpers"
@@ -82,7 +82,7 @@ module.exports.getMachineInfo = async () => {
     edgeInfo: await envinfo.helpers.getEdgeInfo(),
     firefoxInfo: await envinfo.helpers.getFirefoxInfo(),
     safariInfo: await envinfo.helpers.getSafariInfo(),
-    playwrightBrowsers: await getInstalledBrowsers(),
+    playwrightBrowsers: await getPlaywrightBrowsers(),
   }
 
   output.print('***************************************')
--- a/lib/command/info.js
+++ b/lib/command/info.js
@@ -4,7 +4,7 @@ const { getConfig, getTestRoot } = require('./utils')
 const Codecept = require('../codecept')
 const output = require('../output')
 
-async function getInstalledBrowsers() {
+async function getPlaywrightBrowsers() {
   try {
     const playwright = require('playwright')
     const browsers = await Promise.all(
--- a/lib/command/info.js
+++ b/lib/command/info.js
@@ -4,22 +4,56 @@ const { getConfig, getTestRoot } = require('./utils')
 const Codecept = require('../codecept')
 const output = require('../output')
 
+async function getInstalledBrowsers() {
+  try {
+    const playwright = require('playwright')
+    const browsers = await Promise.all(
+      ['chromium', 'firefox', 'webkit'].map(async browser => {
+        try {
+          const browserInstance = await playwright[browser].launch()
+          const version = await browserInstance.version()
+          await browserInstance.close()
+          return `${browser}: ${version}`
+        } catch (err) {
+          return `${browser}: not installed`
+        }
+      }),
+    )
+    return browsers.join(`, `)
+  } catch (err) {
+    return 'Playwright not installed'
+  }
+}
+
+async function getOsBrowsers() {
+  const chromeInfo = await envinfo.helpers.getChromeInfo()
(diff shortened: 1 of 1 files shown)


lib/command/info.js
```diff
@@ -4,22 +4,56 @@ const { getConfig, getTestRoot } = require('./utils')
 const Codecept = require('../codecept')
 const output = require('../output')
 
+async function getInstalledBrowsers() {
```
DavertMik:
```suggestion
async function getPlaywrightBrowsers() {
```

lib/command/info.js
```diff
@@ -4,22 +4,56 @@ const { getConfig, getTestRoot } = require('./utils')
...
+  try {
+    const playwright = require('playwright')
+    const browsers = await Promise.all(
+      ['chromium', 'firefox', 'webkit'].map(async browser => {
+        try {
+          const browserInstance = await playwright[browser].launch()
```
DavertMik: no, launching them seems as too much
can we check them without launching?

Can we do it with this: https://playwright.dev/docs/api/class-browsertype#browser-type-executable-path

DavertMik (changes requested):
Better not to launch browsers
But check otherwise

lib/command/info.js
kobenguyent: this one just returns something like this not the version `/Library/Caches/ms-playwright/chromium-1155/chrome-mac/Chromium.app/Contents/MacOS/Chromium`

mirao: https://github.com/microsoft/playwright/blob/main/packages/playwright-core/browsers.json exists and contains current browser versions.
I have it installed in `./node_modules/.pnpm/playwright-core@1.50.0/node_modules/playwright-core/browsers.json`.
 But not sure how easily/universally get that file on various platforms (OS and npm vs. pnpm)

mirao: > https://github.com/microsoft/playwright/blob/main/packages/playwright-core/browsers.json exists and contains current browser versions

But it's for all browsers and there is no info that the browser was really installed. So it's probably useless.

mirao: Co-Pilot advised me this (and it works). Tested in TypeScript.

```ts
import { chromium, firefox, webkit } from "playwright";
import { execSync } from "child_process";
```

This returns `Chromium 133.0.6943.16`
```js
const chromeVersion = execSync(
    `"${chromium.executablePath()}" --version`
)
    .toString()
    .trim();
```
The same way works for Firefox (e.g. `Mozilla Firefox 134.0`)
```js
const firefoxVersion = execSync(
    `"${firefox.executablePath()}" --version`
)
    .toString()
    .trim();
```            
Unfortunately if  you use the same way for WebKit, you get only `WebKitGTK 2.47.2 (unknown)` and not the more common "Safari" version (e.g. 18.2)
```js
const webkitVersion = execSync(`"${webkit.executablePath()}" --version`)
    .toString()
    .trim();
```

kobenguyent: well, I'm not sure if there are easy ways to handle this unless we launch the browsers and get the version. 

what do you think we just return the webkit path to acknowledge this webkit browser is available from playwright?
@DavertMik @mirao

mirao: As for me displaying of Chromium/Firefox version would be great as they change quite often with every Playwright release and as for WebKit any string/build is enough as Safari version doesn't change too often.

kobenguyent: let's see how the playwright team proposes then -> https://github.com/microsoft/playwright/issues/35123

kobenguyent:
```
 Environment information: 

codeceptVersion:  "3.7.2"
nodeInfo:  18.19.0
osInfo:  macOS 14.4
cpuInfo:  (8) x64 Apple M1 Pro
osBrowsers:  "chrome: 134.0.6998.45, edge: 134.0.3124.51, firefox: not installed, safari: 17.4"
playwrightBrowsers:  "chromium: 134.0.6998.35, firefox: 135.0, webkit: 18.4"
helpers:  {
 "Playwright": {
  "browser": "chromium",
```

worked like a charm for now @DavertMik @mirao

thomashohn (approved):

PR #5127, opened 2025-09-01
Bugfix: Properly stop network traffic recording  

Samuel-StO:
### Context  
In the Playwright helper, `startRecordingTraffic()` attaches a listener on the `requestfinished` event to capture completed network requests.  
However, `stopRecordingTraffic()` was calling `removeAllListeners('request')`, which does not remove `requestfinished` listeners. This traffic is still recording event after the stopRecordingTraffic.  

Also, in `seeTraffic`, the condition was too strict: it required both `recording === true` *and* `recordedAtLeastOnce`. This prevented assertions on traffic after `stopRecordingTraffic()`.  

### Fix  
- Updated `stopRecordingTraffic()` to remove listeners for both `request` and `requestfinished`.  
  - Instead of changing the event listened to in `startRecordingTraffic()`, I chose to keep `requestfinished` and just ensure proper cleanup. This minimizes changes and avoids breaking compatibility.  
- Simplified `seeTraffic` condition: now it only checks `recordedAtLeastOnce`. Users can assert collected traffic even after recording is stopped.  

### Tests  
- Added a test ensuring that after calling `stopRecordingTraffic()`, new requests are no longer collected.  

### Impact  
- Keeps `actions.js` generic and safe.  
- No API changes.  
- Fixes stop recordings and makes `seeTraffic` usable after stopping traffic recording.  

Applicable helpers:

- [x] Playwright
- [x] Puppeteer
- [ ] WebDriver
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] TestCafe

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] stepTimeout
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [x] :bug: Bug fix
- [ ] 🧹 Chore
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [x] Tests have been added

### DIFF
--- a/lib/helper/Puppeteer.js
+++ b/lib/helper/Puppeteer.js
@@ -2591,7 +2591,8 @@ class Puppeteer extends Helper {
    *
    * {{> stopRecordingTraffic }}
    */
-  stopRecordingTraffic() {
+  async stopRecordingTraffic() {
+    await this.page.setRequestInterception(false)
     stopRecordingTraffic.call(this)
   }
 
--- a/lib/helper/network/actions.js
+++ b/lib/helper/network/actions.js
@@ -1,68 +1,77 @@
-const assert = require('assert')
-const { isInTraffic, createAdvancedTestResults, getTrafficDump } = require('./utils')
+const assert = require('assert');
+const { isInTraffic, createAdvancedTestResults, getTrafficDump } = require('./utils');
 
 function dontSeeTraffic({ name, url }) {
   if (!this.recordedAtLeastOnce) {
-    throw new Error('Failure in test automation. You use "I.dontSeeTraffic", but "I.startRecordingTraffic" was never called before.')
+    throw new Error('Failure in test automation. You use "I.dontSeeTraffic", but "I.startRecordingTraffic" was never called before.');
   }
 
   if (!name) {
-    throw new Error('Missing required key "name" in object given to "I.dontSeeTraffic".')
+    throw new Error('Missing required key "name" in object given to "I.dontSeeTraffic".');
   }
 
   if (!url) {
-    throw new Error('Missing required key "url" in object given to "I.dontSeeTraffic".')
+    throw new Error('Missing required key "url" in object given to "I.dontSeeTraffic".');
   }
 
   if (isInTraffic.call(this, url)) {
-    assert.fail(`Traffic with name "${name}" (URL: "${url}') found, but was not expected to be found.`)
+    assert.fail(`Traffic with name "${name}" (URL: "${url}') found, but was not expected to be found.`);
   }
 }
 
-async function seeTraffic({ name, url, parameters, requestPostData, timeout = 10 }) {
+async function seeTraffic({
+  name, url, parameters, requestPostData, timeout = 10,
+}) {
   if (!name) {
-    throw new Error('Missing required key "name" in object given to "I.seeTraffic".')
+    throw new Error('Missing required key "name" in object given to "I.seeTraffic".');
   }
 
   if (!url) {
-    throw new Error('Missing required key "url" in object given to "I.seeTraffic".')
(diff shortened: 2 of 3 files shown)


Samuel-StO:
Hmm, the tests are failing on Puppeteer. I’ll take a look tomorrow to try to understand why. Do you have any idea why the second amOnPage doesn’t pass? I’m not very familiar with Puppeteer.

Samuel-StO:
For Puppeteer, I think I may have identified the issue, but I’d like confirmation since I’m not an expert in Puppeteer or the Codecept recorder.

In this helper we set `await this.page.setRequestInterception(true)`:
[https://github.com/codeceptjs/CodeceptJS/blob/15a288eca40f4062c61ebbffcc4ded4bc23d211d/lib/helper/Puppeteer.js#L2602-L2610](https://github.com/codeceptjs/CodeceptJS/blob/15a288eca40f4062c61ebbffcc4ded4bc23d211d/lib/helper/Puppeteer.js#L2602-L2610)

However, when we stop, we never set this value back to `false`:
[https://github.com/codeceptjs/CodeceptJS/blob/15a288eca40f4062c61ebbffcc4ded4bc23d211d/lib/helper/Puppeteer.js#L2594-L2596](https://github.com/codeceptjs/CodeceptJS/blob/15a288eca40f4062c61ebbffcc4ded4bc23d211d/lib/helper/Puppeteer.js#L2594-L2596)

My proposal is to modify it like this:

```js
async stopRecordingTraffic() {
  await this.page.setRequestInterception(false);
  stopRecordingTraffic.call(this);
}
```

From what I’ve read, if `requestInterception` is set to `true`, we must always handle the requests. That means we should reset it to `false` when we flush the listeners: [https://github.com/puppeteer/puppeteer/issues/2055#issuecomment-366789261](https://github.com/puppeteer/puppeteer/issues/2055#issuecomment-366789261)

The tests passed with this modification, but I have one last question: what would be the impact of making `stopRecordingTraffic` asynchronous?

kobenguyent:
> For Puppeteer, I think I may have identified the issue, but I’d like confirmation since I’m not an expert in Puppeteer or the Codecept recorder.
> 
> In this helper we set `await this.page.setRequestInterception(true)`:
> 
> https://github.com/codeceptjs/CodeceptJS/blob/15a288eca40f4062c61ebbffcc4ded4bc23d211d/lib/helper/Puppeteer.js#L2602-L2610
> 
> However, when we stop, we never set this value back to `false`:
> 
> https://github.com/codeceptjs/CodeceptJS/blob/15a288eca40f4062c61ebbffcc4ded4bc23d211d/lib/helper/Puppeteer.js#L2594-L2596
> 
> My proposal is to modify it like this:
> 
> ```js
> async stopRecordingTraffic() {
>   await this.page.setRequestInterception(false);
>   stopRecordingTraffic.call(this);
> }
> ```
> 
> From what I’ve read, if `requestInterception` is set to `true`, we must always handle the requests. That means we should reset it to `false` when we flush the listeners: [puppeteer/puppeteer#2055 (comment)](https://github.com/puppeteer/puppeteer/issues/2055#issuecomment-366789261)
> 
> The tests passed with this modification, but I have one last question: what would be the impact of making `stopRecordingTraffic` asynchronous?

thanks for pointing it out @Samuel-StO, maybe it's a missing piece when trying to consolidate the actions to be shared among Puppeteer, Playwright and WebDriver helpers -> https://github.com/codeceptjs/CodeceptJS/pull/4263/files#diff-8b5994d9478083cafc2a9a5943b10cbff2648e67b092689c43e5f3d8924b4096R107

PR #5211, opened 2025-09-24
Update reports.md: use "reporter" option instead of CLI

danielrentz:

In `docs/reports.md`, the examples for Mocha reporter configurations should use the `"mocha":{"reporter":"..."}` option in codecept.config.js instead of the `--reporter` CLI option to point out that a reporter is always tied to its options.

This eliminates the risk that users forget to specify the correct reporter at some place (package.json scripts, CI scripts, ...).

It is pointless anyway to put the options of one reporter into the config file, and to specify another reporter in the CLI. This may even cause errors/exceptions from the running reporter if it sees invalid config options in the file (seen with mocha-multi).

Applicable helpers:

- [ ] Playwright
- [ ] Puppeteer
- [ ] WebDriver
- [ ] REST
- [ ] FileHelper
- [ ] Appium
- [ ] TestCafe

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] stepTimeout
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [ ] :bug: Bug fix
- [ ] 🧹 Chore
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added


kobenguyent:
@danielrentz I'm not sure about this change, perhaps @DavertMik could advise on this.

DavertMik:
@danielrentz I think both options are correct and there are situations where both can be nice to have
So I would better show all alternatives if possible

danielrentz:
@kobenguyent @DavertMik

Thanks for your feedback. I changed the PR to add the config-only variant as an alternative below the original docs.

The PR contains an additional typo fix in the Mochawecome chapter (copy/pasted "HTML" from above instead of "mochawesome").

The remaining changes in the file are caused by the generator script `npm run docs`.

PR #5381, opened 2026-01-12
Add warning for ts-node/esm usage

DenysKuchma:
## Problem
When using `ts-node/esm` with `"module": "esnext"` in tsconfig.json, users get "Cannot find module" errors because imports without file extensions don't work.
This is a known limitation of `ts-node/esm`, not a bug in CodeceptJS, but users don't understand what's happening. https://github.com/codeceptjs/CodeceptJS/issues/5323

 ## Solution
  Add a warning message when `ts-node/esm` is detected in the config. The warning:
  - Explains the limitation clearly
  - Recommends `tsx/cjs` as an alternative
  - Links to documentation

### DIFF
--- a/lib/utils/loaderCheck.js
+++ b/lib/utils/loaderCheck.js
@@ -107,6 +107,26 @@ Note: TypeScript config files (codecept.conf.ts) and helpers are automatically
 }
 
 /**
+ * Get warning message if ts-node/esm is being used
+ * @param {string[]} requiredModules - Array of required modules from config
+ * @returns {string|null} Warning message or null
+ */
+export function getTSNodeESMWarning(requiredModules = []) {
+  if (!requiredModules.includes('ts-node/esm')) {
+    return null
+  }
+
+  return `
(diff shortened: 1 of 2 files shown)


lib/utils/loaderCheck.js
```diff
@@ -106,6 +106,26 @@ Note: TypeScript config files (codecept.conf.ts) and helpers are automatically
...
+ */
+export function getTSNodeESMWarning(requiredModules = []) {
+  if (!requiredModules.includes('ts-node/esm')) {
+    return null
+  }
+
+  return `
+⚠️  Warning: ts-node/esm with "module": "esnext" requires explicit file extensions in all imports.
+
+This is a known limitation. Use tsx/cjs instead to write imports without extensions.
```
DavertMik, line 122: Please add an example here

Which import is OK which is not OK

DavertMik (changes requested):

DavertMik (approved):

PR #5382, opened 2026-01-13
Export container as named export for ESM

DenysKuchma:
## Problem
  TypeScript cannot import `container` in ESM: https://github.com/codeceptjs/CodeceptJS/issues/5320

## Solution
  Added named export in typings/index.d.ts

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -520,86 +520,6 @@ declare namespace CodeceptJS {
   }
 
   function addStep(step: string | RegExp, fn: Function): Promise<void>
-
-  // Internal API types
-  class Container {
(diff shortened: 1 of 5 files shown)


typings/index.d.ts
```diff
@@ -637,6 +637,7 @@ declare namespace Mocha {
 
 declare module 'codeceptjs' {
   export default codeceptjs
+  export const container: any
```
DavertMik: can we add a type for container?

DavertMik (changes requested):
also can we list all other things we can export from codeceptjs?

like output, recorder, etc? 

Like listed here https://codecept.io/internal-api/#concepts

DavertMik (approved):

PR #5383, opened 2026-01-13
Improve waitUrlEquals error message to include actual error

DenysKuchma:
## Problem
  When `waitUrlEquals` times out but expected URL matches current URL:  https://github.com/codeceptjs/CodeceptJS/issues/5208

## Solution
  Check if URLs match. If yes, show the actual error message

### DIFF
--- a/lib/helper/Playwright.js
+++ b/lib/helper/Playwright.js
@@ -3546,27 +3546,23 @@ class Playwright extends Helper {
       expectedUrl = baseUrl + urlPart
     }
 
-    return this.page
-      .waitForURL(
-        url => {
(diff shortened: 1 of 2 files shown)


DavertMik (approved):

DavertMik (changes requested):
Can you maybe think to rewrite this function completely? 
I don't like its code...

Maybe we can make a better version at least for Playwright?

We can use this method https://playwright.dev/docs/api/class-page#page-wait-for-url
https://playwright.dev/docs/api/class-page#page-wait-for-url

DavertMik (approved):

PR #5421, opened 2026-01-29
Fixed: FileSystem.waitForFile: do not crash if directory does not exist yet

### LINKED ISSUE #5420: `I.waitForFile` (FileSystem helper) fails if directory does not exist (yet)
`I.waitForFile` sometimes fails and throws weird exceptions if the directory containing the expected file does not exist yet.

Example (all output goes to `${codecept_dir}/output/`):
```js
  I.handleDownloads('downloads/example.txt');
  I.click('button.download');
  I.amInPath('output/downloads');
  I.waitForFile('example.txt', 5);
```

If the directory `${codecept_dir}/output/downloads` does not exist yet (it will be created by the browser download at some point in time), the function `I.waitForFile` will throw an exception `ReferenceError: Cannot access 'watcher' before initialization` and the test will fail.

Root cause is in the function [`isFileExists`](https://github.com/codeceptjs/CodeceptJS/blob/3.x/lib/helper/FileSystem.js#L202-L227) of the `FileSystem` helper. The line `const watcher = fs.watch(dir, (eventType, filename) => {...` will throw if `dir` does not exist yet. This exception remains unhandled. Later, the timer for the 5sec timeout detection will try to close the uninitialized `watcher`, leading to the `ReferenceError`.

I will try to create a PR.

danielrentz:
- Resolves #5420

Reworked the complex implementation of `FileSystem.waitForFile`, now it contains a simple polling loop that uses the existing `fileExists` helper function like `FileSystem.seeFile` does.

Applicable helpers:

- [ ] Playwright
- [ ] Puppeteer
- [ ] WebDriver
- [ ] REST
- [x] FileHelper
- [ ] Appium
- [ ] TestCafe

Applicable plugins:

- [ ] allure
- [ ] autoDelay
- [ ] autoLogin
- [ ] customLocator
- [ ] pauseOnFail
- [ ] coverage
- [ ] retryFailedStep
- [ ] screenshotOnFail
- [ ] selenoid
- [ ] stepByStepReport
- [ ] stepTimeout
- [ ] wdio
- [ ] subtitles

## Type of change

- [ ] :fire: Breaking changes
- [x] :bug: Bug fix
- [ ] 🧹 Chore
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added


lib/helper/FileSystem.js
```diff
@@ -84,12 +84,12 @@ class FileSystem extends Helper {
...
   async waitForFile(name, sec = 1) {
     if (sec === 0) assert.fail('Use `seeFile` instead of waiting 0 seconds!')
-    const waitTimeout = sec * 1000
     this.file = path.join(this.dir, name)
     this.debugSection('File', this.file)
-    return isFileExists(this.file, waitTimeout).catch(() => {
-      throw new Error(`file (${name}) still not present in directory ${this.dir} after ${waitTimeout / 1000} sec`)
-    })
+    for (const t = Date.now() + sec * 1000; !fileExists(this.file); ) {
+      if (t < Date.now()) throw new Error(`file (${name}) still not present in directory ${this.dir} after ${sec} sec`)
```
DavertMik: sorry I don't understand how it works

But I agree current impleentation should be improved
Can we use while loop and some interval like 200ms to check this

And if it fails after sec seconds => throw error

DavertMik (changes requested):

lib/helper/FileSystem.js
danielrentz: @DavertMik 
- The `for`-loop initializes `t` as desired timeout timestamp ("now" plus timeout seconds), and loops as long as the fie does not exist.
- The if condition throws out of the loop if the timeout has been reached.
- A 500ms interval delay is contained in the last line of the loop (the `await new Promise`).

Of course, the for loop could be rewritten as while loop to improve readability, and the interval time could be lowered to 200ms if you prefer.

danielrentz: Done. Changed for-loop to while-loop, lowered interval to 200ms.

PR #5435, opened 2026-02-06
Add seeCurrentPathEquals method to ignore query strings

DavertMik:
Adds new assertion method to check URL path equality while ignoring query parameters and URL fragments.

## Usage

```js
I.seeCurrentPathEquals('/info'); // passes for '/info', '/info?user=1', '/info#section'
I.seeCurrentPathEquals('/');     // passes for '/', '/?user=ok', '/#top'
```

## Implementation

- Uses native `URL` class for pathname extraction
- Added to Playwright, Puppeteer, and WebDriver helpers
- Includes tests and documentation

Closes #{add-issue-number-if-any}

### DIFF
--- a/lib/helper/WebDriver.js
+++ b/lib/helper/WebDriver.js
@@ -1845,6 +1845,16 @@ class WebDriver extends Helper {
   }
 
   /**
+   * {{> seeCurrentPathEquals }}
+   */
+  async seeCurrentPathEquals(path) {
+    const currentUrl = await this.browser.getUrl()
+    const baseUrl = this.options.url || 'http://localhost'
+    const actualPath = new URL(currentUrl, baseUrl).pathname
+    return assert.equal(path, actualPath, `expected url path to be ${path}, but found ${actualPath}`)
+  }
+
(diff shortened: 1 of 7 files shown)


lib/helper/WebDriver.js
```diff
@@ -1844,6 +1844,28 @@ class WebDriver extends Helper {
...
+  /**
+   * {{> dontSeeCurrentPathEquals }}
+   */
+  async dontSeeCurrentPathEquals(path) {
+    const currentUrl = await this.browser.getUrl()
+    const baseUrl = this.options.url || 'http://localhost'
+    const actualPath = new URL(currentUrl, baseUrl).pathname
+    const errorMessage = `expected url path not to be ${path}, but found ${actualPath}`
+    const isEqual = path === actualPath
+    if (isEqual) throw new Error(errorMessage)
```
DenysKuchma: Can we use a similar function for Playwright and Puppeteer, without custom logic at the end?

DenysKuchma (approved):

PR #5436, opened 2026-02-06
feat: add switchToTab

gimler:
## Add method to switch to tab

Our app open a window and in this window we call `window.close()` after clicking on button.

The problem is that `switchToPreviousTab()` need a active tab. In our app the window/tab is already close by our application.

So with the new method `switchToTab` you can switch to any open tab.

Applicable helpers:

- [x] Playwright

## Type of change

- [x] :bug: Bug fix

## Checklist:

- [x] Tests have been added

## Question

`switchToNextTab` also call `await targetCreatedHandler.call(this, page)` but `switchToPreviousTab` not. I am not shure if `switchToTab` should also call `await targetCreatedHandler.call(this, page)`?

We can also remove the duplicate code and use `switchTotTab` in `switchToPreviousTab()` and `switchToNextTab()`

### DIFF
--- a/docs/helpers/Playwright.md
+++ b/docs/helpers/Playwright.md
@@ -2335,6 +2335,19 @@ I.switchToPreviousTab(2);
 
 *   `num` **[number][20]**  
 
+### switchToTab
+
+Switch focus to a particular tab by its number. It waits tabs loading and then switch tab
+
+```js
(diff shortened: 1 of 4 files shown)


gimler:
@kobenguyent anything missing?

PR #5438, opened 2026-02-09
fix: run-workers --by suite parallelization broken by test file sorting (4.x)

mirao:
## Summary
- Cherry-pick of #5419 to the `4.x` branch (as requested in https://github.com/codeceptjs/CodeceptJS/pull/5419#issuecomment-3868408807)
- Moved `testFiles.sort()` from `loadTests()` to `run()` so worker distribution uses original file order
- Adapted CommonJS to ESM: converted `alphabetical_order_test.js` to use `import`/`export`, removed `require.cache` cleanup, added `await workers._ensureInitialized()` for async Workers constructor

## Test plan
- [x] `npx mocha test/unit/alphabetical_order_test.js` — passes (1 test)
- [x] `npx mocha test/unit/worker_test.js --grep "should not sort"` — passes (1 test)

### DIFF
--- a/lib/codecept.js
+++ b/lib/codecept.js
@@ -284,6 +284,9 @@ class Codecept {
       // Ignore if gherkin module not available
     }
 
+    // Sort test files alphabetically for consistent execution order
+    this.testFiles.sort()
+
     return new Promise((resolve, reject) => {
       const mocha = container.mocha()
       mocha.files = this.testFiles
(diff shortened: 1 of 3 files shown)


DenysKuchma (approved):

mirao:
@DavertMik Could you look at this and merge to 4.x eventually? Can't verify in 4.x as we still use 3.x, but IMHO the fix didn't get into 4.x and it's in 3.x only.

PR #5474, opened 2026-03-05
fix: make XPath relative in buildLocatorString for within() scope

### LINKED ISSUE #5473: locate().at().find() doesn't work inside within() with Playwright helper
The report was created for me by Claude Sonnet 4.6.

## What are you trying to achieve?

Use `locate().at().find()` inside `within()` to find elements scoped to the `within()` context.

## What do you get instead?

The XPath generated by `at()` and `find()` bypasses the `within()` scope and searches from the document root. Elements from outside the `within()` context are matched instead.

## Minimal reproduction

**codecept.conf.js:**
```js
exports.config = {
  tests: './*_test.js',
  output: './output',
  helpers: {
    Playwright: {
      browser: 'chromium',
      url: 'http://localhost',
      show: false,
    },
  },
};
```

**within_test.js:**
```js
Feature('within + locate().at().find()');

Scenario('locate().at().find() should work inside within()', async ({ I }) => {
  I.amOnPage('data:text/html,<div id="outer"><ul id="list1"><li class="item"><span class="label">First</span></li></ul><ul id="list2"><li class="item"><span class="label">Second</span></li></ul></div>');

  // This works - plain CSS inside within()
  within('#list2', () => {
    I.see('Second', '.label');
  });

  // This FAILS - locate().at().find() inside within()
  // It finds "First" from #list1 instead of "Second" from #list2,
  // because the XPath escapes the within() scope
  within('#list2', () => {
    I.see('Second', locate('.item').at(1).find('.label'));
  });
});
```

**Actual output (FAILS):**
```
expected element {xpath: (//*[...item...])[position()=1]//*[...label...]}
to include "Second"

- First
+ Second
```

The first `within()` block (plain CSS) passes. The second one (with `locate().at().find()`) fails because the XPath searches from the document root, finding "First" from `#list1` instead of "Second" from `#list2`.

## Root cause

The issue is in the interaction between `buildLocatorString()` in `lib/helper/Playwright.js` and Playwright's XPath engine.

**Playwright's XPath engine** (in `injectedScriptSource.js`) auto-converts absolute XPath to relative when the root is not a Document:

```js
// From Playwright's XPathEngine.queryAll:
if (selector.startsWith("/") && root.nodeType !== Node.DOCUMENT_NODE)
    selector = "." + selector;
```

This converts `//div` → `.//div` when searching within an element. **But it only triggers when the selector starts with `/`.**

**CodeceptJS's `Locator.at()`** (in `lib/locator.js`) wraps the XPath in parentheses:

```js
at(position) {
    const xpath = sprintf('(%s)[position()=%s]', this.toXPath(), xpathPosition);
    return new Locator({ xpath });
}
```

This produces `(//*[...])[position()=1]` which starts with `(`, **not** `/`, so Playwright's auto-conversion never triggers. The `//` inside the parentheses then searches from the document root via `document.evaluate()`, completely ignoring the `within()` scope.

## Suggested fix

In `buildLocatorString()` in `lib/helper/Playwright.js`, make XPath relative before passing it to Playwright:

```js
function buildLocatorString(locator) {
  if (locator.isCustom()) {
    return buildCustomLocatorString(locator)
  }
  if (locator.isXPath()) {
    // Make XPath relative so it works correctly within scoped contexts (e.g. within()).
    // Playwright's XPath engine auto-converts "//..." to ".//..." when the root is not a Document,
    // but only when the selector starts with "/". Locator methods like at() wrap XPath in
    // parentheses (e.g. "(//...)[position()=1]"), bypassing that auto-conversion.
    // We fix this by prepending "." before the first "//" that follows any leading parentheses.
    const value = locator.value.replace(/^(\(*)\/\//, '$1.//');
    return `xpath=${value}`
  }
  return locator.simplify()
}
```

This is safe because `.//` is equivalent to `//` when evaluated from the document root, but correctly scopes to descendants when evaluated from an element.

## Affected methods

Any Locator DSL method that wraps XPath in parentheses before the leading `//`:
- `locate().at()` → `(//...)[position()=N]`
- `locate().at().find()` → `(//...)[position()=N]//...`
- `locate().first()` / `locate().last()` (which call `at()`)

## Details

- **CodeceptJS version:** 3.7.6
- **NodeJS version:** 20.x
- **Helper:** Playwright
- **Browser:** Chromium (likely affects all browsers since it's an XPath semantics issue)

mirao:
## Summary

- Make XPath relative in `buildLocatorString()` so `locate().at()`, `.first()`, `.last()` work correctly inside `within()` with Playwright helper
- Playwright's XPath engine auto-converts `//...` to `.//...` when the root is not a Document, but only when the selector starts with `/`. Locator methods like `at()` wrap XPath in parentheses (e.g. `(//...)[position()=1]`), bypassing that auto-conversion
- The fix prepends `.` before the first `//` that follows any leading parentheses

Fixes #5473

## Test plan

- [x] Unit tests for `buildLocatorString` covering `at()`, `first()`, `last()`, `at().find()`, plain XPath, already-relative XPath, and CSS passthrough
- [x] All 336 Playwright helper integration tests pass (1 pre-existing failure in `#seeAttributesOnElements` unrelated to this change)

### DIFF
--- a/lib/helper/Playwright.js
+++ b/lib/helper/Playwright.js
@@ -4134,9 +4134,15 @@ class Playwright extends Helper {
 
 export default Playwright
 
-function buildLocatorString(locator) {
+export function buildLocatorString(locator) {
   if (locator.isXPath()) {
-    return `xpath=${locator.value}`
+    // Make XPath relative so it works correctly within scoped contexts (e.g. within()).
+    // Playwright's XPath engine auto-converts "//..." to ".//..." when the root is not a Document,
+    // but only when the selector starts with "/". Locator methods like at() wrap XPath in
+    // parentheses (e.g. "(//...)[position()=1]"), bypassing that auto-conversion.
+    // We fix this by prepending "." before the first "//" that follows any leading parentheses.
+    const value = locator.value.replace(/^(\(*)\/\//, '$1.//')
+    return `xpath=${value}`
   }
   if (locator.isShadow()) {
     // Convert shadow locator to CSS with >> chaining operator
--- a/test/acceptance/within_test.js
+++ b/test/acceptance/within_test.js
@@ -1,5 +1,12 @@
 Feature('within', { retries: 3 })
 
+Scenario('within with locate().at().find() should scope XPath @Playwright', async ({ I }) => {
+  I.amOnPage('/form/bug5473')
+  await within('#list2', async () => {
+    await I.see('Second', locate('.item').at(1).find('.label'))
+  })
+})
+
 Scenario('within on form @WebDriverIO @Puppeteer @Playwright', async ({ I }) => {
   I.amOnPage('/form/bug1467')
   I.see('TEST TEST')
--- a/test/data/app/view/form/bug5473.php
+++ b/test/data/app/view/form/bug5473.php
@@ -1,0 +1,14 @@
+<!doctype html>
+<html>
+    <head><title>within + locate().at() bug</title></head>
+    <body>
+        <div id="outer">
+            <ul id="list1">
+                <li class="item"><span class="label">First</span></li>
+            </ul>
+            <ul id="list2">
+                <li class="item"><span class="label">Second</span></li>
+            </ul>
+        </div>
+    </body>
+</html>
--- a/test/unit/helper/Playwright_buildLocatorString_test.js
+++ b/test/unit/helper/Playwright_buildLocatorString_test.js
@@ -1,0 +1,49 @@
+import { expect } from 'chai'
+import Locator from '../../../lib/locator.js'
+import { buildLocatorString } from '../../../lib/helper/Playwright.js'
+
+describe('buildLocatorString', () => {
+  it('should make plain XPath relative', () => {
+    const locator = new Locator({ xpath: '//div' })
+    expect(buildLocatorString(locator)).to.equal('xpath=.//div')
+  })
+
+  it('should make XPath with parentheses (from at()) relative', () => {
+    const locator = new Locator('.item').at(1)
+    const result = buildLocatorString(locator)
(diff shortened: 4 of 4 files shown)


DavertMik:
@mirao I didn't see how it is related to within
I didn't find in code or in tests that within will actually work with it

Can we detect that we are triggering locator from page root or from an other element and apply transformation only when it is needed?

mirao:
> Posted by Claude Code on behalf of @mirao

@DavertMik Good points, I've pushed a second commit that adds an acceptance test proving it works with `within()`:

**test/acceptance/within_test.js** — new scenario `within with locate().at().find() should scope XPath`:
```js
Scenario('within with locate().at().find() should scope XPath @Playwright', async ({ I }) => {
  I.amOnPage('/form/bug5473')
  await within('#list2', async () => {
    await I.see('Second', locate('.item').at(1).find('.label'))
  })
})
```

**Without the fix:** fails — finds "First" from `#list1` (XPath escapes the `within()` scope)
**With the fix:** passes — correctly finds "Second" inside `#list2`

Regarding applying the transformation only when needed: `buildLocatorString` is a pure utility function with no access to context state (`this.withinLocator`). Making it context-aware would require threading state through all call sites.

However, the transformation is safe to apply unconditionally because `.//` is semantically equivalent to `//` when evaluated from the document root — Playwright's own XPath engine does the [same conversion](https://github.com/nicknisi/playwright/blob/99f36f99/packages/playwright-core/src/server/injected/selectorUtils.ts#L42-L43). The only difference is `.//` also works correctly when evaluated from a non-root element, which is exactly what `within()` does.

DavertMik:
Thank you! Let's have it!

Issue #5351, opened 2025-12-18
Bug Report: stepByStepReport not capturing screenshots in CodeceptJS 3.6+


ThomasSanson:
Hello,

The `stepByStepReport` plugin does not capture screenshots after each step in CodeceptJS versions 3.6.0 through 3.7.5. The `record_*/` directory is created but only contains an empty `index.html` file with no screenshots (PNG files).

## Affected Versions

- **Broken**: CodeceptJS 3.6.0, 3.6.1, 3.6.2, 3.6.3, 3.6.4, 3.7.0, 3.7.1, 3.7.2, 3.7.3, 3.7.4, 3.7.5
- **Working**: CodeceptJS 3.5.0 and earlier

## Root Cause

The bug was introduced in commit [1040494](https://github.com/codeceptjs/CodeceptJS/commit/10404945b99e53e0acfef37f7d19baaea669e34e) (Jun 8, 2024) titled "fix: screenshot error in beforeSuite/AfterSuite (#4385)".

### The Change

In `lib/plugin/stepByStepReport.js`, the `event.step.after` handler was changed:

**Before (working - v3.5.0):**
```javascript
event.dispatcher.on(event.step.after, (step) => {
  recorder.add('screenshot of failed test', async () => persistStep(step), true);
});
```

**After (broken - v3.6+):**
```javascript
event.dispatcher.on(event.step.after, persistStep);
```

### Why This Breaks

The `persistStep` function is an `async` function that calls `helper.saveScreenshot()`. In the **original implementation**, this was wrapped in `recorder.add()`, which properly queues the async operation and waits for completion.

In the **new implementation**, `persistStep` is called directly as an event handler callback. Since it's an async function, the event dispatcher doesn't wait for it to complete. The screenshots are never captured because:

1. The async function starts executing
2. The event dispatcher immediately moves on to the next event
3. The screenshot operation may be cancelled or never completes

## Reproduction Steps

1. Create a minimal CodeceptJS config with only `stepByStepReport` enabled
2. Run any test that fails (to keep the report)
3. Check the `_output/record_*/` directory
4. The directory contains only `index.html` with no PNG files

### Minimal Reproduction

```javascript
// codecept.minimal.conf.js
const path = require('path');

exports.config = {
  output: path.resolve(__dirname, '_output'),
  helpers: {
    Playwright: {
      browser: 'chromium',
      url: 'https://www.google.com',
      show: true
    }
  },
  plugins: {
    stepByStepReport: {
[...]

PR #5508, opened 2026-04-07
Fix for stepByStepReport: cheerio ESM import and missing screenshot capture

gololdf1sh:
## Summary

Fix two bugs in `stepByStepReport` plugin that made it completely non-functional in v4.x (ESM):

- **cheerio ESM import:** `import cheerio from 'cheerio'` fails because cheerio 1.x removed the default export. Changed to `import * as cheerio from 'cheerio'`
- **Screenshots never captured:** `if (!step.test) return` in `persistStep()` always returned early because `step.test` property doesn't exist on the `Step` class. Replaced with `if (!currentTest) return` which correctly checks the plugin's existing `currentTest` variable

## How it was tested

- Enabled plugin with `{ enabled: true, deleteSuccessful: false }`
- Ran a UI test with multiple steps (clicks, navigation, form fills)
- Result: 55 PNG screenshots captured + working HTML slideshow report

Relates to #5351

### DIFF
--- a/lib/plugin/stepByStepReport.js
+++ b/lib/plugin/stepByStepReport.js
@@ -4,7 +4,7 @@ import figures from 'figures'
 import fs from 'fs'
 import { mkdirp } from 'mkdirp'
 import path from 'path'
-import cheerio from 'cheerio'
+import * as cheerio from 'cheerio'
 
 import Container from '../container.js'
 import recorder from '../recorder.js'
@@ -198,7 +198,7 @@ export default function (config) {
     // Ignore steps from BeforeSuite function
     if (scenarioFailed && config.disableScreenshotOnFail) return
     if (step.metaStep && step.metaStep.name === 'BeforeSuite') return
-    if (!step.test) return // Ignore steps from AfterSuite
+    if (!currentTest) return // Ignore steps from AfterSuite
 
     const fileName = `${pad.substring(0, pad.length - stepNum.toString().length) + stepNum.toString()}.png`
     if (step.status === 'failed') {


gololdf1sh:
## Evidence

### Bug 1: Plugin crashes on load (cheerio ESM import)

**Before fix** — `import cheerio from 'cheerio'` crashes immediately, tests cannot run at all:

```
Could not load plugin stepByStepReport from module './plugin/stepByStepReport.js':
Could not load plugin from './plugin/stepByStepReport.js': The requested module 'cheerio' does not provide an export named 'default'
Error: Could not load plugin from './plugin/stepByStepReport.js': The requested module 'cheerio' does not provide an export named 'default'
    at loadPluginAsync (file:///node_modules/codeceptjs/lib/container.js:638:11)
```

### Bug 2: Screenshots never captured (`step.test` is always undefined)

**Before fix** — after fixing only the cheerio import, the plugin loads but produces an empty report with **0 screenshots**:

```
$ ls output/record_*/
index.html          # only index.html, no PNG files
```

The generated `index.html` slideshow is completely empty — no slides, no screenshots:

empty report 

**Root cause:** `persistStep()` at line 201 checks `if (!step.test) return` but the `Step` class (`lib/step/base.js`) has no `test` property — it's never assigned anywhere in the codebase. So `!step.test` is always `true` and every screenshot is skipped.

### After fix (both changes applied)

```
$ ls output/record_*/
0000.png  0005.png  0010.png  0015.png  0020.png  0025.png  0030.png  0035.png  0040.png  0045.png  0050.png
0001.png  0006.png  0011.png  0016.png  0021.png  0026.png  0031.png  0036.png  0041.png  0046.png  0051.png
0002.png  0007.png  0012.png  0017.png  0022.png  0027.png  0032.png  0037.png  0042.png  0047.png  0052.png
0003.png  0008.png  0013.png  0018.png  0023.png  0028.png  0033.png  0038.png  0043.png  0048.png  0053.png
0004.png  0009.png  0014.png  0019.png  0024.png  0029.png  0034.png  0039.png  0044.png  0049.png  0054.png  0055.png
index.html

$ ls -la output/record_*/*.png | wc -l
56
```

**56 PNG screenshots** (~160KB each) captured across all test steps + a working HTML slideshow.

### Config used for testing

```js
stepByStepReport: {
  enabled: true,
  deleteSuccessful: false,
}
```

PR #5518, opened 2026-04-12
feat: programmatic API with getSuites(), executeSuite(), executeTest()

DavertMik:
## Summary

- Add `getSuites(pattern?)` to `Codecept` class — returns parsed suites with tests as plain objects, using a temporary Mocha instance to avoid polluting container state
- Add `executeSuite(suite)` and `executeTest(test)` — clean execution methods that accept objects returned by `getSuites()`
- Refactor `workers.js` — `createGroupsOfTests()` and `createGroupsOfSuites()` now use `getSuites()` instead of duplicated Mocha boilerplate (~30 lines removed)
- Refactor `dryRun.js` — `printTests()` now uses `getSuites()`, removing direct `Container.mocha()` dependency
- Export `Result` class from `lib/index.js`
- Rewrite "Custom Runner" section in `docs/internal-api.md` as full "Programmatic API" reference

## Usage

```js
import { Codecept, container } from 'codeceptjs'

const codecept = new Codecept(config, { steps: true })
await codecept.init(__dirname)

// List tests
const suites = codecept.getSuites()
for (const suite of suites) {
  console.log(suite.title)
  for (const test of suite.tests) {
    console.log(' -', test.title, test.tags)
  }
}

// Execute
await codecept.bootstrap()
for (const suite of suites) {
  await codecept.executeSuite(suite)
}
const result = container.result()
await codecept.teardown()
```

## Test plan

- [x] `dry-run` command works with refactored `getSuites()` (tested with grep filtering)
- [x] Unit tests pass (401 passing, 7 pre-existing AI test failures unrelated)
- [x] Shard tests pass
- [ ] Workers should be tested with `run-workers 2`


lib/runner.js
```diff
@@ -0,0 +1,116 @@
...
+
+    return this._execute({ files, grep })
+  }
+
+  async runSuite(suite) {
+    return this.run(suite.file)
+  }
+
+  async runTest(test) {
+    return this._execute({ grep: test.fullTitle })
```
DenysKuchma, line 63: test.fullTitle is passed directly to mocha.grep() below, but Mocha treats string grep as a regular expression pattern. Test titles with regex metacharacters like `[brackets]` or `(parens)` will not match themselves and `runTest()` ends up running 0 tests.

docs/internal-api.md
```diff
@@ -225,42 +225,167 @@ Step events provide step objects with following fields:
...
-
-// define main config
-const config = { 
-  helpers: { 
-    WebDriver: { 
-      browser: 'chrome', 
-      url: 'http://localhost' 
-    }
-  }
+import { Codecept, container } from 'codeceptjs';
```
DenysKuchma, line 235: This example does not work with the current exports: `Codecept` is available on the default export object, but it is not a named export. `import { Codecept, container } from 'codeceptjs'` throws `does not provide an export named 'Codecept'`

lib/workers.js
```diff
@@ -117,34 +117,11 @@ const createWorkerObjects = (testGroups, config, testRoot, options, selectedRuns
...
-        0,
-        workerName,
-      )
-      _config.mocha.reporterOptions['mocha-junit-reporter'].options.mochaFile = _tempArray.join(separator)
-    }
-    workerName = worker.getOriginalName() || worker.getName()
-    const workerConfig = worker.getConfig()
-    workersToExecute.push(getOverridenConfig(workerName, workerConfig, _config))
+    const workerName = worker.name.replace(':', '_')
+    const _config = mainConfig.applyRunConfig(config, worker.getConfig())
```
DenysKuchma, line 122: This no longer preserves the previous reporter path overrides for multiple worker runs. With reporterOptions configured, multiple browser/worker runs can write to the same report path and overwrite or mix reports. Was that removal intentional?

DenysKuchma (changes requested):

PR #5545, opened 2026-04-30
fix(workers): make hardcoded run-workers overall timeout configurable

gololdf1sh:
## Problem

`run-workers` has a hardcoded 10-min overall timeout (added in #5370). When it fires, all still-running workers are force-terminated and their results never reach the aggregated summary — silently dropped. No CLI/env/config override exists.

Hit this on a 4-worker run of a ~50-min sequential suite: only 28% of tests reported, with `[Main] Overall timeout reached (10 minutes)` at t=10m07s.

## Fix

Make the timeout configurable. Default `600000` ms preserved (backward-compat). Resolution order:

1. `CODECEPT_WORKERS_TIMEOUT` env var (ms)
2. `workersTimeout` in codecept.conf (ms)
3. Default `600000`

`0` disables the timeout entirely. Log message minutes value is now dynamic.

## Changes

- `lib/workers.js`: extract `_getWorkersTimeoutMs()`, gate `setTimeout` behind `> 0` check
- `typings/index.d.ts`: add `workersTimeout?: number` to `MainConfig`
- `test/unit/worker_test.js`: 6 unit tests (default / env / config / precedence / disabled / non-numeric fallback) — all pass

## Type of change

- [x] :bug: Bug fix

### DIFF
--- a/lib/workers.js
+++ b/lib/workers.js
@@ -501,6 +501,26 @@ class Workers extends EventEmitter {
     return runHook(this.codecept.config.teardownAll, 'teardownAll')
   }
 
+  /**
+   * Resolves the overall `run-workers` timeout in milliseconds.
+   *
+   * Resolution order:
+   * 1. `CODECEPT_WORKERS_TIMEOUT` env var (numeric, ms)
+   * 2. `workersTimeout` in codecept.conf (ms)
+   * 3. Default: 600000 (10 minutes)
+   *
+   * Set the value to `0` (or any non-positive number) to disable the timeout entirely.
+   *
+   * @returns {number} timeout in milliseconds
+   */
+  _getWorkersTimeoutMs() {
+    const envTimeout = parseInt(process.env.CODECEPT_WORKERS_TIMEOUT, 10)
+    if (Number.isFinite(envTimeout)) return envTimeout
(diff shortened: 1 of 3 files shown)


lib/workers.js
```diff
@@ -501,6 +501,26 @@ class Workers extends EventEmitter {
...
+   * Set the value to `0` (or any non-positive number) to disable the timeout entirely.
+   *
+   * @returns {number} timeout in milliseconds
+   */
+  _getWorkersTimeoutMs() {
+    const envTimeout = parseInt(process.env.CODECEPT_WORKERS_TIMEOUT, 10)
+    if (Number.isFinite(envTimeout)) return envTimeout
+    const configTimeout = this.codecept?.config?.workersTimeout
+    if (Number.isFinite(configTimeout)) return configTimeout
+    return 600000
```
DavertMik: `600 000` = `10x60x100`

Add const

```
const WORKER_TIMEOUT_MINUTES =10
```
```suggestion
    return WORKER_TIMEOUT_MINUTES*60*1000
```

DavertMik (changes requested):
Move default timeout to const

DavertMik:
Fixed in #5548

PR #5562, opened 2026-05-14
fix(plugin): guard against undefined step.name in retryFailedStep

gololdf1sh:
## Problem

`retryFailedStep` plugin crashes workers in `run-workers` mode:

```
TypeError: Cannot read properties of undefined (reading 'startsWith')
```

The plugin reads `step.name`, but in worker mode steps are serialized via `simplify()` which drops `name`.

## Fix

Replace `step.name` → `step.title` in the plugin. Depends on #5564 which renames `Step.name` → `Step.title` across core.

## Related

- #5564 — core rename `Step.name` → `Step.title`
- testomatio/e2e-tests#139 — original report

### DIFF
--- a/lib/plugin/retryFailedStep.js
+++ b/lib/plugin/retryFailedStep.js
@@ -111,11 +111,12 @@ export default function (config) {
   }
 
   event.dispatcher.on(event.step.started, step => {
+    if (!step.title) return
     for (const ignored of config.ignoredSteps) {
-      if (step.name === ignored) return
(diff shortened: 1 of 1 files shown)


gololdf1sh:
Root-cause fix: #5564

Instead of patching the plugin with a fallback, this renames `Step.name` → `Step.title` across core + all plugins, so the field name is consistent between full Step objects and `simplify()` serialized ones. No functional change — same value, same field name everywhere.

PR #5565, opened 2026-05-15
fix(workers): deserialize step events before emitting to master process

gololdf1sh:
## Problem

In `run-workers` mode, step events (`step.started`, `step.finished`, `step.passed`, `step.failed`) are emitted as raw plain objects from IPC serialization. These objects lack Step class methods (`toString()`, `toCliStyled()`), so any plugin or reporter calling `step.toString()` gets `"[object Object]"` instead of the actual step description.

This affects all consumers of step events in worker mode — for example, `@testomatio/reporter` shows:

```
################[ Logs ]################
--- Test ---
[object Object]
 [object Object]
```

Test events already go through `deserializeTest()` which restores a full object with methods. Step events had no equivalent deserialization — this PR adds it.

## Solution

Add `deserializeStep()` — a one-liner that wraps the serialized data in a `Step` instance via `Object.assign(new Step(data.title), data)`, matching the pattern already used inside `deserializeTest` for `test.steps` (see `lib/mocha/test.js:80`).

Apply it to all four step event types before emitting.

## Related

- testomatio/e2e-tests#151 — original report with reproduction steps

### DIFF
--- a/lib/workers.js
+++ b/lib/workers.js
@@ -19,6 +19,7 @@ import output from './output.js'
 import event from './event.js'
 import { deserializeTest } from './mocha/test.js'
 import { deserializeSuite } from './mocha/suite.js'
+import Step from './step/base.js'
 import recorder from './recorder.js'
 import store from './store.js'
 import runHook from './hooks.js'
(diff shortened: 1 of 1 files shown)


DenysKuchma (changes requested):
Add regression tests

PR #5634, opened 2026-06-13
fix(effects): make within/tryTo/hopeThat/retryTo work under tsx/cjs (#5632)

### LINKED ISSUE #5632: 4.x: within() and tryTo() don't execute
Create a simple project using https://codecept.io/quickstart

My_test.ts
```ts
import { within } from "codeceptjs";
import { tryTo } from "codeceptjs/effects";

Feature("My");

  I.amOnPage("https://example.com");
  tryTo(() => {
    I.see("Example DomainX");
  });
  within("body", () => {
    I.click("a");
  });
});
```

codecept.conf.ts:
```ts
import { setHeadlessWhen, setCommonPlugins } from '@codeceptjs/configure';

setCommonPlugins();

  tests: './tests/*_test.ts',
  output: './output',
  helpers: {
    Playwright: {
      browser: 'chromium',
      url: 'http://localhost',
      show: true
    }
  },
  include: {
    I: './steps_file.ts'
  },
  noGlobals: true,
  plugins: {},
  name: 'my',
  require: ['tsx/cjs']
}
```

tsconfig.json:
```json
{
  "compilerOptions": {
    "target": "ES2022",
    "lib": ["ES2022", "DOM"],
    "esModuleInterop": true,
    "module": "ESNext",
    "moduleResolution": "bundler",
    "strictNullChecks": false,
    "types": ["codeceptjs", "node"],
    "declaration": true,
    "skipLibCheck": true
  },
  "exclude": ["node_modules"]
}
```

**Expected result:**

- both `within()` and `tryTo()` work

In 3.7.9, it works well ✔️ :
```
mirao@rog:~/workspace/my_379$ codeceptjs run --verbose
nodeInfo:  24.15.0
chromeInfo:  149.0.7827.114
edgeInfo:  "N/A"
firefoxInfo:  151.0.3
safariInfo:  N/A
CodeceptJS v3.7.9 #StandWithUkraine
Using test root "/home/mirao/workspace/my_379"
Helpers: Playwright
Plugins: screenshotOnFail, retryFailedStep, eachElement

My --
/home/mirao/workspace/my_379/My_test.ts
    Timeouts: 
  test something
  Scenario()
    I am on page "https://example.com"
 › Auto retries disabled inside tryTo effect
    I see "Example DomainX"
    [1] <tryTo>  Error (Non-Terminated) | Error | err => { step.status = 'failed' step.endTime= +Da...
 › Unsuccessful try > expected web application to include "Example DomainX"
    Within "body"
      I click "a"
      › [Clicked] <a href="https://iana.org/domains/example">Learn more</a>...
  ✔ OK in 1187ms


  OK  | 1 passed   // 1s
```

**Actual result:**

- `within()` and `tryTo()` are not executed at all in 4.x. They are missing in the test output and e.g. if you even used `I.click("ax")` instead of `I.click("a")` no error would be reported 🐛 

4.0.7:

```
mirao@rog:~/workspace/my$ codeceptjs run tests/My_test.ts --verbose
nodeInfo:  24.15.0
chromeInfo:  149.0.7827.114
edgeInfo:  "N/A"
firefoxInfo:  151.0.3
safariInfo:  N/A
CodeceptJS v4.0.7 #StandWithUkraine
Using test root "/home/mirao/workspace/my"
Helpers: Playwright
Plugins: screenshot, retryFailedStep

My --
/home/mirao/workspace/my/tests/My_test.ts
    Timeouts: 
  test something
  Scenario()
    I am on page "https://example.com"
  ✔ OK in 146ms
```

**Used SW:**
- NodeJS v24.15.0, Ubuntu 24.04

package.json
```json
{
  "devDependencies": {
    "@types/node": "^25.9.3",
    "codeceptjs": "^4.0.7",
    "playwright": "^1.60.0",
    "tsx": "^4.22.4",
    "typescript": "^6.0.3"
  }
}
```

mirao:
### Problem

`within()` and `tryTo()` (and `hopeThat()` / `retryTo()`) silently do nothing in 4.x when a TypeScript test imports effects through a CommonJS loader, e.g.:

```ts
import { within } from "codeceptjs";
import { tryTo } from "codeceptjs/effects";
```

run under `require: ['tsx/cjs']` (the setup the official Quickstart generates).

### Root cause

CodeceptJS 4.x is native ESM, and it loads test files through Mocha's **synchronous `require()`** (`lib/mocha/factory.js`), i.e. the **CommonJS realm**. So when a test imports `codeceptjs/effects`, `tsx/cjs` turns it into a `require()` and Node materializes a **second, disconnected CJS copy** of `effects.js` together with its own copies of the `recorder` and `container` singletons.

That recorder is never started (`running = false`), so every `recorder.add()` short-circuits to `Promise.resolve()`:

- `tryTo()` / `hopeThat()` return `undefined` without ever running their callback (no error is reported even if the inner steps would fail),
- `within()` skips all of its inner steps, and its empty CJS container never calls `_withinBegin`, so the context is never applied,
- `retryTo()` never resolves and **hangs**.

Because test files are always loaded via `require()`, this is the default for essentially every TypeScript project — including the bare Quickstart.

### Fix

Make the singletons realm-agnostic using the same `globalThis` bridge the framework already uses for `global.codeceptjs` (`lib/globals.js`):

- **`lib/recorder.js`** — `start()` registers the running instance on `globalThis.__codeceptjs_recorder` (only the instance the runner actually starts registers itself).
- **`lib/container.js`** — `create()` registers the live container on `globalThis.__codeceptjs_container`.
- **`lib/effects.js`** — resolves the recorder/container through `_getRecorder()` / `_getContainer()`, which prefer the `globalThis` instances and fall back to the local module imports. Under pure ESM both point at the same object, so there is **no behavior change** on the ESM path.

### Tests

Adds a runner regression test that drives a **real `tsx/cjs` project** and asserts all four effects execute:

- `test/data/effects-tsx-cjs/` — fixture (config with `require: ["tsx/cjs"]`, a helper, and a test that imports effects through the CJS loader).
- `test/runner/effects_tsx_test.js` — asserts `within`, `tryTo`, `hopeThat` and `retryTo` all run.

It **fails without the fix** and passes with it. The `exec` call has a timeout guard so that a future regression which re-hangs `retryTo()` fails cleanly instead of hanging the suite.

The existing pure-ESM coverage (`test/unit/effects_test.js`, `test/runner/within_test.js`) continues to pass — that path previously had **no** coverage for the CJS-loaded case, which is why the regression shipped.

### DIFF
--- a/lib/container.js
+++ b/lib/container.js
@@ -80,6 +80,14 @@ class Container {
     this.createMocha = () => (container.mocha = MochaFactory.create(mochaConfig, opts || {}))
     this.createMocha()
 
+    // Expose this container globally so a duplicate copy of the framework (loaded as
+    // CommonJS when a test does `import { within } from 'codeceptjs'` through a CJS
+    // loader such as tsx/cjs) resolves helpers, support objects and plugins from the
+    // live container that actually runs the tests. Without this bridge `within()` sees
+    // an empty helpers map and never calls `_withinBegin`, so the within-context is
+    // silently ignored. Mirrors the `globalThis.__codeceptjs_recorder` bridge in recorder.js.
+    if (typeof globalThis !== 'undefined') globalThis.__codeceptjs_container = Container
+
     // create support objects
     container.support = {}
     container.helpers = await createHelpers(config.helpers || {})
--- a/lib/effects.js
+++ b/lib/effects.js
@@ -7,38 +7,46 @@ import MetaStep from './step/meta.js'
 import { empty } from './assert/empty.js'
 import { isAsyncFunction } from './utils.js'
 
+// When a test imports effects through a CommonJS loader (e.g. tsx/cjs), a second,
+// disconnected copy of this module and its `recorder`/`container` singletons is loaded.
+// That copy's recorder is never started and its container has no helpers, so effects
+// would silently do nothing. Resolve the live instances registered on globalThis by the
+// runner instead, falling back to the local singletons when running purely under ESM.
+const _getRecorder = () => (typeof globalThis !== 'undefined' && globalThis.__codeceptjs_recorder) || recorder
+const _getContainer = () => (typeof globalThis !== 'undefined' && globalThis.__codeceptjs_container) || container
+
 /**
  * @param {CodeceptJS.LocatorOrString}  context
  * @param {Function}  fn
  * @return {Promise<*> | undefined}
  */
 function within(context, fn) {
-  const helpers = store.dryRun ? {} : container.helpers()
+  const helpers = store.dryRun ? {} : _getContainer().helpers()
   const locator = typeof context === 'object' ? JSON.stringify(context) : context
 
-  return recorder.add(
+  return _getRecorder().add(
     'register within wrapper',
     () => {
       const metaStep = new WithinStep(locator, fn)
       const defineMetaStep = step => (step.metaStep = metaStep)
-      recorder.session.start('within')
+      _getRecorder().session.start('within')
 
       event.dispatcher.prependListener(event.step.before, defineMetaStep)
 
       Object.keys(helpers).forEach(helper => {
-        if (helpers[helper]._withinBegin) recorder.add(`[${helper}] start within`, () => helpers[helper]._withinBegin(context))
+        if (helpers[helper]._withinBegin) _getRecorder().add(`[${helper}] start within`, () => helpers[helper]._withinBegin(context))
       })
 
       const finalize = () => {
         event.dispatcher.removeListener(event.step.before, defineMetaStep)
-        recorder.add('Finalize session within session', () => {
+        _getRecorder().add('Finalize session within session', () => {
           output.stepShift = 1
-          recorder.session.restore('within')
+          _getRecorder().session.restore('within')
         })
       }
       const finishHelpers = () => {
         Object.keys(helpers).forEach(helper => {
-          if (helpers[helper]._withinEnd) recorder.add(`[${helper}] finish within`, () => helpers[helper]._withinEnd())
+          if (helpers[helper]._withinEnd) _getRecorder().add(`[${helper}] finish within`, () => helpers[helper]._withinEnd())
         })
       }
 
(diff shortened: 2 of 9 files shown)


mirao:
Heads-up: #5636 (fix for #5635) addresses the same dual-realm split as this PR, but **at the source** rather than in `effects.js`.

#5636 makes the `recorder` and `container` singletons themselves realm-agnostic (via a shared `globalThis` instance in the new `lib/realm.js`), so the CJS-loaded copy of `effects.js` resolves the **live** recorder/container directly through its normal `import` — no `_getRecorder()`/`_getContainer()` indirection needed. The same source-level fix also covers `config`, `event`, `store`, `output`, and transitively `within`/`session`/`pause`/`inject`/etc.

I verified empirically on the #5636 branch (which does **not** include this PR — `effects.js` is unpatched there) that all four effects run from a CJS-loaded test:

```
EFFECTS_TRYTO result=false
EFFECTS_CLICK withinActive=true
EFFECTS_PASS ran
EFFECTS_HOPETHAT result=true
EFFECTS_FLAKY try=2
EFFECTS_RETRY done   ← retryTo resolves, does not hang
```

So this PR's `_getRecorder()`/`_getContainer()` changes become redundant once #5636 lands (both sides of `globalThis.__codeceptjs_recorder || recorder` resolve to the same shared object).

To avoid losing the valuable regression coverage from this PR, I've **ported its effects test** into #5636 (`test/data/internal-api-tsx-cjs/effects_test.ts` + the corresponding runner assertions), so `within`/`tryTo`/`hopeThat`/`retryTo` under `tsx/cjs` stay guarded.

Suggest we **close this PR in favor of #5636** — but happy to keep it if you'd prefer the narrower change first. Either way the effects coverage is preserved.

mirao:
> Suggest we **close this PR in favor of #5636** — but happy to keep it if you'd prefer the narrower change first. Either way the effects coverage is preserved.

Closing as it was fixed in #5636 by more general way.

PR #5639, opened 2026-06-15
fix: guard sort() with shuffle check so --shuffle is not silently ignored

### LINKED ISSUE #5605: bug: --shuffle flag silently ignored after #5438 (sort() in run() overwrites shuffle from loadTests())
## Bug Report

**Introduced in:** commit `4f1eb3f` (PR #5438) — *fix: run-workers --by suite parallelization broken by test file sorting*

---

### What was fixed in #5438

`testFiles.sort()` was moved from `loadTests()` to `run()` to fix worker suite distribution. Good fix — but it introduced a regression for `--shuffle` users.

### The regression

In `lib/codecept.js`, the execution order is now:

```js
// loadTests() — line ~217
if (this.opts.shuffle) {
  this.testFiles = shuffle(this.testFiles)  // ✅ shuffled here
}

// run() — line 293 (added by #5438)
this.testFiles.sort()  // ❌ always sorts — overwrites the shuffle!

// then:
mocha.files = this.testFiles  // receives sorted, not shuffled, files
```

`sort()` in `run()` is unconditional — it runs regardless of whether `--shuffle` was specified. This means `--shuffle` is silently ignored on every run after this commit.

### Steps to reproduce

```bash
# Run twice with --shuffle — expect different order each time
npx codeceptjs run --shuffle
npx codeceptjs run --shuffle

# Both runs will produce identical alphabetical order — shuffle has no effect
```

### Expected behavior

`--shuffle` randomizes test execution order. Each run should produce a different sequence.

### Actual behavior

Both runs produce identical alphabetical order. The shuffle applied in `loadTests()` is overwritten by `sort()` in `run()`.

### Suggested fix

Guard the sort with a shuffle check:

```js
// run() — only sort if shuffle is not active
if (!this.opts.shuffle) {
  this.testFiles.sort()
}
```

This preserves alphabetical order for normal runs while respecting shuffle when requested.

---

Happy to submit a PR if the fix looks right to maintainers.

kapil971390:
Fixes #5605

## What

`run()` calls `this.testFiles.sort()` unconditionally after `loadTests()` has already applied `shuffle()` when `--shuffle` is set. The sort overwrites the randomised order every time, making `--shuffle` silently have no effect.

```js
// loadTests() — line 217
if (this.opts.shuffle) {
  this.testFiles = shuffle(this.testFiles)  // ✅ shuffled
}

// run() — line 293 (added by #5438)
this.testFiles.sort()  // ❌ always runs — shuffle is lost
```

## Fix

Guard the sort with `!this.opts.shuffle`:

```js
// Sort alphabetically for consistent order, but preserve shuffle when active.
if (!this.opts.shuffle) {
  this.testFiles.sort()
}
```

Normal runs continue to get alphabetical order. `--shuffle` runs now keep their randomised order.

## Why this happened

The `sort()` was introduced in #5438 to fix worker suite distribution — a correct fix for that issue, but it introduced this regression for `--shuffle` users.

## Change

Single guard condition in `lib/codecept.js` — no other behaviour changed.

### DIFF
--- a/lib/codecept.js
+++ b/lib/codecept.js
@@ -289,8 +289,11 @@ class Codecept {
       // Ignore if gherkin module not available
     }
 
-    // Sort test files alphabetically for consistent execution order
-    this.testFiles.sort()
+    // Sort test files alphabetically for consistent execution order,
+    // but skip sorting when --shuffle is active so the randomised order is preserved.
+    if (!this.opts.shuffle) {
+      this.testFiles.sort()
+    }
 
     return new Promise((resolve, reject) => {
       const mocha = container.mocha()


kapil971390:
Hi! Could a maintainer approve the workflows to run CI on this? The fix is a single guard condition in `lib/codecept.js` — happy to answer any questions.

kapil971390:
All relevant checks passing. appium failures are pre-existing, saucelabs infrastructure issues, unrelated to this change.

DavertMik:
Thank you for catching it!

kapil971390:
Hi @DavertMik really enjoying contributing to CodeceptJS. would love to stay involved and contribute more regularly.
Happy to help with open issues!! :-)

PR #5644, opened 2026-06-17
fix(run-multiple): avoid temp.mjs filename collision across forked processes

### LINKED ISSUE #5642: 4.x: The test in the mode "run-multiple" sometimes fails: "Cannot find module *.temp.mjs"
I hit it once after migration of my 3.x project to 4.x.
A problem is that the `customhelper.temp.mjs` is shared during the transpilation in "run-multiple" and when it's deleted by one of the processes the others are missing it.

Here is the setup and script (prepared by AI) how to reproduce the issue reliably:

repro/tests/a_test.ts
```ts
Feature('A');

Scenario('a noop', ({ I }) => {
  I.noop();
});
```
repro/tests/b_test.ts
```ts
Feature('B');

Scenario('b noop', ({ I }) => {
  I.noop();
});
```
repro/tests/c_test.ts
```ts
Feature('C');

Scenario('c noop', ({ I }) => {
  I.noop();
});

```

repro/codecept.conf.ts
```ts
import { WORKERS } from './shared';

// Minimal, browser-free CodeceptJS config that reproduces the run-multiple
// "Cannot find module *.temp.mjs" race in CodeceptJS 4.x.
//
// Every .ts file referenced here (this config, ./shared.ts, ./CustomHelper.ts,
// ./steps_file.ts and each tests/*_test.ts) is transpiled by the built-in
// TypeScript transpiler into a FIXED "<name>.temp.mjs" file next to the source,
// imported, then deleted on cleanup. run-multiple forks one OS process per
// "browser" entry, all at once, so the workers fight over the same temp paths.
  tests: './tests/*_test.ts',
  output: './output',
  helpers: {
    Custom: { require: './CustomHelper.ts' },
  },
  include: {
    I: './steps_file.ts',
  },
  multiple: {
    // One named run expanded into WORKERS.length concurrent forked processes.
    parallel: {
      browsers: WORKERS,
    },
  },
  name: 'repro',
  // Matches the real project. Note: this loader only affects *test* files.
  // codecept.conf.ts / helpers / include files are still transpiled by the
  // built-in transpiler into "*.temp.mjs" (it runs before any loader exists),
  // which is exactly what races under run-multiple.
  require: ['tsx/cjs'],
};

```

repro/CustomHelper.ts
```ts
// A browser-free helper so the repro needs no Playwright / network.
import Helper from '@codeceptjs/helper';

class CustomHelper extends Helper {
  async noop() {
    return true;
  }
}

export default CustomHelper;
```

repro/shared.ts
```ts
// Imported by codecept.conf.ts so the config itself transpiles into TWO temp
// files (codecept.conf.temp.mjs + shared.temp.mjs), widening the race window.
export const WORKERS: string[] = Array.from({ length: 12 }, (_, i) => `w${i + 1}`);

```

output
```
mirao@rog:~/workspace/my$ cd repro
for i in $(seq 1 40); do
  if node ../node_modules/codeceptjs/bin/codecept.js run-multiple --all -c . 2>&1 \
       | grep -qi "temp.mjs\|Cannot find module\|ENOENT"; then
    echo "iter $i FAILED"
  else
    echo "iter $i ok"
  fi
done
iter 1 FAILED
iter 2 FAILED
iter 3 FAILED
iter 4 FAILED
iter 5 FAILED
iter 6 FAILED
iter 7 FAILED
iter 8 FAILED
iter 9 ok
iter 10 FAILED
iter 11 ok
iter 12 FAILED
iter 13 FAILED
iter 14 FAILED
iter 15 ok
iter 16 FAILED
iter 17 FAILED
iter 18 ok
iter 19 FAILED
iter 20 FAILED
iter 21 FAILED
iter 22 FAILED
iter 23 FAILED
iter 24 FAILED
iter 25 FAILED
iter 26 ok
iter 27 FAILED
iter 28 FAILED
iter 29 FAILED
iter 30 FAILED
iter 31 FAILED
iter 32 FAILED
iter 33 FAILED
iter 34 FAILED
iter 35 FAILED
iter 36 FAILED
iter 37 ok
iter 38 FAILED
iter 39 FAILED
iter 40 ok
```

**Used SW:**

- CodeceptJS 4.0.7
- NodeJS 24

kapil971390:
## Problem

Fixes #5642

When `run-multiple` launches N concurrent worker processes, each process independently transpiles the same TypeScript files (config, helpers, page objects, includes) to temporary `.temp.mjs` files. Since the temp filename is derived purely from the source path:

```js
// before
const tempFile = filePath.replace(/\.ts$/, '.temp.mjs')
```

all workers write to the **same path** (e.g. `CustomHelper.temp.mjs`). When one worker finishes and deletes its temp file via `cleanupTempFiles()`, any other worker that hasn't yet `import()`ed the file fails with:

```
Error: Cannot find module '.../CustomHelper.temp.mjs'
```

This is a classic TOCTOU race; the repro in #5642 shows it failing ~85% of the time with 12 workers.

## Fix

Embed `process.pid` in the temp filename so each worker writes its own isolated copy:

```js
// after
const tempFile = filePath.replace(/\.ts$/, `.${process.pid}.temp.mjs`)
```

No other changes needed:
- `cleanupTempFiles()` iterates `allTempFiles` which is built from `transpiledFiles.values()` — already contains the pid-suffixed paths, so cleanup is correct.
- `lib/step/base.js` uses `.includes('.temp.mjs')` (substring), which matches `*.{pid}.temp.mjs` unchanged.

## Test plan

- [ ] `run-multiple` no longer races on the repro from #5642 (run 40 iterations, all pass)
- [ ] Single-process `run` still works (temp files created, imported, cleaned up normally)
- [ ] TypeScript configs, helpers, and page objects all load correctly

### DIFF
--- a/lib/utils/typescript.js
+++ b/lib/utils/typescript.js
@@ -385,7 +385,9 @@ const __dirname = __dirname_fn(__filename);
     )
 
     // Write the transpiled file with updated imports
-    const tempFile = filePath.replace(/\.ts$/, '.temp.mjs')
+    // Use process.pid in the filename to avoid collisions when run-multiple forks
+    // several processes that all transpile the same .ts files concurrently.
+    const tempFile = filePath.replace(/\.ts$/, `.${process.pid}.temp.mjs`)
     fs.writeFileSync(tempFile, jsContent)
     transpiledFiles.set(filePath, tempFile)
   }


kapil971390:
Closing as duplicate of #5643 by @mirao which fixes the same race condition. Great minds!

DavertMik:
Indeed. That's great!

PR #5653, opened 2026-06-30
fix(pause): stop pause() at the right step in 4.x (#5652)

### LINKED ISSUE #5652: CodeceptJS 4.x: pause() doesn't stop at the right place — it pauses after subsequent steps instead of before them
Tested with 4.0.x and 4.1 beta. Both are affected.
It works well with 3.7.9.

Run this test:

tests/My_test.ts:
```ts
Feature('My');

Scenario('test something', ({ I }) => {
    I.amOnPage('https://example.com');
    pause();
    I.see('Example Domain');
});
```

**Expected result:**
* the pause() stops the test **before** `I.see('Example Domain');`

**Actual result:**
* the pause() stops the test  **after** `I.see('Example Domain');`

```
mirao@rog:~/workspace/my_esm_beta$ codeceptjs run --verbose
nodeInfo:  24.16.0
chromeInfo:  149.0.7827.200
edgeInfo:  "N/A"
firefoxInfo:  152.0.3
safariInfo:  N/A
playwrightBrowsers:  "chromium: 149.0.7827.55, firefox: 151.0, webkit: 26.5"
CodeceptJS v4.0.8 #StandWithUkraine
Using test root "/home/mirao/workspace/my_esm_beta"
Helpers: Playwright
Plugins: screenshot

My --
/home/mirao/workspace/my_esm_beta/tests/My_test.ts
    Timeouts: 
  test something
  Scenario()
    I am on page "https://example.com"
    I see "Example Domain"
 Interactive shell started
 Use JavaScript syntax to try steps in action
 - Press ENTER to run the next step
 - Press TAB twice to see all available commands
 - Type exit + Enter to exit the interactive shell
 - Prefix => to run js commands 
 I.
```

**Workaround:**
- Use `await pause()`

**Configuration:**

codecept.conf.ts
```ts
  tests: './tests/*_test.ts',
  output: './output',
  helpers: {
    Playwright: {
      browser: 'chromium',
      url: 'http://localhost',
      show: true
    }
  },
  include: {
    I: './steps_file.ts'
  },
  noGlobals: true,
  plugins: {},
  name: 'my_esm_beta',
  require: ['tsx/cjs']
}
```

package.json
```json
{
  "devDependencies": {
    "@types/node": "^26.0.1",
    "codeceptjs": "^4.0.8",
    "playwright": "^1.61.1",
    "tsx": "^4.22.4",
    "typescript": "^6.0.3"
  }
}
```

Tested with NodeJS v24.16.0

mirao:
Fixes #5652

## Problem

In 4.x, `pause()` stops **after** the following step instead of **before** it:

```ts
Scenario('test something', ({ I }) => {
  I.amOnPage('https://example.com');
  pause();                  // expected: stops here
  I.see('Example Domain');  // actual: pause stops after this
});
```

Works correctly in 3.7.9; broken in 4.0.x and 4.1 beta.

## Root cause

`global.pause` was wired as an **async wrapper** doing a lazy dynamic import in `lib/globals.js`:

```js
global.pause = async (...args) => {
  const pauseModule = await import('./pause.js')
  return (pauseModule.default || pauseModule)(...args)
}
```

A scenario body runs **synchronously** to build the recorder queue — every `I.*` call does `recorder.add(...)` synchronously. Because `pause` was async, calling it returned at the `await import(...)`, deferring its `recorder.add('Start new session', ...)` to a later microtask — **after** the rest of the synchronous body had already been queued.

`DEBUG=codeceptjs:recorder` for the repro:

```
Queued | amOnPage
Queued | see: "Example Domain"     <-- whole body queued synchronously
Running | amOnPage
Queued | Start new session         <-- pause inserted AFTER `see`
```

This also explains why `await pause()` worked as a workaround: awaiting suspends the synchronous body until pause's `recorder.add` has run.

## Fix

`initCodeceptGlobals` is already `async`, so resolve the module up front and assign the real function directly, making `pause` synchronous at call time:

```js
const pauseModule = await import('./pause.js')
global.pause = pauseModule.default || pauseModule
```

Ordering is restored:

```
Queued | amOnPage
Queued | Start new session         <-- between the two steps
Queued | see: "Example Domain"
```

## Test

Added a regression test in `test/unit/pause_test.js` that wires `global.pause` via `initCodeceptGlobals` and asserts:
- `global.pause` is the real (synchronous) `pause` function, not an async wrapper;
- mirroring the repro (`amOnPage → pause → see`), the `Start new session` step is queued **between** the surrounding steps (checked synchronously via `recorder.scheduled()`).

The test fails against the old async-wrapper implementation and passes with the fix. Full `test/unit` suite is green (759 passing).

> Note: `global.within` and `global.session` use the same async-lazy-import pattern and share the same latent hazard (currently masked because they're always awaited). Left out of scope here.

### DIFF
--- a/lib/globals.js
+++ b/lib/globals.js
@@ -30,10 +30,8 @@ export async function initCodeceptGlobals(dir, config, container) {
   // pause/inject/share stay global even under noGlobals — they're the everyday
   // debugging/wiring entry points and have no useful import alternative for
   // page-object code that runs before the container is available.
-  global.pause = async (...args) => {
-    const pauseModule = await import('./pause.js')
-    return (pauseModule.default || pauseModule)(...args)
-  }
+  const pauseModule = await import('./pause.js')
+  global.pause = pauseModule.default || pauseModule
   global.inject = () => container.support()
   global.share = container.share
 
--- a/test/unit/pause_test.js
+++ b/test/unit/pause_test.js
@@ -4,6 +4,7 @@ import pause, { setPauseHandler, pauseNow } from '../../lib/pause.js'
 import recorder from '../../lib/recorder.js'
 import event from '../../lib/event.js'
 import store from '../../lib/store.js'
+import { initCodeceptGlobals } from '../../lib/globals.js'
 
 const settles = (promise, ms = 2000) =>
   Promise.race([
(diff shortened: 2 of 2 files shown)


DavertMik:
Thank you!

PR #5655, opened 2026-07-02
Add a custom Firefox binary example to the Playwright helper docs

feder-cr:
browser: 'firefox' plus a firefox: { executablePath } object already work here, mirroring how electron: { executablePath } is documented further down. This just adds a concrete example, motivated by invisible_playwright, a Playwright wrapper around a Firefox build patched at the source level for a realistic fingerprint. Opened as draft, happy to adjust wording or placement.


DavertMik:
thank you!

Issue #5645, opened 2026-06-19
4.x: junitReporter doesn't record Before/After suite failures


mirao:
In CodeceptJS 3.x there could be the "mocha-junit-reporter" configured in a CodeceptJS config in "mocha" section.
If the test failed in `BeforeSuite()`, `Scenario()` or `AfterSuite()` I could see the failure, typically in Jenkins's JUnit test result report plugin ✔️ 

For example:

<img width="2230" height="535" alt="Image" src="https://github.com/user-attachments/assets/ab096756-dab7-466b-a3c5-06badfd3232b" />

---

In CodeceptJS 4.x, it was replaced by the plugin "junitReporter". It doesn't record any failure when the test fails in `BeforeSuite()` or `AfterSuite()` 🐛 . It records only `Scenario()` failures.

tests/My_test.ts:
```ts
Feature('My');

BeforeSuite(async ({ I }) => {
    I.say("Before Suite");
    throw new Error("Before Suite Error");
});

Scenario('test something',  ({ I }) => {
    I.amOnPage("https://example.com");
});

AfterSuite(async ({ I }) => {
    I.say("After Suite");
    throw new Error("After Suite Error");
});

```
codecept.conf.ts
```ts
import { setHeadlessWhen, setCommonPlugins } from "@codeceptjs/configure";
// turn on headless mode when running with HEADLESS=true environment variable
// export HEADLESS=true && npx codeceptjs run
setHeadlessWhen(process.env.HEADLESS);

// enable all common plugins https://github.com/codeceptjs/configure#setcommonplugins
setCommonPlugins();

export const config: CodeceptJS.MainConfig = {
  tests: "./tests/*_test.ts",
  output: "./output",
  helpers: {
    Playwright: {
      browser: "chromium",
      url: "http://localhost",
      show: true,
    },
  },
  include: {
    I: "./steps_file.ts",
  },
  noGlobals: true,
  plugins: {
    junitReporter: {
      enabled: true,
    },
  },
  name: "my",
  require: ["tsx/cjs"],
};

```

output:
```
mirao@rog:~/workspace/my$ codeceptjs run --verbose
***************************************
nodeInfo:  24.16.0
osInfo:  Linux 6.17 Ubuntu 24.04.4 LTS 24.04.4 LTS (Noble Numbat)
cpuInfo:  (16) x64 AMD Ryzen 7 9700X 8-Core Processor
chromeInfo:  149.0.7827.155
[...]

PR #5664, opened 2026-07-18
fix(junit): preserve suite identity for hook failures reported from workers

kapil971390:
## What's the bug

Under `run-workers`, when a `BeforeSuite`/`AfterSuite` hook fails, the JUnit report puts the failure under a generic `<testsuite name="Tests">` instead of the suite that actually failed, and each failure gets split into its own `<testsuite>` element instead of being grouped with other failures from the same suite.

This is a follow-up to #5657, which fixed the main bug from #5645 (hook failures not appearing in the report at all). That PR's description explicitly notes this as a known limitation:

> Under `run-workers` the forwarded payload (`hook.simplify()`) doesn't carry the suite, so hook-failure testcases group under a fallback `Tests` suite instead of the feature title. Fixable later by including the suite title in the worker message.

This PR does that.

## Root cause

`lib/plugin/junitReporter.js`'s `event.hook.failed` listener reads the suite from `hook.ctx.test.parent`:

```js
const runnable = hook.ctx && hook.ctx.test
const suite = runnable && runnable.parent
```

On the main thread `hook.ctx` is a live Mocha context and this works (verified separately with a real Mocha run — `ctx.test` for a suite-level hook is the hook's own runnable, and `runnable.parent` is the real Suite). But under `run-workers`, the failure is forwarded from the worker process via `Hook.simplify()` (`lib/mocha/hooks.js`):

```js
simplify() {
  return {
    hookName: this.hookName,
    title: this.title,
    error: this.err ? serializeError(this.err) : null,
  }
}
```

No `ctx`, so `runnable` and `suite` are both `undefined` for every worker-forwarded hook failure. `buildXml`'s `groupBySuite` then keys each failure by itself (`test.parent || test`), and the suite name falls back to the literal `'Tests'`:

```js
const suiteName = (suite && suite.title) || 'Tests'
```

## Fix

`Hook.simplify()` now also serializes the suite identity from `this.runnable.parent` (the same object the main-thread path already reads successfully):

```js
simplify() {
  const suite = this.runnable?.parent
  return {
    hookName: this.hookName,
    title: this.title,
    error: this.err ? serializeError(this.err) : null,
    suiteTitle: suite?.title || null,
    suiteFile: suite?.file || null,
    suiteTags: suite?.tags || [],
  }
}
```

`junitReporter.js`'s listener falls back to these fields when `hook.ctx` is absent, reusing one synthetic suite object per `title+file` pair so multiple failures from the same worker-reported suite still group into a single `<testsuite>` instead of one per failure:

```js
let suite = runnable && runnable.parent
if (!suite && hook.suiteTitle) {
  const key = `${hook.suiteTitle} ${hook.suiteFile || ''}`
  if (!workerSuiteByKey.has(key)) {
    workerSuiteByKey.set(key, { title: hook.suiteTitle, file: hook.suiteFile, tags: hook.suiteTags })
  }
  suite = workerSuiteByKey.get(key)
}
```

## Reproduction steps

1. `tests/My_test.js`:
   ```js
   Feature('My');
   BeforeSuite(async () => { throw new Error('Before Suite Error'); });
   Scenario('test something', () => {});
   AfterSuite(async () => { throw new Error('After Suite Error'); });
   ```
2. `codecept.conf.js` with `junitReporter: { enabled: true }`
3. Run: `node bin/codecept.js run-workers 2 --config codecept.conf.js`
4. Inspect `output/report.xml`

Before this change:
```
testsuite name="Tests"
testsuite name="Tests"
```

After:
```
testsuite name="My"
```
(single element, both failures grouped inside)

## Edge cases verified

All run against the actual `run-workers` command, not simulated.

| Scenario | Before | After |
|---|---|---|
| BeforeSuite + AfterSuite fail, same suite, run-workers | 2x `<testsuite name="Tests">`, one testcase each | 1x `<testsuite name="My">`, both testcases inside |
| Two different suites each with a failing BeforeSuite/AfterSuite, run-workers | not tested pre-fix (same "Tests" collapse issue) | each suite gets its own correctly-named `<testsuite>`; no cross-suite mixing |
| Per-test `Before` hook failure (not BeforeSuite/AfterSuite), run-workers | unaffected (different hookName, filtered out before this code path) | unaffected, same as before |
| Plain `Scenario` failure (no hook failure at all), run-workers | unaffected (goes through `result.tests`, not `hookFailures`) | unaffected, same as before |
| Main-thread (no `run-workers`) BeforeSuite/AfterSuite failure | correct suite name (existing `hook.ctx` path) | unchanged — this path doesn't touch the new fallback at all |

Existing test suite: `mocha test/unit/junitReporter_test.js` — 10/10 passing, including the two hook-failure tests already in that file.
Full suite: `mocha test/unit` — 761 passing, 11 pending, 0 failing.
`mocha test/runner` — 275 passing, 2 pending, 0 failing.

## Out of scope

`lib/listener/result.js` also listens on `event.hook.failed` but only for stats counting (`failedHooks`), unaffected by this change. The other known limitation mentioned in #5657 (`_beforeSuite`/`_afterSuite` helper-level failures via `suiteSetup`/`suiteTeardown` not emitting `event.hook.failed` at all) is a separate, pre-existing gap not touched here.

### DIFF
--- a/lib/mocha/hooks.js
+++ b/lib/mocha/hooks.js
@@ -33,12 +33,22 @@ class Hook {
   }
 
   simplify() {
+    // this.runnable (context.ctx.test) is the hook's own runnable; its
+    // .parent is the real Mocha Suite that owns this hook. Included here so
+    // run-workers can forward suite identity to the main process, where
+    // listeners (e.g. junitReporter) have no other way to recover it — the
+    // worker's live Suite/ctx objects aren't serializable across the thread
+    // boundary, only these plain fields are.
+    const suite = this.runnable?.parent
     return {
       hookName: this.hookName,
       title: this.title,
       // test: this.test ? serializeTest(this.test) : null,
       // suite: this.suite ? serializeSuite(this.suite) : null,
       error: this.err ? serializeError(this.err) : null,
+      suiteTitle: suite?.title || null,
+      suiteFile: suite?.file || null,
+      suiteTags: suite?.tags || [],
     }
   }
 
--- a/lib/plugin/junitReporter.js
+++ b/lib/plugin/junitReporter.js
@@ -67,13 +67,29 @@ export default function (config = {}) {
 
   let written = false
   const hookFailures = []
+  // groupBySuite() (below) groups by object identity, not by value — reused
+  // across BeforeSuite/AfterSuite failures from the same worker-forwarded
+  // suite so they land in one <testsuite>, not one per failure.
+  const workerSuiteByKey = new Map()
 
   event.dispatcher.on(event.hook.failed, hook => {
     if (!hook || !['BeforeSuite', 'AfterSuite'].includes(hook.hookName)) return
     const err = hook.err || hook.error
     if (!err) return
     const runnable = hook.ctx && hook.ctx.test
-    const suite = runnable && runnable.parent
+    // Under run-workers, hook.ctx is absent — the failure arrives as the
+    // plain object from Hook.simplify() in the main process instead of a
+    // live Hook instance. Fall back to the suiteTitle/suiteFile/suiteTags
+    // simplify() carries across the worker boundary so the failure still
+    // groups under its real suite instead of the "Tests" fallback.
+    let suite = runnable && runnable.parent
+    if (!suite && hook.suiteTitle) {
+      const key = `${hook.suiteTitle} ${hook.suiteFile || ''}`
+      if (!workerSuiteByKey.has(key)) {
+        workerSuiteByKey.set(key, { title: hook.suiteTitle, file: hook.suiteFile, tags: hook.suiteTags })
+      }
+      suite = workerSuiteByKey.get(key)
+    }
     hookFailures.push({
       title: hook.title || `${hook.hookName} hook failed`,
       state: 'failed',
--- a/test/data/sandbox/codecept.workers-junit.conf.js
+++ b/test/data/sandbox/codecept.workers-junit.conf.js
@@ -1,0 +1,20 @@
+export const config = {
+  tests: './workers-junit-suite-hooks/*.js',
+  timeout: 10000,
+  output: './output',
+  helpers: {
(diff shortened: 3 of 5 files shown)


kapil971390:
Real terminal screenshots of the before/after run-workers reproduction (not simulated) will be attached here manually.
After:
codeceptjs-junitReporter-AFTER-worker-suite-name
Before:
codeceptjs-junitReporter-BEFORE-worker-suite-name

DenysKuchma (changes requested):
Add run-workers JUnit regression test

kapil971390:
Added a run-workers regression test in the latest commit.

It spawns a real `run-workers` process (not a synthetic event emission) against a dedicated fixture (`test/data/sandbox/workers-junit-suite-hooks/junit_suite_hooks.worker.js` — `BeforeSuite`/`AfterSuite` both throwing) with `junitReporter` enabled, then parses the generated `report.xml` and asserts both hook failures land under the real suite name (`JunitWorkerSuiteHooks`) in a single `<testsuite>`, not split across two `<testsuite name="Tests">` elements.

Verified it actually catches the regression: reverted to the pre-fix `junitReporter.js`/`hooks.js` locally and the new test fails with a clear assertion (`Expected length: 1, Received length: 2`, suite name `"Tests"`) rather than passing or hanging.

Note: the new fixture lives in its own `workers-junit-suite-hooks/` directory rather than the shared `workers/` directory — several other tests in `run_workers_test.js` glob `workers/*.js` against the base config, and adding a new failing suite there changed their expected pass/fail counts (caught this by running the full `run_workers_test.js` suite before pushing — all 30 tests pass now, no regressions).

Existing suites still pass: `mocha test/unit/junitReporter_test.js` (10/10) and the full `mocha test/runner/run_workers_test.js` (30/30, including the new test).

DenysKuchma (approved):

DavertMik:
Thank you!

PR #5676, opened 2026-08-04
fix: workerInitializationDelay is not hardcoded

kobenguyent:
The standard solutions to the "Thundering Herd" problem with concurrent browser initialization. The previous implementation had a couple of architectural flaws:                                                    
                                                                                                                                                                                                                                      
  1. Massive CPU Spikes: By instantly spinning up 25 Node.js workers inside a for loop in the master process, the system had to simultaneously create 25 isolated V8 instances and start parsing files for all of them at the exact   
  same millisecond. This causes a massive CPU usage spike which can starve the main process and actually delay everything downstream.                                                                                                 
  2. Artificial Delay in Worker: Relying on the workerIndex * delay inside the worker script is generally seen as a band-aid. The V8 instances are already fighting for resources while they evaluate their file imports, only to hit 
  an arbitrary setTimeout when they reach the code execution phase.                                                                                                                                                                   
                                                                                                                                                                                                                                      
  ### A Better Implementation: Master-Coordinated Staggering                                                                                                                                                                          
                                                                                                                                                                                                                                      
  The standard practice in large-scale concurrent runners (like Playwright test and Jest worker-pools) is to stagger the creation of the worker threads themselves in the master process or use a bounded connection pool.            
                                                                                                                                                                                                                                      
 by removing the artificial delay logic from the worker file altogether and instead moved a 200ms delay directly into the worker-creation loop inside lib/workers.js.                                          
  • The CPU spike from spawning V8 instances is drastically smoothed out.                                                                                                                                                             
                                                                                                                                                                                                                                      
  Why this is better:                                                                                                                                                                                                                 
                                                                                                                                                                                                                                      
  • The master process now creates the first worker, waits 200ms, then creates the second worker, and so on.                                                                                                                          
  • The browser initialization is naturally staggered because each worker starts its entire lifecycle (including imports and setups) exactly 200ms behind the previous one.                                                           
  • You no longer have hacky math based on workerIndex inside your test runner script.                                                                                                                                                
                                                                                                                                                                                                                                      
Users can now specify workerInitializationDelay: <number> in their codecept.conf.js root configuration. If they omit it, it defaults to 200ms, and setting it to 0 will disable the delay entirely.                                 
                                                                                                                                                                                                                                      
  Here's an example of how a user would use this in their codecept.conf.js:                                                                                                                                                           
                                                                                                                                                                                                                                      
```
    export const config = {                                                                                                                                                                                                           
      tests: './tests/*_test.js',                                                                                                                                                                                                     
      output: './output',                                                                                                                                                                                                             
      // Custom initialization delay between spinning up workers (in milliseconds)                                                                                                                                                    
      workerInitializationDelay: 500,                                                                                                                                                                                                 
                                                                                                                                                                                                                                      
      helpers: {                                                                                                                                                                                                                      
        Playwright: {                                                                                                                                                                                                                 
          url: 'http://localhost',                                                                                                                                                                                                    
          show: false,                                                                                                                                                                                                                
          browser: 'chromium'                                                                                                                                                                                                         
        }                                                                                                                                                                                                                             
      },                                                                                                                                                                                                                              
      // ...                                                                                                                                                                                                                          
    }     
```       

## Type of change

- [ ] :fire: Breaking changes
- [ ] :bug: Bug fix
- [ ] 🧹 Chore
- [ ] :hotsprings: Hot fix
- [ ] :nail_care: Polish code

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -45,6 +45,7 @@ export const config = {
 - `timeout` — default per-test timeout in seconds; a test is killed if it stops responding.
 - `mocha` — [Mocha options](https://mochajs.org/#configuring-mocha-nodejs), including extra reporters. See [Reporters](/reports).
 - `workerInitializationDelay` — delay in milliseconds between spinning up parallel workers to prevent CPU spikes and stagger browser startup. Defaults to `200`. Set to `0` to disable.
+- `workerInitializationMaxDelay` — maximum total delay (in milliseconds) for worker initialization staggering. Defaults to `10000` (10 s). Set to `0` to disable capping.
 
 **BDD**
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -44,6 +44,7 @@ export const config = {
 
 - `timeout` — default per-test timeout in seconds; a test is killed if it stops responding.
 - `mocha` — [Mocha options](https://mochajs.org/#configuring-mocha-nodejs), including extra reporters. See [Reporters](/reports).
+- `workerInitializationDelay` — delay in milliseconds between spinning up parallel workers to prevent CPU spikes and stagger browser startup. Defaults to `200`. Set to `0` to disable.
 
 **BDD**
 
--- a/docs/parallel.md
+++ b/docs/parallel.md
@@ -22,6 +22,8 @@ npx codeceptjs run-workers 4
 
 Steps are not streamed to the console in this mode — output from separate threads can't be interleaved cleanly. While workers run, CodeceptJS sets `process.env.RUNS_WITH_WORKERS=true`, so plugins and helpers can branch on it. All `run` options work here too: `--grep "@smoke"`, `-c codecept.conf.js`, `--debug`, and the rest.
 
+By default, workers are created with a staggered delay of 200ms to prevent CPU spikes and stagger browser initializations. You can adjust this via `workerInitializationDelay` in your configuration.
+
 ### Distribution strategies
 
 `--by` controls how tests spread across workers:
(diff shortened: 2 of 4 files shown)


DenysKuchma:
@kobenguyent One thing still worth adding: the spawn stagger is still linear and unbounded — (workers - 1) * delay, so at very high worker counts the original "grows without bound" complaint returns (just 10× milder). Could cap the total stagger window, e.g.:
```
 const maxStagger = this.codecept.config.workerInitializationMaxDelay ?? 10000
 const effectiveDelay = staggerDelay > 0
     ? Math.min(staggerDelay, Math.floor(maxStagger / Math.max(1, this.workers.length - 1)))
     : 0
```
That spreads all workers across at most maxStagger (default 10s) regardless of count, while preserving the per-launch spacing for normal worker counts.

kobenguyent:
> @kobenguyent One thing still worth adding: the spawn stagger is still linear and unbounded — (workers - 1) * delay, so at very high worker counts the original "grows without bound" complaint returns (just 10× milder). Could cap the total stagger window, e.g.:
> 
> ```
>  const maxStagger = this.codecept.config.workerInitializationMaxDelay ?? 10000
>  const effectiveDelay = staggerDelay > 0
>      ? Math.min(staggerDelay, Math.floor(maxStagger / Math.max(1, this.workers.length - 1)))
>      : 0
> ```
> 
> That spreads all workers across at most maxStagger (default 10s) regardless of count, while preserving the per-launch spacing for normal worker counts.

thanks @DenysKuchma . Updated.

DenysKuchma (approved):

PR #5683, opened 2026-08-07
fix(junitReporter): stamp suites with their real start time

### LINKED ISSUE #5668: junitReporter: per-<testsuite> timestamp uses report-write time, not suite-start time
**Describe the bug**

In the JUnit XML produced by `junitReporter`, the root `<testsuites timestamp="...">` is correct (the actual run-start time), but the nested `<testsuite timestamp="...">` for each suite is wrong — it holds the moment the XML file was written (i.e. after that suite, including `AfterSuite`, finished), not when the suite started.

**Root cause**

`lib/plugin/junitReporter.js`:

```js
setAttr(root, 'timestamp', toIso(result.stats && result.stats.start))       // correct: Mocha's real run-start stat
...
setAttr(suiteEl, 'timestamp', toIso(suite && suite.startedAt))              // suite.startedAt is never set anywhere
```

```js
function toIso(value) {
  const date = value ? new Date(value) : new Date()
  return Number.isNaN(date.getTime()) ? new Date().toISOString() : date.toISOString()
}
```

`suite.startedAt` doesn't exist on Mocha's `Suite` — I grepped the whole `codeceptjs` package and `startedAt` is only ever set on individual **test** objects (`lib/listener/steps.js:20`), never on a `Suite`. So `suite && suite.startedAt` is always `undefined`, `toIso()` falls through to its `new Date()` fallback, and the `<testsuite>` gets stamped with "now" at write time — effectively the run's *end* time, not its start.

**To Reproduce**

1. Run any suite with `-p junitReporter` (or the `junitReporter` plugin enabled in config).
2. Open the generated `report.xml`.
3. Compare the root `<testsuites timestamp>` against the nested `<testsuite timestamp>` — the nested one matches the file's mtime / run-finish time instead of run-start time.

Example from a real run (root suite ran 08:09:58–08:10:20 local / 06:09:58–06:10:20 UTC):

```xml
<testsuites name="CodeceptJS" tests="5" ... timestamp="2026-07-30T06:09:58.974Z">
  <testsuite name="LeaveRequest" ... timestamp="2026-07-30T06:10:20.079Z" ...>
```

The root timestamp (06:09:58.974Z) is right. The suite timestamp (06:10:20.079Z) matches `report.xml`'s file mtime (`08:10:20.079420438 +0200` local), i.e. write time, not start time — a ~22s discrepancy in this run.

**Expected behavior**

The `<testsuite>` `timestamp` attribute should reflect when that suite actually started (e.g. captured via `event.suite.before`, or derived from the earliest `test.startedAt`/`test.opts.startedAt` among the suite's tests), not the report-write time.

**Environment**
- CodeceptJS: 4.0.9
- Node: 24.18.0
- OS: Linux 6.17, Ubuntu 24.04.4 LTS

**Related**

Same underlying failure pattern — "now" captured at report-build/write time gets mislabeled as a start timestamp — also found and reported separately in `@testomatio/reporter`'s HTML report ("Started At" field): testomatio/reporter#903

---
🤖 Filed with the help of Claude Sonnet 5 (Anthropic)

luantaraschi:

Resolves #5668.

`junitReporter` writes each `<testsuite timestamp>` from `suite.startedAt` (`lib/plugin/junitReporter.js:135`), but nothing ever set that field. Confirming @mirao's read: `startedAt` is assigned in exactly one place, `lib/listener/steps.js:20`, and only to individual tests. Mocha's `Suite` does not carry it. So `toIso()` always fell through to its `new Date()` fallback and every suite got stamped with the moment the XML was serialized, which is after that suite and its `AfterSuite` had already finished.

The steps listener now stamps the suite on `event.suite.before`, mirroring what it already does for tests two lines below.

One thing worth flagging: the existing `junitReporter_test.js` builds its fixtures with `startedAt: Date.now()` on each suite object, so the tests were exercising a field that production never produced. That is why the reporter's own suite passed while the output was wrong. I left those fixtures alone since they now describe reality, and put the new coverage on the listener instead.

Scope note: this covers the in-process run. Under `run-workers`, `lib/mocha/test.js` serializes a test's parent down to `{ title }` and drops `startedAt` from the test payload as well, so a suite timestamp cannot cross that boundary without changing the worker payload. That looked like a separate change, so I left it out rather than widen this PR.

## Type of change

- [x] :bug: Bug fix

## Checklist:

- [x] Tests have been added

Two tests in `test/unit/listener/steps_suite_started_at_test.js`. They fail on the commit before this change with `expected undefined to be a number`, and pass with it.

Full unit suite on Windows: 758 passing / 13 failing before, 760 passing / 11 failing after. The 11 remaining failures are pre-existing path assertions that expect POSIX-style paths and see a `C:` drive letter (`utils_test.js`, `utils/trace_test.js`); they are identical with and without this change.

### DIFF
--- a/lib/listener/steps.js
+++ b/lib/listener/steps.js
@@ -16,6 +16,14 @@ const EXCLUDED_SESSIONS = ['tryTo', 'hopeThat']
  * Register steps inside tests
  */
 export default function () {
+  // Mocha's Suite has no start timestamp of its own, and junitReporter reads
+  // `suite.startedAt` for each `<testsuite timestamp>`. Without this the
+  // reporter falls back to `new Date()` at write time, stamping every suite
+  // with the moment the XML was serialized. (#5668)
+  event.dispatcher.on(event.suite.before, suite => {
+    suite.startedAt = +new Date()
+  })
+
   event.dispatcher.on(event.test.before, test => {
     test.startedAt = +new Date()
   })
--- a/test/unit/listener/steps_suite_started_at_test.js
+++ b/test/unit/listener/steps_suite_started_at_test.js
@@ -1,0 +1,45 @@
+import { expect } from 'chai'
+import event from '../../../lib/event.js'
+import recorder from '../../../lib/recorder.js'
+
+import stepsListener from '../../../lib/listener/steps.js'
+
+// junitReporter reads `suite.startedAt` for each `<testsuite timestamp>`.
+// Mocha never sets it, so before this listener existed the reporter fell back
+// to `new Date()` at write time. (#5668)
+describe('Steps Listener - suite.startedAt', () => {
+  beforeEach(() => {
+    recorder.reset()
+    recorder.start()
+    event.cleanDispatcher()
+    stepsListener()
+  })
+
+  afterEach(() => {
+    event.cleanDispatcher()
+    recorder.reset()
+  })
+
+  it('stamps the suite when it starts', () => {
+    const suite = { title: 'Login' }
+    const before = Date.now()
+
+    event.emit(event.suite.before, suite)
+
+    expect(suite.startedAt).to.be.a('number')
+    expect(suite.startedAt).to.be.at.least(before)
+    expect(suite.startedAt).to.be.at.most(Date.now())
+  })
+
+  it('stamps each suite independently', () => {
+    const first = { title: 'Login' }
+    const second = { title: 'Dashboard' }
+
+    event.emit(event.suite.before, first)
+    event.emit(event.suite.before, second)
+
+    expect(first.startedAt).to.be.a('number')
+    expect(second.startedAt).to.be.a('number')
+    expect(second.startedAt).to.be.at.least(first.startedAt)
+  })
+})


luantaraschi:
The red `appium` job on this PR was not about the change, and it should be gone now. Recording what it was in case it comes up on other fork PRs.

The last CI here ran on 2026-08-07 and the Android job failed at the Sauce Labs handshake:

```
{"detail":"Authorization failed"}
WebDriverError: failed serving request POST /wd/hub/session: Unauthorized
Appium "before all" hook in "Appium"  Can't connect to WebDriver.
```

That is a missing credential, not a broken test. Fork pull requests do not receive `SAUCE_USERNAME` / `SAUCE_ACCESS_KEY`, and neither do Dependabot ones, so the job cannot pass from either. Grouping the last hundred Android runs by trigger shows the split cleanly: pushes to `4.x` and pull requests whose head branch lives in the repo pass, while every Dependabot pull request in the same window fails at the same line.

You already fixed this on 2026-08-12 in `6a5ce8d3` / `eb1bcdc5`, "fix(ci): dont run appium tests for PRs from forked repos", which added the `github.event.pull_request.head.repo.full_name == github.repository` guard. This PR's run predated it by five days, which is the whole reason it looked red.

So I pressed Update branch to pick up the guard rather than leave a red X sitting on an approved PR. The Android job is skipped on the new head, as intended. The fresh runs are sitting at `action_required`, which is the fork-contributor gate, so they need an approval from a maintainer before they will report.

luantaraschi:
`Unit tests (26.x)` went red after I pressed Update branch here this morning, and I do not think it belongs to this change. Here is what I checked, including the one way this PR could plausibly have caused it.

The failure is a single assertion:

```
✖ els #element should fail on timeout if timeout is set
  expected 'should have thrown error' to include 'was interrupted on timeout 10ms'
```

That lives in `test/unit/els_test.js` and the message comes from `lib/recorder.js`. A 10 ms budget on a loaded runner is a thin margin.

**The same code passed this job a week ago.** Run `31176426017` on this branch, sha `5205311b`, was green on `Unit tests (26.x)` on August 7. The only thing that changed since is the `4.x` merge from Update branch.

**The PR that took the same merge in the same minute is green.** #5684 got the identical `4.x` merge and its `Unit tests (26.x)` passed.

**The one real suspicion, checked.** My new test calls `event.cleanDispatcher()` and `recorder.reset()`, which are global, so a file-order interaction with `els_test.js` was worth ruling out rather than waving away. Mocha loads `test/unit/els_test.js` at position 18 of 64 and `test/unit/listener/steps_suite_started_at_test.js` at position 28, so my file runs after that suite has already finished. It cannot reach it.

**Locally**, on Windows with Node 22.20.0:

| tree | passing | failing |
| --- | --- | --- |
| this PR, `45a56b4c` | 760 | 11 |
| `4.x` control | 758 | 11 |

The same eleven both times, all of them Windows-specific (workers, sharding, screencast, trace, MCP), and `els #element` is not among them either way. The +2 is exactly the two tests this PR adds.

So a re-run of that job should clear it. Sorry for the noise, the Update branch was to pick up `eb1bcdc5` so the fork-side Appium job would be skipped, which it now is.

luantaraschi:
The red job is `Unit tests (26.x)`, failing on `els #element should fail on timeout if timeout is set` (`test/unit/els_test.js:78`). `Unit tests (20.x)` passed in the same run.

That test races a 10ms step timeout against a 1000ms sleep, and this change does not go near it. It adds one `event.suite.before` handler, and `els_test.js` never emits suite events. In that run els_test reported before `test/unit/listener/steps_suite_started_at_test.js`, so the new file could not have reached it, and its setup and teardown copy `steps_issue_4619_test.js` sitting next to it.

I could not reproduce the failure. On Node 26.7.0 the file passes on its own, 15 of 15, and the full `test/unit` run gives 760 passing and 11 failing. Checking out `4.x` at eb1bcdc and running the same command gives 758 passing and the same 11 failures, which are Windows path expectations in `utils_test.js` and `utils/trace_test.js`. Node 22.20.0 gives the same set.

@kobenguyent could you re-run that job when you get a chance?

DavertMik:
Thank you so much!

PR #5684, opened 2026-08-07
fix(hooks): emit hook.failed when a helper suite hook throws

### LINKED ISSUE #5660: 4.x: junitReporter still doesn't record helper _beforeSuite()/_afterSuite() failures (follow-up to #5645)
Follow-up to #5645 and the fix in 5b2517c6bc830c78f22609d3582ee80bdd96d339.

That fix listens on `event.hook.failed`, which is only emitted by `fireHook()` in `lib/mocha/asyncWrapper.js`, and `fireHook()` is only called from the `injected()` wrapper used for the test-file-defined `BeforeSuite()` / `AfterSuite()` / `Before()` / `After()` functions (registered in `lib/mocha/ui.js`).

A **custom helper's** `_beforeSuite()` / `_afterSuite()` lifecycle methods run through a different mocha hook — `suite.beforeAll('codeceptjs.beforeSuite', suiteSetup(suite))` (`lib/mocha/ui.js:106`), which calls `runAsyncHelpersHook('_beforeSuite', ...)` (`lib/listener/helpers.js:32`). Looking at `suiteSetup()` / `suiteTeardown()` in `asyncWrapper.js`, on failure they call `doneFn(err)` directly and never call `fireHook()` — so `event.hook.failed` is never emitted for helper lifecycle failures, and `junitReporter` still produces an empty/incomplete report for these.

### Repro

```ts
// steps_file.ts / a custom helper
import Helper from "@codeceptjs/helper";

class MyHelper extends Helper {
  async _beforeSuite() {
    throw new Error("Simulated error in helper _beforeSuite");
  }
}
export = MyHelper;
```

```ts
// My_test.ts
Feature("My");

  I.amOnPage("https://example.com");
});
```

```ts
// codecept.conf.ts
  tests: "./*_test.ts",
  output: "./output",
  helpers: {
    Playwright: { browser: "chromium", url: "http://localhost", show: false },
    MyHelper: { require: "./my_helper.ts" },
  },
  noGlobals: true,
  plugins: { junitReporter: { enabled: true } },
  name: "my",
  require: ["tsx/cjs"],
};
```

Running `codeceptjs run --verbose` fails the suite (`"before all" hook: codeceptjs.beforeSuite for "..."`), but `output/report.xml` comes out as:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites name="CodeceptJS" tests="0" failures="0" skipped="0" errors="0" time="0.000" .../>
```

Same empty-report symptom as #5645, just triggered from a helper's lifecycle hook instead of the test file's `BeforeSuite()`.

### Suggested fix

We worked around it by listening on Mocha's own `runner.on('fail', ...)` event instead of (or in addition to) `event.hook.failed`. `runner.on('fail')` fires for **every** suite-level hook failure regardless of origin (test-file hooks and helper lifecycle hooks alike), so a single title-based filter (`/^"(before all|after all)" hook:/`) covers both cases without duplicating the entries already covered by `event.hook.failed`. Happy to share the patch we're running if useful for a PR.

luantaraschi:

Resolves #5660.

@mirao's trace is correct and I followed it through the code. `event.hook.failed` is only emitted by `fireHook()`, which is only reached from the `injected()` wrapper used for test-file-defined `BeforeSuite()` / `AfterSuite()`. A custom helper's `_beforeSuite()` / `_afterSuite()` takes a different route:

`lib/mocha/ui.js:106` registers `suite.beforeAll('codeceptjs.beforeSuite', suiteSetup(suite))` → `suiteSetup` emits `event.suite.before` → `lib/listener/helpers.js:30` queues `runAsyncHelpersHook('_beforeSuite', ...)` on the recorder. When that throws, it surfaces in `suiteSetup`'s `recorder.errHandler`, which called `doneFn(err)` and nothing else. No `hook.failed`, so `junitReporter`'s listener never fired and the report stayed empty.

Both error handlers now emit the matching hook object before calling done. I passed the mocha suite straight through, which is what `fireHook()` already does, so `hook.ctx.test` and `hook.hookName` come out the same shape junitReporter already consumes (`['BeforeSuite', 'AfterSuite'].includes(hook.hookName)`).

I did not reuse `fireHook()` here: it derives the hook kind from `suite.ctx?.test?.title?.match(/"([^"]*)"/)[1]`, and these hooks are titled `codeceptjs.beforeSuite`, which has no quoted segment, so that match returns null and the indexing throws.

## Type of change

- [x] :bug: Bug fix

## Checklist:

- [x] Tests have been added

Two tests appended to `test/unit/mocha/asyncWrapper_test.js`, one per hook. They queue a throwing helper method the same way `runAsyncHelpersHook` does and assert both that `done` still receives the error and that `hook.failed` carries the right `hookName`. Both fail on the commit before this change with `hook.failed was emitted`.

Full unit suite on Windows: 758 passing / 13 failing before, 760 passing / 11 failing after. The 11 remaining are pre-existing path assertions that expect POSIX paths and see a `C:` drive letter (`utils_test.js`, `utils/trace_test.js`), identical with and without this change.

Related: #5683 fixes the other half of the junit reporting gap @mirao reported, the suite timestamp.

### DIFF
--- a/lib/mocha/asyncWrapper.js
+++ b/lib/mocha/asyncWrapper.js
@@ -4,7 +4,7 @@ import recorder from '../recorder.js'
 import assertThrown from '../assert/throws.js'
 import { ucfirst, isAsyncFunction } from '../utils.js'
 import { getInjectedArguments } from './inject.js'
-import { fireHook } from './hooks.js'
+import { fireHook, BeforeSuiteHook, AfterSuiteHook } from './hooks.js'
 
 const injectHook = function (inject, suite) {
   try {
@@ -232,6 +232,10 @@ export function suiteSetup(suite) {
 
     // Set up error handler for suite setup
     recorder.errHandler(err => {
+      // A helper's `_beforeSuite()` runs through this hook, not through the
+      // `injected()` wrapper, so nothing here used to emit `hook.failed` and
+      // reporters listening for it never saw the failure. (#5660)
+      event.emit(event.hook.failed, new BeforeSuiteHook(suite, err))
       doneFn(err)
     })
 
@@ -254,6 +258,8 @@ export function suiteTeardown(suite) {
 
     // Set up error handler for suite teardown
     recorder.errHandler(err => {
+      // Same for a helper's `_afterSuite()`. (#5660)
+      event.emit(event.hook.failed, new AfterSuiteHook(suite, err))
       doneFn(err)
     })
 
--- a/test/unit/mocha/asyncWrapper_test.js
+++ b/test/unit/mocha/asyncWrapper_test.js
@@ -269,4 +269,48 @@ describe('AsyncWrapper', () => {
       expect(arg, 'done called with no error').to.be.undefined
     })
   })
+  describe('helper lifecycle hook failures (#5660)', () => {
+    beforeEach(() => recorder.start())
+
+    // A helper's _beforeSuite()/_afterSuite() is queued on the recorder from an
+    // event.suite.before/after listener (lib/listener/helpers.js), not through
+    // the injected() wrapper, so it lands in suiteSetup/suiteTeardown's
+    // errHandler rather than in the path that fires hook.failed.
+    function queueFailingHelperHook(evt, message) {
+      event.dispatcher.on(evt, () => {
+        recorder.add(`hook MyHelper.${message}()`, () => {
+          throw new Error(message)
+        })
+        recorder.catch()
+      })
+    }
+
+    it('suiteSetup(): a failing helper _beforeSuite emits hook.failed', async () => {
+      const failed = sinon.spy()
+      event.dispatcher.on(event.hook.failed, failed)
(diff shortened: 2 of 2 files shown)


kobenguyent (approved):

luantaraschi:
A note on the red X that used to sit on this PR, since it was never about the change.

The `appium` job here ran on 2026-08-07 and stopped at the Sauce Labs handshake with `{"detail":"Authorization failed"}`. Fork pull requests do not receive `SAUCE_USERNAME` or `SAUCE_ACCESS_KEY`, so it could not have passed from this side.

You had already handled that on 2026-08-12 in `eb1bcdc5`, which skips the Appium job when the head branch lives outside the repository. I pressed Update branch to pick the guard up rather than leave a red X on an approved PR. On the new head the job is skipped and the rest of the matrix is green.

DavertMik:
Thanks!

PR #5685, opened 2026-08-07
fix(typescript): map included .ts files in step output

### LINKED ISSUE #5675: Scenario Steps show .temp.mjs instead of .ts for TypeScript include modules

Scenario Steps / failure output for actions called from TypeScript `include` modules (page objects, custom steps) should show the original `.ts` path, as the migration guide says:

> [Error stack traces point at `.ts` source lines, not transpiled output.](https://codecept.io/migration-4#:~:text=Error%20stack%20traces%20point%20at%20.ts%20source%20lines%2C%20not%20transpiled%20output.)

Related: #5362 (closed), #5403. Still reproduces on 4.0.8 and 4.1.0.

#### What do you get instead?

Test files show `.ts` correctly. Steps from `include`d `.ts` files show deleted sibling `.temp.mjs` paths (and wrong line numbers):

```bash
✔ I.click("...") at Object.somePageMethod (file://./pages/home.page.18611.jdt12cwb.temp.mjs:80:11)
✖ I.click("missing") at Test.<anonymous> (./tests/home_test.ts:5:5)
```

```ts
// pages/home.page.ts
const { I } = inject();
module.exports = {
  open() {
    I.amOnPage('/');
    I.click('this does not exist');
  },
};

// home_test.ts
Feature('temp paths');
Scenario('fail from page object', ({ homePage }) => {
  homePage.open();
});
```

### Details

* CodeceptJS version: 4.0.8 (also verified still present in 4.1.0)
* NodeJS Version: 26.x
* Operating System: macOS
* playwright version: (Playwright helper; unrelated to the path bug)
* Configuration file:

```ts
exports.config = {
  tests: './tests/**/*.ts',
  helpers: { Playwright: { url: 'http://localhost', browser: 'chromium' } },
  include: { homePage: './pages/home.page.ts' },
  require: ['tsx/cjs'],
};
```

luantaraschi:

Resolves #5675.

@djyarber's observation that test files map correctly but included page objects do not comes down to two adjacent blocks in `lib/container.js` that do almost the same thing and disagree about where the result goes.

The helper block, around line 470, merges the transpile mapping into **`store.tsFileMapping`**. The include/support block, around line 889, merges the same shape of mapping into **`container.tsFileMapping`** only.

`Step.line()` in `lib/step/base.js:156` reads `store.tsFileMapping`. So a step whose stack frame points into an included `.ts` module has no entry to match and keeps the `.temp.mjs` path, which by then has been deleted, hence the paths in the report.

Error stacks were never affected, which is why the migration guide's promise holds for failures: `fixErrorStack()` is handed the mapping object directly by the caller rather than reading it from `store`.

The include block now merges into `store` as well, mirroring the helper block.

## Type of change

- [x] :bug: Bug fix

## Checklist:

- [ ] Tests have been added

I want to be straight about the missing test. Exercising this needs a container built from a config with a TypeScript `include`, transpiled for real, and I did not get a harness for that working that I would trust. `test/unit/utils/typescript_test.js` drives `transpileTypeScript` directly and never touches the container. A test asserting `Step.line()` maps a path when `store.tsFileMapping` already holds the entry would pass before and after this change, so it would prove nothing.

What I did verify is the asymmetry itself: both blocks receive the same `mapping` from `transpileTypeScript`, only one writes to `store`, and `store` is what the reader uses. If you point me at the right fixture or harness for a config-level include I will add the regression test.

Unit suite on Windows: 758 passing / 11 failing, unchanged by this PR. Those 11 are pre-existing path assertions that expect POSIX paths and see a `C:` drive letter (`utils_test.js`, `utils/trace_test.js`).

### DIFF
--- a/test/data/sandbox/typescript-step-paths/codecept.conf.js
+++ b/test/data/sandbox/typescript-step-paths/codecept.conf.js
@@ -1,0 +1,13 @@
+export const config = {
+  tests: './tests/*Test.ts',
+  helpers: {
+    FakeHelper: {
+      require: './fakeHelper.js',
+    },
+  },
+  include: {
+    fooPage: './pages/fooPage.ts',
+  },
+  require: ['tsx/cjs'],
+  name: 'typescript-step-paths',
+}
--- a/test/runner/typescript_step_paths_test.js
+++ b/test/runner/typescript_step_paths_test.js
@@ -9,23 +9,25 @@ const runner = path.join(__dirname, '../../bin/codecept.js')
 const codeceptDir = path.join(__dirname, '../data/sandbox/typescript-step-paths')
 
 describe('TypeScript step paths', () => {
-  it('maps included page object steps back to their source file', done => {
-    execFile(
-      process.execPath,
-      [runner, 'run', '--config', path.join(codeceptDir, 'codecept.conf.js')],
-      { cwd: codeceptDir, env: { ...process.env, FORCE_COLOR: '0' } },
-      (err, stdout) => {
-        try {
-          expect(err).toBeTruthy()
-          expect(stdout).toContain('Scenario Steps:')
-          expect(stdout).toMatch(/at Object\.open \(\.\/pages\/fooPage\.ts:\d+:\d+\)/)
-          expect(stdout).not.toContain('.temp.mjs')
-          expect(stdout).not.toContain('file://./pages/fooPage.ts')
-          done()
-        } catch (error) {
-          done(error)
-        }
-      },
-    )
-  })
+  for (const configFile of ['codecept.conf.js', 'codecept.esm.conf.js']) {
+    it(`maps included page object steps back to their source file with ${configFile}`, done => {
+      execFile(
+        process.execPath,
+        [runner, 'run', '--config', path.join(codeceptDir, configFile)],
+        { cwd: codeceptDir, env: { ...process.env, FORCE_COLOR: '0' } },
+        (err, stdout) => {
+          try {
+            expect(err).toBeTruthy()
+            expect(stdout).toContain('Scenario Steps:')
+            expect(stdout).toMatch(/at Object\.open \(\.\/pages\/fooPage\.ts:\d+:\d+\)/)
+            expect(stdout).not.toContain('.temp.mjs')
+            expect(stdout).not.toContain('file://./pages/fooPage.ts')
+            done()
+          } catch (error) {
+            done(error)
+          }
+        },
+      )
+    })
+  }
 })
--- a/test/runner/typescript_step_paths_test.js
+++ b/test/runner/typescript_step_paths_test.js
@@ -1,0 +1,31 @@
+import { execFile } from 'child_process'
+import { expect } from 'expect'
+import path from 'path'
+import { fileURLToPath } from 'url'
+
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = path.dirname(__filename)
+const runner = path.join(__dirname, '../../bin/codecept.js')
+const codeceptDir = path.join(__dirname, '../data/sandbox/typescript-step-paths')
+
+describe('TypeScript step paths', () => {
+  it('maps included page object steps back to their source file', done => {
+    execFile(
+      process.execPath,
+      [runner, 'run', '--config', path.join(codeceptDir, 'codecept.conf.js')],
+      { cwd: codeceptDir, env: { ...process.env, FORCE_COLOR: '0' } },
+      (err, stdout) => {
+        try {
+          expect(err).toBeTruthy()
+          expect(stdout).toContain('Scenario Steps:')
+          expect(stdout).toMatch(/at Object\.open \(\.\/pages\/fooPage\.ts:\d+:\d+\)/)
+          expect(stdout).not.toContain('.temp.mjs')
+          expect(stdout).not.toContain('file://./pages/fooPage.ts')
+          done()
+        } catch (error) {
+          done(error)
+        }
+      },
+    )
+  })
+})
--- a/lib/container.js
+++ b/lib/container.js
@@ -892,10 +892,6 @@ async function loadSupportObject(modulePath, supportObjectName) {
           for (const [key, value] of mapping.entries()) {
             container.tsFileMapping.set(key, value)
           }
-          // Step.line() maps temp paths back through store, not through the
-          // container, so an include has to land there too. Without this a step
-          // that originates in an included .ts page object is printed with the
-          // deleted .temp.mjs sibling. (#5675)
           if (!store.tsFileMapping) {
             store.tsFileMapping = new Map()
           }
--- a/lib/container.js
+++ b/lib/container.js
@@ -892,6 +892,16 @@ async function loadSupportObject(modulePath, supportObjectName) {
           for (const [key, value] of mapping.entries()) {
             container.tsFileMapping.set(key, value)
           }
+          // Step.line() maps temp paths back through store, not through the
+          // container, so an include has to land there too. Without this a step
+          // that originates in an included .ts page object is printed with the
+          // deleted .temp.mjs sibling. (#5675)
+          if (!store.tsFileMapping) {
+            store.tsFileMapping = new Map()
+          }
+          for (const [key, value] of mapping.entries()) {
+            store.tsFileMapping.set(key, value)
+          }
         } catch (tsError) {
           throw new Error(`Failed to load TypeScript file ${importPath}: ${tsError.message}. Make sure 'typescript' package is installed.`)
         }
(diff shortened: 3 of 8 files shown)


djyarber:
Thanks for digging into this! Your read of the write/read mismatch is correct. Here’s a minimal browser-free config that reproduces it, with only this PR applied the Scenario Steps path is still wrong.

## Working Repro

Browser-free, no network, ~1.1s. CodeceptJS **4.1.0** (latest on npm) with the one-line change from this PR hand-applied to `lib/container.js`. Same result on 4.0.8.

`codecept.conf.js`:

```js
export const config = {
  tests: './tests/*Test.ts',
  helpers: { FakeHelper: { require: './fakeHelper.js' } },
  include: { fooPage: './pages/fooPage.ts' },
  require: ['tsx/cjs'],
  name: 'ts-step-paths-repro',
};
```

`fakeHelper.js` — avoids needing a browser at all:

```js
import Helper from 'codeceptjs/lib/helper';

export default class FakeHelper extends Helper {
  doThing(label) { return label; }
  failNow(message) { throw new Error(message); }
}
```

`pages/fooPage.ts`:

```ts
export {};
const { I } = inject();

export default {
  open() {
    I.doThing('called from a .ts page object under include');
  },
};
```

`tests/fooTest.ts` — the failing step is what makes the `Scenario Steps` block print:

```ts
Feature('ts step paths');

Scenario('reports page object step path', ({ I, fooPage }) => {
  fooPage.open();
  I.failNow('deliberate failure so Scenario Steps is printed');
});
```

## Result matrix

| `container.js` (this PR) | `step/base.js` ordering | Step output |
|---|---|---|
| stock | stock | ✖ `.temp.mjs` |
| **this PR** | stock | ✖ `.temp.mjs` |
| stock | fixed (below) | ✖ `.temp.mjs` |
| this PR | fixed (below) | ✔ `file://./pages/fooPage.ts` (`.ts`, but `file://` leftover) |

Neither change is sufficient alone, both are necessary.

## Defect

Even with this PR’s change applied (hand-patched into `lib/container.js` on `4.1.0`), Scenario Steps still report `.temp.mjs`. Instrumenting `Step.line()` shows why:

```
[DEBUG] line to match: "at Object.open (file://./pages/fooPage.74609.ryob11cp.temp.mjs:4:11)"
[DEBUG] mapping present: true size: 1
[DEBUG]   entry mjs: "/abs/path/to/pages/fooPage.74609.ryob11cp.temp.mjs" -> matched: false
```

The mapping entry now exists, your change works, but the lookup still misses.

`Step.line()` shortens absolute paths to `.` _before_ it looks them up in the mapping. The mapping’s keys are still absolute (`transpileTypeScript` stores them that way), so after shortening, `line.includes(mjsFile)` never finds a match.

Doing the mapping first, then shortening, seems to fix it:

```diff
   line() {
     const lines = this.stack.split('\n')
     if (lines[STACK_LINE]) {
-      let line = lines[STACK_LINE].trim()
-        .replace(store.codeceptDir || '', '.')
-        .trim()
+      let line = lines[STACK_LINE].trim()

-      // Map .temp.mjs back to original .ts files using container's tsFileMapping
+      // Map .temp.mjs back to original .ts files using container's tsFileMapping.
+      // The mapping holds absolute paths, so this must run before codeceptDir is
+      // shortened to '.', or the lookup can never match.
       const fileMapping = store.tsFileMapping
       if (line.includes('.temp.mjs') && fileMapping) {
         for (const [tsFile, mjsFile] of fileMapping.entries()) {
           if (line.includes(mjsFile)) {
             line = line.replace(mjsFile, tsFile)
             break
           }
         }
       }

-      return line
+      return line.replace(store.codeceptDir || '', '.').trim()
     }
     return ''
   }
```

Non-mapped step lines still shorten the same way, eg at `Test.<anonymous> (./tests/fooTest.ts:7:5)`.

Even when mapping works, include steps still show `file://./pages/fooPage.ts` rather than `./pages/fooPage.ts`, because shortening leaves the `file://` scheme.

## Line numbers are still wrong (separate issue)

#5675 also mentioned wrong line numbers, and that part is **not** fixed by either change. Green output reports `./pages/fooPage.ts:4:11`, but `I.doThing` is at line 6, column 9, the numbers are still the transpiled `.mjs`'s, off by the stripped `export {}` and blank line.

`fileMapping` is file-level only, so no path-substitution fix can recover line numbers. That needs line-preserving transpilation or real source maps. Worth splitting out rather than closing #5675 as fully fixed.

luantaraschi:
Thanks for the detailed repro and result matrix. You're right: this PR populates the mapping, but `Step.line()` shortens the absolute path before the lookup, so the absolute `.temp.mjs` key can never match.

I'll update this PR to apply the mapping before `codeceptDir` normalization, remove the leftover `file://` prefix, and add the browser-free regression you outlined. I'll keep the incorrect line numbers separate because file-level mapping cannot recover source positions. This PR should only claim to restore the original `.ts` path.

test/runner/typescript_step_paths_test.js
```diff
@@ -0,0 +1,31 @@
...
+      process.execPath,
+      [runner, 'run', '--config', path.join(codeceptDir, 'codecept.conf.js')],
+      { cwd: codeceptDir, env: { ...process.env, FORCE_COLOR: '0' } },
+      (err, stdout) => {
+        try {
+          expect(err).toBeTruthy()
+          expect(stdout).toContain('Scenario Steps:')
+          expect(stdout).toMatch(/at Object\.open \(\.\/pages\/fooPage\.ts:\d+:\d+\)/)
+          expect(stdout).not.toContain('.temp.mjs')
+          expect(stdout).not.toContain('file://./pages/fooPage.ts')
```
djyarber: _⛏️ nitpick (non-blocking)_

I think this is redundant with the `./pages/fooPage.ts` match above, that regex already rejects the `file://./` form

djyarber (approved):
Verified this fixes what I was seeing (unexpected file references)

One small comment suggestion left to consider

kobenguyent:
Out of curiosity, does require: ['tsx/esm'] resolve the issue quickly?

luantaraschi:
I tested that exact change against the previous PR head (`44cb8b70`) using the browser-free repro. It still reports the included step as `fooPage.<hash>.temp.mjs`, so `tsx/esm` alone does not fix the path mapping.

This matches the current behavior in `Codecept.requireModules`, which already replaces `tsx/cjs` with `tsx/esm`. I also ran the regression on the current PR head with the fixture explicitly set to `tsx/esm`, and it passes with the original `./pages/fooPage.ts` path. The mapping and normalization change is still needed.

test/data/sandbox/typescript-step-paths/codecept.conf.js
```diff
@@ -0,0 +1,13 @@
...
+  tests: './tests/*Test.ts',
+  helpers: {
+    FakeHelper: {
+      require: './fakeHelper.js',
+    },
+  },
+  include: {
+    fooPage: './pages/fooPage.ts',
+  },
+  require: ['tsx/cjs'],
```
kobenguyent, line 11: Perhaps we shall add another test configuration using tsx/ESM to increase our coverage?

kobenguyent:
Thanks for the quick response. So anyway we still need the fix.

test/data/sandbox/typescript-step-paths/codecept.conf.js
luantaraschi: Good suggestion. I added a second config that uses `require: ['tsx/esm']` directly and now run the same regression against both configs. Both cases pass locally (`2 passing`), and ESLint passes for the related test files.

kobenguyent (approved):

test/runner/typescript_step_paths_test.js
luantaraschi: You are right that the regex above already rejects the `file://` form on that line: it anchors on `(` immediately followed by `./pages/`, so `at Object.open (file://./pages/fooPage.ts:4:11)` cannot match it.

The one case the extra line still covers is a second occurrence. `toMatch` is satisfied by any match anywhere in stdout, so a run that printed one correct step line and one leftover `file://` line elsewhere in the Scenario Steps block would pass the regex and fail this. That is the shape I actually saw while working on it: the normalization was fixed for the mapped line first, and the unmapped ones still carried the scheme.

Happy to drop it if you would rather keep the test tight. It is a cheap belt for a failure mode I hit once, not something I feel strongly about.

djyarber: No strong feelings about it, you hit it and the test is already implemented, I would opt for leaving it and moving on! 🎉

luantaraschi: Great, leaving it as is then. Thanks for the review and for the repro that pinned down the mapping order.

DavertMik:
Seems good, thank you

DavertMik:
Hey, @luantaraschi thanks for active contributions
It would be nice if we could communicate more on development
Please join our Slack https://bit.ly/chat-codeceptjs

PR #5700, opened 2026-09-07
fix: report Bun runtime in environment info

### LINKED ISSUE #5698: 4.x: Bun: environment info reports no bunInfo, and nodeInfo is "Not Found" or the wrong runtime

Bun is a supported runtime for CodeceptJS 4 — the [4.0 release post](https://codecept.io/blog/codeceptjs-4/) says under *Native ESM*: "Your project needs `"type": "module"` and NodeJS 20 or Bun."

When running under Bun (`bunx --bun codeceptjs run` / `info`), I expect the environment report to tell me which runtime actually executed the process, and at which version — that is the whole point of the block, since `codeceptjs info` explicitly asks you to paste it into GitHub issues.

#### What do you get instead?

There is no `bunInfo`. The report only ever has `nodeInfo`, and under Bun it is wrong in one of two ways.

**1. `Not Found` — on a perfectly working install.**

```
 Environment information:

codeceptVersion:  "4.1.0"
nodeInfo:  Not Found
osInfo:  Linux 7.0 Ubuntu 24.04.4 LTS 24.04.4 LTS (Noble Numbat)
osBrowsers:  "chrome: 151.0.7922.169, edge: /, firefox: 153.0.4, safari: N/A"
playwrightBrowsers:  "chromium: 151.0.7922.34, firefox: 153.0, webkit: 26.5"
```

Same in the machine-info block printed by `run` / `run-workers` on failure:

```
nodeInfo:  Not Found
osInfo:  Linux 6.8 Ubuntu 24.04.4 LTS 24.04.4 LTS (Noble Numbat)
cpuInfo:  (16) x64 AMD Ryzen 9 7940HS w/ Radeon 780M Graphics
```

**2. A real but irrelevant version — which is worse, because it looks correct.**

If a Node install happens to be on `PATH`, the report shows *that* Node's version even though Node executed nothing. On my machine, running under Bun 1.4.2:

```
runtime bun:  1.4.2
getNodeInfo:  ["Node","24.18.0","/home/mirao/.nvm/versions/node/v24.18.0/bin/node"]
```

So an issue reported from a Bun project reads as "Node 24.18.0" and the Bun version — the thing that actually matters — is nowhere.

#### Cause

`lib/command/info.js` hardcodes `envinfo.helpers.getNodeInfo()` in both places, and `envinfo` answers by probing `PATH`, not by looking at the running process:

https://github.com/codeceptjs/CodeceptJS/blob/4.x/lib/command/info.js#L56
https://github.com/codeceptjs/CodeceptJS/blob/4.x/lib/command/info.js#L84

`Not Found` then has two independent triggers:

* **No Node at all.** Our CI image is Bun-only; there is no `node` binary to probe.
  ```
  $ env PATH=~/.bun/bin:/usr/bin:/bin bun --bun -e \
      'console.log(await require("envinfo").helpers.getNodeInfo())'
  [ "Node", "Not Found" ]
  ```
* **`bunx --bun` shadows `node` with its own shim.** Bun prepends a temp dir containing a `node` that re-execs Bun, and that shim rejects `--version`:
  ```
  PATH head:      /tmp/bunx-1000-bun@latest/node_modules/.bin | /tmp/bun-node-744846f84 | ...
  which node:     /tmp/bun-node-744846f84/node
  node --version: error: Missing script to execute. Pass --interactive to start the Node.js-compatible REPL.
  getNodeInfo:    ["Node","Not Found"]
  getbunInfo:     ["bun","1.4.2","/tmp/bun-node-744846f84/bun"]
  ```

Note the last line: `envinfo` already ships a Bun helper (`getbunInfo`, lowercase `b`), it is simply never called.

#### Minimal reproduction

```
package.json      { "name": "bun-info-repro", "type": "module", "private": true }
codecept.conf.js  (below)
```

```js
// codecept.conf.js
export const config = {
    tests: './*_test.js',
    helpers: {},
    name: 'bun-info-repro',
}
```

```
bun add --dev codeceptjs@4.1.0
bunx --bun codeceptjs info
```

→ `nodeInfo:  Not Found`, no `bunInfo`.

#### Suggested fix

Report the runtime that is actually executing CodeceptJS. `process.versions` is authoritative and needs no subprocess; the `PATH` probe is only correct on Node.

```diff
--- a/lib/command/info.js
+++ b/lib/command/info.js
@@
+async function getRuntimeInfo() {
+  // Report the runtime actually executing CodeceptJS. `process.versions` is authoritative;
+  // envinfo's PATH probe is not: under `bunx --bun` the `node` first on PATH is a Bun shim
+  // that has no --version, so getNodeInfo() reports "Not Found" on a working install.
+  if (process.versions.bun) return { bunInfo: ['bun', process.versions.bun, process.execPath] }
+  return { nodeInfo: await envinfo.helpers.getNodeInfo() }
+}
+
 export default async function (path) {
@@
   info.codeceptVersion = Codecept.version()
-  info.nodeInfo = await envinfo.helpers.getNodeInfo()
+  Object.assign(info, await getRuntimeInfo())
   info.osInfo = await envinfo.helpers.getOSInfo()
@@
 export const getMachineInfo = async () => {
   const info = {
-    nodeInfo: await envinfo.helpers.getNodeInfo(),
+    ...(await getRuntimeInfo()),
     osInfo: await envinfo.helpers.getOSInfo(),
```

The 3-element array keeps the shape `getNodeInfo()` returns, so the existing `Array.isArray(value)` printer prints `value[1]` — the version — unchanged.

Applied against 4.1.0 and verified locally on Bun 1.4.2 / Node 24.18.0:

```
$ bunx --bun codeceptjs info
codeceptVersion:  "4.1.0"
bunInfo:  1.4.2
osInfo:  Linux 7.0 Ubuntu 24.04.4 LTS 24.04.4 LTS (Noble Numbat)

$ bun --bun  -e 'const {getMachineInfo} = await import("codeceptjs/lib/command/info.js"); console.log(Object.keys(await getMachineInfo()))'
bunInfo:  1.4.2
[ "bunInfo", "osInfo", "cpuInfo", ... ]

$ node --input-type=module -e 'import {getMachineInfo} from "codeceptjs/lib/command/info.js"; console.log(Object.keys(await getMachineInfo()))'
nodeInfo:  24.18.0
[ "nodeInfo", "osInfo", "cpuInfo", ... ]
```

i.e. the Node output is byte-identical to today's, and Bun gains a correct `bunInfo`.

Two variations, if you prefer them:

* Emit `bunInfo` **in addition to** `nodeInfo` rather than instead of it. I'd argue against it — the Node line is meaningless noise on a Bun run, and `Not Found` reads as a broken install to anyone triaging the issue.
* Use `envinfo.helpers.getbunInfo()` instead of `process.versions.bun`. Also works, but it reports whichever `bun` is on `PATH`, which is not necessarily the one running the process — the same class of bug as the current code.

Happy to open a PR if the approach looks right.

#### Details

* CodeceptJS version: 4.1.0 (`4.x` branch HEAD is identical at these lines)
* Bun: 1.4.2
* Node (present but unused): 24.18.0
* OS: Ubuntu 24.04.4 LTS

---

_Created by AI agent (Claude Code)_

mirao:
Closes #5698

## Problem

`codecept info` — and the machine-info block printed by `run --verbose`, `run-workers --verbose` and `check` — hardcoded `envinfo.helpers.getNodeInfo()`. That helper probes `PATH`, not the running process, so under Bun the report was wrong in one of two ways:

- **`nodeInfo: Not Found` on a perfectly working install**, either because the image is Bun-only and has no `node` at all, or because `bunx --bun` prepends a temp dir whose `node` re-execs Bun and rejects `--version`.
- **A real but irrelevant Node version**, when some Node happens to be on `PATH`. This is the worse case, because it looks correct: an issue reported from a Bun project read as "Node 24.18.0" while Bun — the thing that actually matters — appeared nowhere.

Bun is a supported runtime for CodeceptJS 4, and `codeceptjs info` explicitly asks users to paste this block into GitHub issues, so getting the runtime wrong defeats the purpose of the block.

## Fix

Choose the helper by the runtime that is actually executing, and resolve both runtimes the same way through `envinfo`:

```js
async function getRuntimeInfo() {
  if (process.versions.bun) return { bunInfo: await envinfo.helpers.getbunInfo() }
  return { nodeInfo: await envinfo.helpers.getNodeInfo() }
}
```

`envinfo` already ships `getbunInfo` (lowercase `b`); it was simply never called. The branch has to read `process.versions.bun` because `envinfo` probes `PATH` and therefore cannot say which runtime is executing — that is runtime detection, distinct from version reporting.

This covers all four surfaces at once, since `info`, `run --verbose`, `run-workers --verbose` and `check` all print through `getMachineInfo()`.

**Node output is unchanged**, and the returned array keeps the shape of `getNodeInfo()`, so the existing `Array.isArray` printer shows the version at index 1 with no printer change. Nothing anywhere reads the `nodeInfo` key — it is only ever printed — so introducing `bunInfo` breaks no consumer. `envinfo` is pinned to an exact `7.21.0`, so `getbunInfo` is guaranteed present.

## Verification

Tested on the minimal reproduction from #5698, on Bun 1.4.2 / Node 24.18.0:

| scenario | before | after |
|---|---|---|
| `bunx --bun codeceptjs info` | `nodeInfo: Not Found` | `bunInfo: 1.4.2` |
| `bun --bun … info` / `run --verbose` | `nodeInfo: Not Found` | `bunInfo: 1.4.2` |
| Bun-only env, no `node` on PATH at all | `nodeInfo: Not Found` | `bunInfo: 1.4.2` |
| both runtimes on PATH, executed by Bun | `nodeInfo: 24.18.0` (wrong runtime) | `bunInfo: 1.4.2` |
| `node … info` / `run --verbose` | `nodeInfo: 24.18.0` | `nodeInfo: 24.18.0` (unchanged) |

Exactly one runtime line is printed, and `PATH` does not choose it — the executing runtime does.

Note that the version *number* still comes from `PATH`, so it is accurate whenever the executing binary is the one on `PATH`. That is the pre-existing behaviour of `getNodeInfo()` and it now applies identically to both runtimes: `getNodeInfo()` already misreports on plain Node when `PATH` disagrees with the executing binary (running v24.18.0 with v18.20.6 first on `PATH` reports `18.20.6`). Keeping Bun symmetric with Node was the deliberate choice here rather than special-casing one runtime.

## Tests

Three cases added to `test/unit/command/info_test.js`, covering the branch taken for each runtime and the array shape the printer depends on. They stub `process.versions.bun` so they run hermetically under either runtime, and assert no literal version, so they pass on machines where the probe returns `Not Found`.

Full unit suite: 773 passing. eslint and prettier clean.

### DIFF
--- a/lib/command/info.js
+++ b/lib/command/info.js
@@ -44,6 +44,14 @@ async function getOsBrowsers() {
   ].join(', ')
 }
 
+// Which runtime executes CodeceptJS can only be told from `process.versions`; envinfo probes
+// PATH, where under `bunx --bun` the first `node` is a Bun shim with no --version, so
+// getNodeInfo() returns "Not Found" on a working install.
+async function getRuntimeInfo() {
+  if (process.versions.bun) return { bunInfo: await envinfo.helpers.getbunInfo() }
+  return { nodeInfo: await envinfo.helpers.getNodeInfo() }
+}
+
 export default async function (path) {
   const testsPath = getTestRoot(path)
   const config = await getConfig(testsPath)
@@ -53,7 +61,7 @@ export default async function (path) {
   output.print('\n Environment information: \n')
   const info = {}
   info.codeceptVersion = Codecept.version()
-  info.nodeInfo = await envinfo.helpers.getNodeInfo()
+  Object.assign(info, await getRuntimeInfo())
   info.osInfo = await envinfo.helpers.getOSInfo()
   info.cpuInfo = await envinfo.helpers.getCPUInfo()
   info.osBrowsers = await getOsBrowsers()
@@ -77,11 +85,11 @@ export default async function (path) {
   output.print('***************************************')
 }
 
-export { parsePlaywrightBrowsers }
+export { parsePlaywrightBrowsers, getRuntimeInfo }
 
 export const getMachineInfo = async () => {
   const info = {
-    nodeInfo: await envinfo.helpers.getNodeInfo(),
+    ...(await getRuntimeInfo()),
     osInfo: await envinfo.helpers.getOSInfo(),
     cpuInfo: await envinfo.helpers.getCPUInfo(),
     chromeInfo: await envinfo.helpers.getChromeInfo(),
--- a/test/unit/command/info_test.js
+++ b/test/unit/command/info_test.js
@@ -1,7 +1,47 @@
 import { expect } from 'chai'
-import { parsePlaywrightBrowsers } from '../../../lib/command/info.js'
+import { parsePlaywrightBrowsers, getRuntimeInfo } from '../../../lib/command/info.js'
 
 describe('info command', () => {
+  describe('getRuntimeInfo', () => {
+    let originalBunVersion
+
+    beforeEach(() => {
+      originalBunVersion = process.versions.bun
+    })
+
+    afterEach(() => {
+      if (originalBunVersion === undefined) {
+        delete process.versions.bun
+      } else {
+        process.versions.bun = originalBunVersion
+      }
+    })
+
+    it('should report bunInfo when running under Bun', async () => {
+      process.versions.bun = '1.4.2'
+      const info = await getRuntimeInfo()
+      expect(info).to.have.property('bunInfo')
+      expect(info).to.not.have.property('nodeInfo')
+      expect(info.bunInfo[0]).to.equal('bun')
+    })
+
+    it('should report nodeInfo when not running under Bun', async () => {
+      delete process.versions.bun
+      const info = await getRuntimeInfo()
+      expect(info).to.have.property('nodeInfo')
+      expect(info).to.not.have.property('bunInfo')
+      expect(info.nodeInfo[0]).to.equal('Node')
+    })
+
+    it('should return an array so the printer shows the version at index 1', async () => {
+      delete process.versions.bun
+      const { nodeInfo } = await getRuntimeInfo()
+      expect(nodeInfo).to.be.an('array')
+      expect(nodeInfo.length).to.be.at.least(2)
+      expect(nodeInfo[1]).to.be.a('string')
+    })
+  })
+
   describe('parsePlaywrightBrowsers', () => {
     describe('old format (Playwright < 1.58)', () => {
       const oldFormatOutput = `browser: chromium version 140.0.7339.186


kobenguyent (approved):

DavertMik:
Agreed

PR #5701, opened 2026-09-07
feat: fillField works with role=combobox widgets

DavertMik:
## Problem

`I.fillField('Country', 'Ukr')` does not work against comboboxes whose trigger is not a text field — the shadcn / Base UI pattern where a `<button role="combobox">` opens a popup containing the search input.

Worse, against Base UI it **silently passes while doing nothing**. Base UI renders:

```html
<label for="value-input" id="country-label">Country</label>
<button role="combobox" aria-labelledby="country-label"></button>
<input id="value-input" name="content" tabindex="-1" aria-hidden="true" style="clip:rect(0,0,0,0);position:fixed">
```

The `<label for>` points at the **aria-hidden form-value proxy**, not at the trigger. `Locator.field.labelEquals` is tried first in `findFields` and matches that proxy, so `fillField` types into a zero-size invisible input and the step resolves green while the app never sees the interaction.

## What changed

**`Locator.field` skips aria-hidden proxies** (`lib/locator.js`) — one predicate added to the `input | textarea | select` branches of all four builders, so all three helpers are fixed in one place. With the proxy filtered, resolution falls through to `labelContains` → `aria-labelledby` → the trigger.

**`fillField` handles combobox triggers** (`lib/helper/extras/comboBox.js`) — a `<button role="combobox">`, a `<div role="combobox">`, or a readonly input trigger is expanded, its search input located, and the value typed there. The popup is left open and filtered so an option can be picked:

```js
I.fillField('Country', 'Ukr')
I.click('Ukraine')
```

Locating the input walks, in order: an editable descendant of the trigger (ARIA 1.1 wrapper, no click needed), the `aria-controls`/`aria-owns` container, a visible `[role=dialog]`/`[role=listbox]` popup, then the focused element. Two details are measured rather than assumed:

- Focus lands **asynchronously ~80ms after `click()` resolves** in Base UI (`pointerdown t=0 → click t=3 → focusin t=82`), so the search is polled to a 1s deadline instead of sampled once.
- The element focused **before** the click is excluded, so a previously focused field never receives the keystrokes.

A combobox exposing no text input raises `combobox exposes no text input to type into. Use I.selectOption() to pick one of its options.` rather than typing into nothing.

`role="combobox"` inputs (Base UI inline, MUI, Downshift) also route through this path, so they are typed with real keystrokes rather than a one-shot `fill()` — comboboxes that filter on `keydown` need that.

Detection reuses the single `evaluate` that `richTextEditor` already runs per `fillField`, so no extra round-trip is added to ordinary fills.

## Tests

New belt in `test/helper/webapi.js`, so it runs against Playwright, Puppeteer and WebDriver — mirroring the rich-editor belt. Fixtures load Base UI 1.0.0-rc.0 from CDN with no build step (`test/data/app/view/form/combobox/`), covering the button trigger, the inline input, and `custom_select.php` as the negative case.

Each variant asserts: typing filters the options, the picked option submits, a pre-populated value is rewritten, and keystrokes do not leak to an outer focused input. `Locator.field` gets unit tests in `test/unit/locator_test.js`.

Local: 9/9 combobox belt on Playwright and Puppeteer, 75/75 `fillField` on Playwright (rich editors included, no regressions), 15/15 `selectOption`, 772 unit tests. WebDriver has no local server here — relying on CI.

Note for review: `test/data/app/view/form/combobox/*` pull React and Base UI from **esm.sh**, a CDN this repo has not used before (existing fixtures use jsdelivr). Versions are pinned exactly.

https://claude.ai/code/session_01Ut1rjeUiZmKU7V61CeYhns

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -241,6 +241,13 @@ I.uncheckOption('Subscribe')
 
 > [selectOption](/web-api#iselectoption) works with native `<select>` elements as well as custom components using `role="combobox"` or `role="listbox"`.
 
+> [fillField](/web-api#ifillfield) works with searchable comboboxes built on `role="combobox"` — including triggers rendered as a `<button>`. The combobox is expanded, its search input is located, and the value is typed into it, leaving the list filtered so an option can be picked:
+>
+> ```js
+> I.fillField('Country', 'Ukr')
+> I.click('Ukraine')
+> ```
+
 ### Assertions
 
 CodeceptJS provides **built-in browser assertions** instead of generic `expect()` calls. This keeps tests readable and produces clear failure messages without extra assertion libraries.
--- a/lib/helper/Playwright.js
+++ b/lib/helper/Playwright.js
@@ -40,7 +40,8 @@ import { findByPlaywrightLocator } from './extras/PlaywrightLocator.js'
 import { dropFile } from './scripts/dropFile.js'
 import WebElement from '../element/WebElement.js'
 import { selectElement } from './extras/elementSelection.js'
-import { fillRichEditor } from './extras/richTextEditor.js'
+import { detectFillTarget, fillRichEditor, FILL_TARGET } from './extras/richTextEditor.js'
+import { fillComboBox } from './extras/comboBox.js'
 
 let playwright
 let perfTiming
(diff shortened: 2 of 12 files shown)


DavertMik:
Absoltely terrible code

PR #5714, opened 2026-09-10
fix: use bunx instead of npx to read Playwright versions under Bun

### LINKED ISSUE #5713: codecept info reports "Playwright not installed" under Bun because it shells out to `npx`
### What are you trying to achieve?

Run `codecept info` (or the machine-info block printed by `run --verbose`) under [Bun](https://bun.sh) and get the installed Playwright browser versions reported.

### What do you get instead?

`playwrightBrowsers: "Playwright not installed"` on an install where Playwright and its browsers are present and working.

`getPlaywrightBrowsers()` in `lib/command/info.js` hardcodes the Node package runner:

```js
const info = execSync('npx playwright install --dry-run').toString().trim()
```

A Bun-only environment has no `npx` on `PATH`, so `execSync` throws and the `catch` reports `Playwright not installed`. That is a false negative rather than a visible error, which makes it look like a genuine misconfiguration.

### Details

Reproduce with Bun installed and Node/npm absent from `PATH`:

```
$ PATH=/path/to/bun/bin:/usr/bin:/bin bun ./bin/codecept.js info <dir>
  bunInfo:  1.4.2
  playwrightBrowsers:  "Playwright not installed"
```

Expected:

```
```

The workaround currently needed is symlinking `npx` to `bun` in the image, which is fragile and shouldn't be required just to read a version string.

### Proposed fix

Pick the package runner from the runtime that is actually executing rather than from whatever `PATH` happens to hold — `bunx` under Bun, `npx` otherwise — detected via `process.versions.bun`. This is the same detection already used for `getRuntimeInfo()` in #5700. `bunx playwright install --dry-run` resolves the local `node_modules/.bin/playwright` and produces output identical to the `npx` form, so `parsePlaywrightBrowsers()` is unaffected and Node behaviour is unchanged.

This is the only place CodeceptJS actually shells out to `npx`; every other occurrence under `lib/` is documentation text.

mirao:
Closes #5713

### Problem

`getPlaywrightBrowsers()` in `lib/command/info.js` hardcoded the Node package runner:

```js
const info = execSync('npx playwright install --dry-run').toString().trim()
```

A Bun-only environment has no `npx` on `PATH`, so `execSync` throws and the `catch` reports `Playwright not installed` on an install where Playwright and its browsers are present. It is a false negative rather than a visible error, so it reads as a genuine misconfiguration.

### Fix

Pick the package runner from the runtime that is actually executing rather than from whatever `PATH` holds, via `process.versions.bun` — the same detection already used by `getRuntimeInfo()` in #5700.

```js
function getPackageRunner() {
  if (process.versions.bun) return 'bunx'
  return 'npx'
}
```

`bunx playwright install --dry-run` resolves the local `node_modules/.bin/playwright` and emits output identical to the `npx` form, so `parsePlaywrightBrowsers()` needs no change and Node behaviour is unchanged.

This is the only place CodeceptJS actually shells out to `npx`; every other occurrence under `lib/` is documentation text. It removes the need to symlink `npx` to `bun` in Bun images.

### Verification

`codecept info` with `npx` absent from `PATH` entirely, before and after:

```
# before
$ PATH=/path/to/bun/bin:/usr/bin:/bin bun ./bin/codecept.js info <dir>
  bunInfo:  1.4.2
  playwrightBrowsers:  "Playwright not installed"

# after
  bunInfo:  1.4.2
```

Node is byte-identical to before the change:

```
  nodeInfo:  24.18.0
```

Unit tests cover both branches of the detection; `test/unit/command/*.js` passes.

### DIFF
--- a/lib/command/info.js
+++ b/lib/command/info.js
@@ -21,9 +21,16 @@ function parsePlaywrightBrowsers(output) {
   return versions.join(', ')
 }
 
+// Bun has its own package runner and a Bun-only install has no `npx` on PATH at all, so the
+// runner has to follow the runtime that is actually executing rather than what PATH happens to hold.
+function getPackageRunner() {
+  if (process.versions.bun) return 'bunx'
+  return 'npx'
+}
+
 async function getPlaywrightBrowsers() {
   try {
-    const info = execSync('npx playwright install --dry-run').toString().trim()
+    const info = execSync(`${getPackageRunner()} playwright install --dry-run`).toString().trim()
     return parsePlaywrightBrowsers(info)
   } catch (err) {
     return 'Playwright not installed'
@@ -85,7 +92,7 @@ export default async function (path) {
   output.print('***************************************')
 }
 
-export { parsePlaywrightBrowsers, getRuntimeInfo }
+export { parsePlaywrightBrowsers, getRuntimeInfo, getPackageRunner }
 
 export const getMachineInfo = async () => {
   const info = {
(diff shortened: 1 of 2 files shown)


DavertMik:
Thanks! Let's have more bun!

